PR #9, opened 2019-11-26
fix compile issue caused by fbs

halx99:


halx99:
see also: https://github.com/simdsoft/engine-v5/issues/8

PR #36, opened 2020-02-04
Refactor UserDefault, use file mapping implementation.

halx99:
Refactor UserDefault, use file mapping & plain binary encode/decode for all platforms


cocos/base/CCUserDefault.cpp
```diff
@@ -407,76 +243,71 @@ void UserDefault::setDelegate(UserDefault *delegate)
...
+        posix_lseek(_fd, 0, SEEK_SET);
 
-    return bRet;
-}
+        if (filesize < _curMapSize) {
+            posix_fsetsize(_fd, _curMapSize);
+        }
+        /// load to memory _values
+        _rwmmap = std::make_shared<mio::mmap_sink>(posix_fd2fh(_fd), 0, _curMapSize);
+        if (_rwmmap->is_mapped()) { // no error
```
huizhen: what if _rwmmap->is_mapped() return false, should have some fallback method here.

cocos/base/CCUserDefault.cpp
```diff
@@ -407,76 +243,71 @@ void UserDefault::setDelegate(UserDefault *delegate)
...
-    pugi::xml_node rootNode;
-    pugi::xml_document doc;
-    pugi::xml_node node;
-
-    // check the params
-    if (!key)
-    {
-        CCLOG("the key is invalid");
-        return;
+    // TODO: Win32 store to UserDefault.plain file?
```
huizhen: For win32, indeed UserDefault.plain could be better, especially better for use to debug. may be put it in next enhancement.

cocos/base/CCUserDefault.cpp
halx99: When the file size is zero, the map will failed,  maybe there need set _curMapSize to 0 when map failed, what do you think?

cocos/base/CCUserDefault.cpp
halx99: Yeah, I will continue this PR later.

cocos/base/CCUserDefault.cpp
```diff
@@ -407,76 +243,71 @@ void UserDefault::setDelegate(UserDefault *delegate)
...
+    yasio::obstream obs;
+    obs.write_i7(this->_values.size());
+    for (auto& item : this->_values) {
```
huizhen: Everytime I do a flush() operation like set or delete, it iterate all the _value map to wirte to the obstream. What if the map is already huge data and more ever, one do the  "delete" or "set" operation in a for each loop. Maybe you should do some pressure test to make sure its robust. And by the way, is there any more eligant or efficient way to do the flush?  Like perform flush action in one time per frame or other way.

cocos/base/CCUserDefault.cpp
huizhen: Set _curMapSize to 0 can be ok, but add some fail log is better. Another thing is when you set _curMapSize to 0, next time you flush, you may encounter the error code, better to catch the error_code and print fail log instead. Check the code below:
if (obs.length() > _curMapSize) {
        _rwmmap->unmap();
        std::error_code error;
        _curMapSize <<= 1; // X2
        _rwmmap->map(posix_fd2fh(_fd), 0, _curMapSize, error);
    }

As shown in the flush code, if _curMapSize is 0, I believe map() will failed and result in error_code.

halx99: Yeah, agree with you.

cocos/base/CCUserDefault.cpp
halx99: Yeah, even through, it's still fast than xml serialize per flush(), and generally,  this UserDefault may 
 only design for small amount of data(0~1000 entities).

halx99: There is a optimizing idea, always use append until the serialized binary data large than 4k.

halx99:
@huizhen you can continue review.
For diff with first commit please see: https://github.com/halx99/engine-v5/commit/e312fd356e4f9081daf5ef79fc8f1db1148b7acf

halx99:
handle it & fix enlarge mapsize problem.

PR #40, opened 2020-02-06
Refactor UserDefault, cbc128 encrypt support.

halx99:


huizhen:
Seems good, upload the test example latter on.

halx99:
ok, I will add testcase later

PR #44, opened 2020-02-07
Fix travis ci ios build issue

halx99:


halx99:
everything is ok except sign

PR #56, opened 2020-02-14
Make astc works on ios platform.

halx99:


cocos/renderer/backend/metal/Utils.mm
```diff
@@ -123,6 +123,10 @@ MTLPixelFormat getSupportedDepthStencilFormat()
             return MTLPixelFormatPVRTC_RGB_2BPP;
         case PixelFormat::ETC:
             return MTLPixelFormatETC2_RGB8;
+        case PixelFormat::ASTC4:
+            return MTLPixelFormatASTC_4x4_LDR;
```
weiwest: use 
MTLPixelFormatASTC_4x4_sRGB  
MTLPixelFormatASTC_8x8_sRGB  
Instead

halx99:
so let's merge it

PR #58, opened 2020-02-14
Make multi-meta-textures feature works on Apple Metal.

halx99:


halx99:
android, ios etc1 & effects test ok

PR #73, opened 2020-02-24
Clean JSB sources, optmize script event dispatch.

halx99:


cocos/scripting/lua-bindings/manual/CCLuaEngine.cpp
```diff
@@ -202,61 +202,58 @@ bool LuaEngine::parseConfig(ConfigType type, const std::string& str)
...
+int LuaEngine::sendEvent(const ScriptEvent& evt)
 {
-    if (NULL == evt)
-        return 0;
-    
-    switch (evt->type)
+    switch (evt.type)
     {
         case kNodeEvent:
             {
```
halx99: The parantheses doesn't need

PR #100, opened 2020-05-20
Resume scissor of CCScrollView, indent correct with vs2019

halx99:
see also: https://github.com/c4games/engine-x/issues/97


halx99:
ok, merge it

PR #124, opened 2020-07-30
Add CMAKE option CC_USE_ALSOFT_ON_APPLE to use kcat openal-soft 

halx99:
- [x] Add CMAKE option CC_USE_ALSOFT_ON_APPLE to use kcat openal-soft, currently, by default is OFF
- [x] Use apple system decoder for non .ogg audio files

Known issue use kcat openal-soft on apple platform
* Will random crash at CoreAudio init SIGABRT at AudioComponentFindNext on iphonesimulator, please see: https://github.com/kcat/openal-soft/issues/455, so currently, the provided libopenal.a for ios only contains arm device binaries.


halx99:
- [ ] Add CMAKE option CC_USE_ALSOFT_ON_APPLE to use kcat openal-soft
- [ ] Use apple system decoder for non .ogg audio files

PR #134, opened 2020-08-02
Refactor full screen support

halx99:
fix different behavior when enter fullscreen between 1 and >=2 times


halx99:
@aismann please help me test whehter this PR works well for full screen mode

halx99:
I have test 2 moniters, works fine.

PR #136, opened 2020-08-03
The switch to the third monitor was not working

aismann:


halx99:
looks good to me, thanks

PR #154, opened 2020-08-09
Disable double use in chipmunk_types.h instead CMakeLists.txt 

halx99:
More stable for ENGX DNT use double for precision.


halx99:
See also: https://github.com/c4games/engine-x/issues/150

PR #167, opened 2020-08-19
Fix label crashing

halx99:


halx99:
## Bug affects
  * official-v4
  * EGNX

## Root cause
The class ```Label::BatchCommand``` not properly for std::vector resize, see the std::vector enlarge when resize step:
1. Alloc new buffer for storage element
2. Call assign constructor to assign old element to new element, **if move assignment present, stl vector will call it perferred**
3. Call destructor of old element

If a class use to std::vector has any non-RAII resource, above step will cause dead pointer and crashing

## Bug report links:
  * https://github.com/cocos2d/cocos2d-x/issues/20523
  * https://discuss.cocos2d-x.org/t/label-render-crash/50218/19?u=ericb365

## Other solution:
  * https://github.com/zhongfq/cocos-lua/commit/9af9ea7b343a2fd3916a2beb16e8f8e29ed67d85
  * https://github.com/sgosztyla/cocos2d-x/commit/2f87a7af3744ff04a55e44aa2acdbacbe0626d4f

PR #171, opened 2020-08-25
Add utils::createInstance to instead macro CREATE_FUNC

halx99:


aismann:
What is the reason?
Any API changes too?

halx99:
No chnages, just add the new API

halx99:
Simple usage:
```cpp
struct LevelData {
 // ...
};
class LevelScene : public Scene
{
public:
    bool initWithLevelData(const LevelData& data) {
        if(!Scene::init()) return false;
        // don somethings init with the levelData
    }
};

// Then the scene can be create anywhere with the custom init function 'initWithLevelData'
auto scene = utils::createInstance<LevelScene>(&LevelScene::initWithLevelData, levelData);
```

aismann:
Ist there any cpptest example?
Documentation about new API too?

halx99:
The ```cpp-empty-tests``` and the template project ```HelloCpp``` use the new API

PR #175, opened 2020-08-27
Freetype stream parsing

halx99:


halx99:
If your .ttf file is large and up to MB, this is useful reduce memory use
If you only 1 .ttf file and file size is xKB, you may don't need enabled, by default stream parsing is disabled, use follow API to enabled:
```cpp
cocos2d::FontFreeType::setStreamParsingEnabled(true);
```

PR #180, opened 2020-08-28
Add switch renderer backend to GLES for iOS, win32

halx99:
* Add CMake ```-DCC_FORCE_USE_GLES``` to control whether force use gles on win32,iOS
* The old ```CC_USE_GLES_ON_DESKTOP``` for win32 specific replaced by the ```-DCC_FORCE_USE_GLES=TRUE```
* By default, CC_FORCE_USE_GLES=FALSE


halx99:
mark: when switch to gles the first sub test of ```Shader - Sprite``` performance is slow than metal
* GELS: 30fps on ```iOS Simulator```
* Metal: 60fps on ```iOS Simulator```

PR #186, opened 2020-08-29
Fix ASTC bytesPerRow calculation

halx99:


halx99:
Other renderer engines references:
https://github.com/google/filament/blob/main/filament/backend/src/BackendUtils.cpp#L166
https://github.com/google/filament/blob/main/filament/backend/src/metal/MetalHandles.mm#L422
https://github.com/bkaradzic/bgfx/blob/master/src/renderer_mtl.mm#L2841

halx99:
more refer: https://forums.developer.nvidia.com/t/how-does-gl-unpack-alignment-work/39432/4

PR #189, opened 2020-08-31
Update README.md

aismann:
Some more better description what ENGX is.


halx99:
looks good to me, thanks.

PR #229, opened 2020-10-07
Use pyenv manage python version for travis-ci

halx99:


halx99:
see also: https://github.com/napalm-automation/napalm-nxos/issues/54

PR #246, opened 2020-10-23
Refine ProgramState mat id

halx99:


halx99:
> > We also can provide setUniformID manually for user to avoid calculate uniformID with uniform buffer automatically, what do you think about?
> 
> That would be a good idea, in case the performance is affected by the XXH32 calculation.

@rh101 Please help review the latest commit, I implement the idea

rh101:
The latest commit with the `setUniformID` seems fine.

halx99:
Thanks

halx99:
well, i have test again, with seems with your implementation, cocos-re full fbo test fps is 36fps

halx99:
so, currently, i think your implementation is best

rh101:
I don't know much about rendering pipelines and such, so please correct me if I'm wrong, but isn't the sprite texture stored in the FBO?  If that is the case, then the XXH32 processing time would greatly affected by the texture size, and that would explain the low FPS.

halx99:
Because every render node call setUniform per frame

rh101:
I understand, so it would be quite intensive if the hashing algorithm is running every frame.

halx99:
> I understand, so it would be quite intensive if the hashing algorithm is running every frame.

Yes, so I modify the code, let's user to call updateUniformID when their shader have unstable uniform variables needs to set

halx99:
@rh101 If you have no other qusetion, I will merge this pr

rh101:
The changes look good.  Thank you for working on all of this.

halx99:
Thanks, merged

PR #257, opened 2020-11-11
extend lua by register extensions

wtfdnvxk:
advoid old lua register pullute the global environment


halx99:
thanks, looks good to me, once ci check passed, I  will merge this excellent pr

halx99:
with this pr, wen can use lua extension as follow:
local yasio = require 'yasio'
local cjson = require 'cjson'

PR #259, opened 2020-11-11
Fix crash when enable encrypt for UserDefault

x8710162:
write with variable  size,  calculate size use int32_t


halx99:
@x8710162 Excellent, thanks

PR #268, opened 2020-11-18
Fix manifest version greater check

Obg1:


halx99:
thanks, looks good to me

PR #276, opened 2020-11-23
Same as on 3.17 (with small changes)

aismann:
The Box2d test is the same as on 3.14 (small changes because of engine-x and Box2D version)
So somebody can try to add the implementation of the commented SetDebugDraw() method: 
//    b2World::SetDebugDraw(&Box2DTest::draw);


halx99:
Thanks

PR #277, opened 2020-11-23
Update cocos.bat

aismann:
I got always:
Python was not found; run without arguments to install from the Microsoft Store, or disable this shortcut from Settings > Manage App Execution Aliases.

Normaly "Python.exe" is on the path variable too. and will be found.
=> Works now with removed "@python".


halx99:
Thanks

PR #283, opened 2020-11-27
Allow user to set custom program type value

rh101:
Make setProgramType public so it can be used to set the program type of custom shaders.


halx99:
Thanks

PR #285, opened 2020-11-27
Fix ListView dimensions when child elements are scaled

rh101:
If a ListView has items that are scaled, then the ListView dimensions are not correctly set based on the size of those items, so some of the items cannot be scrolled to.  More info [here](https://github.com/cocos2d/cocos2d-x/issues/20431)


halx99 (approved):

halx99:
thanks

PR #287, opened 2020-11-27
MOD: support metal and opengl as the backend render simultaneously

GaeaUni:
使用变量代替宏判断，支持ios全系列机型。


halx99:
Now it's offical supported by axmol dev(v3) since `Jan 16, 2026` with dynamic RHI implementation.

PR #300, opened 2020-12-03
same/correct wording style

aismann:


halx99:
looks good to me, thanks

PR #310, opened 2020-12-16
FIX: next drawable maybe nil crash

wtfdnvxk:


halx99:
Thanks, fixed by https://github.com/c4games/engine-x/commit/208d07697f82ddb93f4cb76217b82c7b8ef81665

PR #340, opened 2021-04-22
Enhance FileStream support to allow usage of virtual file systems

rh101:
This makes FileStream more flexible.  A use case would be allowing the implementation of a FileStream that uses RAM as the disk storage, or reading resources directly from ZIP files without having to extract them first (for instance, using [PhysFS](https://icculus.org/physfs/)).

All file operations must go through `FileUtils`, and a new FileStream is created via `FileUtils::openFileStream()`.  A custom FileUtils can be created, and `openFileStream()` can be overriden to provide custom implementations of `FileStream`.

Note:  The changes are not yet complete, as there may be sections that are still using direct access (through `fopen` etc).  This is just a proof of concept.  It's working on Windows, so sound and graphics resources are being loaded transparently from a ZIP file (using PhysFS).


rh101:
[mio](https://github.com/mandreyel/mio) doesn't seem to support custom file IO, so it needs direct access to the underlying file system path.  I've added `FileUtils::getNativeWritableAbsolutePath()`, which is guaranteed to return the native file system path in these cases.  The only times I can imagine `getNativeWritableAbsolutePath` being used is in cases where something needs write access to the underlying file system, so it's fine that it bypasses any VFS (virtual file system) that may be implemented.

By default `FileUtils::getWritablePath()` will call `getNativeWritableAbsolutePath`, but if a VFS is used, `getWritablePath` would be overriden to return a suitable path for the VFS (such as "/").

cocos/2d/CCFontFreeType.cpp
```diff
@@ -174,21 +174,24 @@ bool FontFreeType::createFontObject(const std::string &fontName, float fontSize)
...
         std::unique_ptr<FT_StreamRec> fts(new FT_StreamRec());
         fts->read = ft_stream_read_callback;
         fts->close = ft_stream_close_callback;
-        fts->size = fs.seek(0, SEEK_END);
-        fs.seek(0, SEEK_SET);
+        fts->size = fs->seek(0, SEEK_END);
+        fs->seek(0, SEEK_SET);
 
-        fts->descriptor.pointer = new FileStream(std::move(fs));
+        fts->descriptor.pointer = std::move(fs);
```
halx99, line 190: The std::move can be remove

rh101: Good point, I'll update it soon.

cocos/platform/CCFileUtils.cpp
```diff
@@ -1101,6 +1105,11 @@ void FileUtils::listFilesRecursivelyAsync(const std::string& dirPath, std::funct
     }, std::move(callback));
 }
 
+FileStream* FileUtils::openFileStream(const std::string& filePath, FileStream::Mode mode)
```
halx99, line 1108: 1. Maybe add matched closeFileStream is better
2. Maybe openFileStream retrun nullptr when open failed, such as
```cpp
PosixFileStream fs;
if(fs.open(fullPath, mode)) return new PosixFileStream(std::move(fs));
return nullptr;
```

rh101: Yes that is possible, I'll just have to add a default constructor to it so it will allow `PosixFileStream fs;` to work.

What do you think about renaming `FileUtils::openFileStream` to `FileUtils::createFileStream`? That way it won't confuse things with the `open` in the name, and we won't need to add a `closeFileStream()` method, since closing the stream should be as simple as "stream->close()".

The only type user could would know is FileStream.  No code outside openFileStream should ever create a concrete implementation of a FileStream (like PosixFileStream).

halx99: It's may ok, my purpose is reduce new/delete when open file failed.

rh101: Changes done.  I didn't change the `openFileStream()` name, since the name does represent the purpose of that method.

All usages of openFileStream now check for a nullptr on a failed open.

halx99: ok, looks good to me

halx99: Merge it

halx99: Well, I have a question, does the openFileStream require fullPath?

rh101: It depends on the implementation of FileStream.  In terms of the default PosixFileStream, the value passed in would need to be a full path, but if FileStream was a wrapper around a virtual file system, then it wouldn't need to be a full path. Does that make sense?

rh101: > Merge it

I don't have permission on this project to merge.  Also, I haven't checked all code yet for any direct file access, so I'll post back here when I've reviewed the rest of the code for any remaining `fopen` and such.

halx99:
ok, I merged it

rh101:
All good, thank you.  In a few days I'll continue checking the other sections of code for any direct file access that isn't going through FileStream/FileUtils, and I'll submit another PR if I find anything that needs updating.

halx99:
I found a crash problem due to the _iof not initialized, and fixed on latest commit.

rh101:
> I found a crash problem due to the `_iof` not initialized, and fixed on the commit: [63ac990](https://github.com/c4games/engine-x/commit/63ac9904c47acc5936ca53d37778b85e36d8f905)

Excellent, thank you!

PR #341, opened 2021-04-24
Add support for FileStream in ZipUtils and ZipFile

rh101:
This will allow unzip to route calls through FileStream objects.


rh101:
@halx99 The changes to ZipUtils should be working now.  It passes all tests, including multi-file archives (.zip, .z01, .z02 etc.),

I found a few other locations using direct file access via `fopen`, but I'll put them in a separate pull request once this one is merged in.

halx99 (approved):

halx99:
ok, merged

PR #342, opened 2021-04-25
Ensure FileStream is used for file access where possible

rh101:
Use FileStream where we can.  Certain libraries do not support overriding the file access routines, so they need the raw filesystem path, such as SQLite, CURL etc, so they will still work as intended.


rh101:
@halx99 It's almost ready, just still fixing a bug in somewhere in the ZipUtils.  I'll let you know when it's all done.

rh101:
@halx99 All done. This has been tested on Windows and Android with the default PosixFileStream, and all works fine.  I've also tested with a custom FileStream implementation that wraps PhysFS as a virtual file system into ZIP files, and that works well on Android and Windows.  I haven't had a chance to test iOS or Linux, but they should work as well, since I've updated any relevant fopen references in the platform specific code.

cocos/platform/CCFileUtils.cpp
```diff
@@ -497,20 +497,18 @@ void FileUtils::writeDataToFile(Data data, const std::string& fullPath, std::fun
...
         // Read the file from hardware
-        FILE* fp = fopen(fullPath.c_str(), mode);
-        CC_BREAK_IF(!fp);
-        fwrite(data, dataSize, 1, fp);
-
-        fclose(fp);
+        CC_BREAK_IF(!fileStream);
 
+        fileStream->write(data, dataSize);
+        fileStream->close();
```
halx99, line 510: Does this line could be remove?

halx99:
> @halx99 All done. This has been tested on Windows and Android with the default PosixFileStream, and all works fine. I've also tested with a custom FileStream implementation that wraps PhysFS as a virtual file system into ZIP files, and that works well on Android and Windows. I haven't had a chance to test iOS or Linux, but they should work as well, since I've updated any relevant fopen references in the platform specific code.

Thanks for you provide this excellent feature, and I have idea abount use openFileStream, just see follow code:
```cpp
std::unique_ptr<FileStream> fs(FileUtils::getInstance()->openFileStream(fullPath, mode);
```

rh101:
This is the perfect candidate for the use of a `unique_ptr`, but initially I assumed it wasn't preferred after searching the engine code and find very little usage of it.  I would be more than happy to change it to use `unique_ptr`, since it wouldn't affect performance, but would actually simplify the usage of FileStream, and also reduce or prevent any memory leaks from forgetting to delete raw pointers.

Do you want me to go ahead with changing it to use `unique_ptr`?

cocos/platform/CCFileUtils.cpp
rh101: Yep, good catch, I'll update it.

rh101:
@halx99 I just noticed a bit of an issue.  What data do you prefer FileStream::seek() to return? Some implementations of seek return the current position, while others return 0 for success, -1 for error.  Which one is preferred?

I just found sections of code that try to use the return value, assuming it's size, so I need to fix them all, but need to know what `seek` should return before I do so.

halx99:
I think, if there is tell API, should like fseek, otherwise, should like lseek

halx99:
You can create a seperate pr to use unique_ptr, once this pr merged.

rh101:
> You can create a seperate pr to use unique_ptr, once this pr merged.

Will do! This PR should be complete unless you notice something that needs fixing.

halx99 (approved):

halx99:
looks good to me, merged

rh101:
Excellent! New PR should be up soon with the `unique_ptr` changes.

PR #345, opened 2021-04-26
Use unique_ptr for FileStream

rh101:
While I was updating the code to use unique_ptr, I found a few issues with zip close method overrides not correctly freeing the FileStream data, so they've been fixed as well (in ZipUtils, AssetManager and AssetManagerEx)

### DIFF
--- a/cocos/audio/src/AudioDecoderOgg.cpp
+++ b/cocos/audio/src/AudioDecoderOgg.cpp
@@ -56,6 +56,8 @@ namespace cocos2d {
     }
 
     static int ov_fclose_r(void* handle) {
+        auto* fs = static_cast<FileStream*>(handle);
+        delete fs;
         return 0;
     }
     
@@ -70,8 +72,8 @@ namespace cocos2d {
 
     bool AudioDecoderOgg::open(const std::string& fullPath)
     {
-        _fileStream = FileUtils::getInstance()->openFileStream(fullPath, FileStream::Mode::READ);
-        if (!_fileStream)
+        auto fs = FileUtils::getInstance()->openFileStream(fullPath, FileStream::Mode::READ).release();
+        if (!fs)
         {
             ALOGE("Trouble with ogg(1): %s\n", strerror(errno));
             return false;
@@ -83,8 +85,8 @@ namespace cocos2d {
                ov_fclose_r,
                ov_ftell_r
         };
-              
-        if (0 == ov_open_callbacks(_fileStream.get(), &_vf, nullptr, 0, OV_CALLBACKS_POSIX))
+
+        if (0 == ov_open_callbacks(fs, &_vf, nullptr, 0, OV_CALLBACKS_POSIX))
         {
             // header
             vorbis_info* vi = ov_info(&_vf, -1);
@@ -104,7 +106,6 @@ namespace cocos2d {
         {
             ov_clear(&_vf);
             _isOpened = false;
-            _fileStream.reset();
         }
     }
 
--- a/cocos/audio/src/AudioDecoderOgg.cpp
+++ b/cocos/audio/src/AudioDecoderOgg.cpp
@@ -39,21 +39,25 @@ namespace cocos2d {
 
     static size_t ov_fread_r(void* buffer, size_t element_size, size_t element_count, void* handle)
     {
-        return ((FileStream*)handle)->read(buffer, static_cast<uint32_t>(element_size * element_count));
+        auto& fs = *static_cast<std::unique_ptr<FileStream>*>(handle);
+        return fs->read(buffer, static_cast<uint32_t>(element_size * element_count));
     }
 
     static int ov_fseek_r(void * handle, ogg_int64_t offset, int whence)
     {
-        return ((FileStream*)handle)->seek(offset, whence) < 0 ? -1 : 0;
+        auto& fs = *static_cast<std::unique_ptr<FileStream>*>(handle);
+        return fs->seek(offset, whence) < 0 ? -1 : 0;
     }
     
     static long ov_ftell_r(void * handle)
     {
-        return ((FileStream*)handle)->seek(0, SEEK_CUR);
+        auto& fs = *static_cast<std::unique_ptr<FileStream>*>(handle);
+        return fs->tell();
     }
 
     static int ov_fclose_r(void* handle) {
-        ((FileStream*)handle)->close();
+        auto& fs = *static_cast<std::unique_ptr<FileStream>*>(handle);
+        fs = nullptr;
         return 0;
     }
     
@@ -102,6 +106,7 @@ namespace cocos2d {
         {
             ov_clear(&_vf);
             _isOpened = false;
+            _fileStream = nullptr;
         }
     }
 
--- a/cocos/audio/src/AudioDecoderOgg.cpp
+++ b/cocos/audio/src/AudioDecoderOgg.cpp
@@ -39,25 +39,23 @@ namespace cocos2d {
 
     static size_t ov_fread_r(void* buffer, size_t element_size, size_t element_count, void* handle)
     {
-        auto& fs = *static_cast<std::unique_ptr<FileStream>*>(handle);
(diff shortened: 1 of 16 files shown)


cocos/platform/linux/CCDevice-linux.cpp
```diff
@@ -346,9 +346,9 @@ class BitmapDC
         std::transform(lowerCasePath.begin(), lowerCasePath.end(), lowerCasePath.begin(), ::tolower);
         if ( lowerCasePath.find(".ttf") != std::string::npos ) {
             fontPath = cocos2d::FileUtils::getInstance()->fullPathForFilename(fontPath);
-            auto* fileStream = cocos2d::FileUtils::getInstance()->openFileStream(fontPath, FileStream::Mode::READ);
+            auto fileStream = cocos2d::FileUtils::getInstance()->openFileStream(fontPath, FileStream::Mode::READ);
             if ( fileStream ) {
-                delete fileStream;
+                fileStream = nullptr;
```
halx99, line 351: redundant code since unique_ptr auto manage memory

cocos/network/CCDownloader-curl.cpp
```diff
@@ -284,10 +284,10 @@ class DownloadTaskCURL : public IDownloadTask {
     string _tempFileName;
     std::string _checksumFileName;
     vector<unsigned char> _buf;
-    FileStream* _fs = nullptr;
+    std::unique_ptr<FileStream> _fs = nullptr;
```
halx99, line 287: Since we use unique_ptr, this line could be simplified to `std::unique_ptr<FileStream> _fs{};`

cocos/platform/CCImage.cpp
```diff
@@ -2178,34 +2178,34 @@ bool Image::saveImageToPNG(const std::string& filePath, bool isToRGB)
...
-            delete outStream;
+            outStream = nullptr;
             break;
         }
 
         info_ptr = png_create_info_struct(png_ptr);
         if (nullptr == info_ptr)
         {
-            delete outStream;
+            outStream = nullptr;
```
halx99, line 2196: I prefer use `outStream.reset` when I want release the memory explicitly.

cocos/platform/linux/CCDevice-linux.cpp
rh101: Oh I left that in there because it operates the same as the original non-filestream code, which was freeing the `FILE*` if the conditional result is true, so fileStream is null after the conditional block is finished.  Assuming no more code will go after the `if` conditional block, I'll remove the `fileStream = nullptr;`

cocos/network/CCDownloader-curl.cpp
rh101: I'll update all instances to use the `{}`.

cocos/platform/CCImage.cpp
rh101: If that's the preferred way, then I'll update all instances of it to use `reset()`.

cocos/audio/src/AudioDecoderOgg.cpp
```diff
@@ -39,21 +39,25 @@ namespace cocos2d {
...
     static long ov_ftell_r(void * handle)
     {
-        return ((FileStream*)handle)->seek(0, SEEK_CUR);
+        auto& fs = *static_cast<std::unique_ptr<FileStream>*>(handle);
+        return fs->tell();
     }
 
     static int ov_fclose_r(void* handle) {
-        ((FileStream*)handle)->close();
+        auto& fs = *static_cast<std::unique_ptr<FileStream>*>(handle);
```
halx99, line 59: Use `ov_open_callbacks(_fileStream.get(), ...) ` for ogg callbacks, and access via `FileStream*` in ogg callbacks may be better.

rh101: The problem is this `ov_fclose_r`:
```
    static int ov_fclose_r(void* handle) {
        auto& fs = *static_cast<std::unique_ptr<FileStream>*>(handle);
        fs = nullptr;
        return 0;
    }
```

I can change to be like this:
```
    static int ov_fclose_r(void* handle) {
        auto* fs = static_cast<FileStream*>(handle);
        fs->close();
        return 0;
    }
```

The pointer will still be valid until this is called to free the FileStream:
```
    void AudioDecoderOgg::close()
    {
        if (isOpened())
        {
            ov_clear(&_vf);
            _isOpened = false;
            _fileStream.reset();
        }
    }
```

halx99: Maybe we can make ov_fclose_r empty when the fileStream auto managed by unique_ptr

rh101: That should work too, so I'll change it so `ov_fclose_r` does nothing.  If any issues come up because of it, we can just add the call to FileStream::close() in there.

halx99: sounds good

halx99: or maybe same with FreeType streaming parsing mode is better?
```cpp
       auto fullPath = FileUtils::getInstance()->fullPathForFilename(fontName);
        if (fullPath.empty()) return false;

        auto fs = FileUtils::getInstance()->openFileStream(fullPath, FileStream::Mode::READ);
        if (!fs)
        {
            return false;
        }

        std::unique_ptr<FT_StreamRec> fts(new FT_StreamRec());
        fts->read = ft_stream_read_callback;
        fts->close = ft_stream_close_callback;
        fs->seek(0, SEEK_END);
        fts->size = fs->tell();
        fs->seek(0, SEEK_SET);

        fts->descriptor.pointer = fs.release();
```

rh101: > or maybe same with FreeType streaming parsing mode is better?

Sorry, I'm not sure what you mean by that.

halx99: 1. Remove std::unique_ptr<FileStream> _fileStream from class AudioDecoderOgg
2. delete stream ta ov_close_r
```
static int ov_fclose_r(void* handle) {
auto* fs = static_cast<FileStream*>(handle);
delete fs;
return 0;
}
```
3. Then the ov_open_callbacks code could be:
```cpp
bool AudioDecoderOgg::open(const std::string& fullPath)
    {
        auto fs = FileUtils::getInstance()->openFileStream(fullPath, FileStream::Mode::READ);
        if (!fs)
        {
            ALOGE("Trouble with ogg(1): %s\n", strerror(errno));
            return false;
        }

        static ov_callbacks OV_CALLBACKS_POSIX = {
               ov_fread_r,
               ov_fseek_r,
               ov_fclose_r,
               ov_ftell_r
        };
              
        if (0 == ov_open_callbacks(fs.release(), &_vf, nullptr, 0, OV_CALLBACKS_POSIX))
        {
            // header
            vorbis_info* vi = ov_info(&_vf, -1);
            _sampleRate = static_cast<uint32_t>(vi->rate);
            _channelCount = vi->channels;
            _bytesPerBlock = vi->channels * sizeof(short);
            _totalFrames = static_cast<uint32_t>(ov_pcm_total(&_vf, -1));
            _isOpened = true;
            return true;
        }
        return false;
    }

```

rh101: Oh I understand what you mean now, and yes that is a cleaner implementation, reducing the duplication of the pointer storage. I'll modify it now.

rh101: I'll add `.release()` on the same line so that there isn't a duplicate unique_ptr and raw pointer floating around.  It'll prevent any accidental usage of the unique_ptr after it's been released.

halx99: it's ok, good job.

halx99 (approved):

halx99:
merged

rh101:
Thank you!

PR #346, opened 2021-04-27
Add std::iostream support to FileStream

rh101:
This adds `std::iostream` support to the FileStream, so you should be able use it with any code that accepts a `std::istream/std::ostream/std::iostream`.

Example of usage:
```
bool FileCopy(const std::string& src_loc, const std::string& dest_loc)
{
    auto in = FileStream::ifstream(src_loc);

    if (in)
    {
        auto out = FileStream::ofstream(dest_loc);

        const int BUFFER_SIZE = 4096;
        std::vector<char> buffer(BUFFER_SIZE, 0);

        while (true)
        {
            in.read(buffer.data(), BUFFER_SIZE);
            const std::streamsize s = (in ? BUFFER_SIZE : in.gcount());

            out.write(buffer.data(), s);

            if (!in)
            {
                break;
            }
        }

        return true;
    }

    return false;
}
```

Would this functionality be useful at all?


halx99:
Why not use filestream directly

rh101:
There is nothing particularly important about the support of the iostream classes, they just allow the use of any existing code that relies on std::iostream.  

I created it as a test case to see if it can work, so if it's not useful then there really is no need to merge in. :)  Please close this PR if that is the case.

halx99:
Yeah, since the FileStream API simlar to std::ofstream, we can use it directly, like follow:
```cpp
bool FileCopy(const std::string& src_loc, const std::string& dest_loc)
{
    auto in = FileUtils::getInstance()->openFileStream(src_loc, FileStream::Mode::READ);

    if (in)
    {
        auto out = FileUtils::getInstance()->openFileStream(src_loc, FileStream::Mode::WRITE);
        if(out) {
            const int BUFFER_SIZE = 4096;
            std::vector<char> buffer(BUFFER_SIZE, 0);
            int numberOfRead = 0;
        
            while ( (numberOfRead = in->read(buffer.data(), BUFFER_SIZE) > 0) )
            {
                out->write(buffer.data(), numberOfRead );
            }
        
            return true;
        }
    }

    return false;
}
```

rh101:
> Yeah, since the FileStream API simlar to std::ofstream, we can use it directly, like follow:

That's exactly how I have it implemented, and the method I used to modify for the iostream example. :)  

I'll close this PR since it's somewhat pointless.

PR #347, opened 2021-04-28
ImGUI font crash fix when using virtual file system

rh101:
Load fonts into IMGUI using memory routines rather than allowing it to use its own file handling routines.


halx99:
thanks

PR #363, opened 2021-05-29
Add Apple audio virtual file system support by using FileStream

rh101:
This ensures that the audio file is opened via FileStream.  This will allow for streaming audio data from memory, virtual files etc..


cocos/audio/src/AudioDecoderEXT.mm
```diff
@@ -158,4 +158,53 @@ of this software and associated documentation files (the "Software"), to deal
...
+        return noErr;
+    }
+
+    SInt64 AudioDecoderEXT::getSizeCallback(void *inClientData)
+    {
+        auto* fileStream = (cocos2d::FileStream*)inClientData;
+        //auto currentPos = fileStream->tell();
+        fileStream->seek(0, SEEK_END);
+        auto fileSize = (SInt64)fileStream->tell();
+        //fileStream->seek(currentPos, SEEK_SET); //go back to where we were
```
halx99, line 207: Why this line is commented?

rh101: Oh sorry I was testing if the file position needed to be set back to the original location before that getSize is called, but it seems that it's only ever called once at the start, so no need to reset the position again. I'll delete the commented code.

halx99:
Thanks, merged

PR #364, opened 2021-05-31
Cache stream size because Apple audio decoder requests the value too often

rh101:
This change will improve performance, because the total audio stream size should never change throughout audio playback.  The `AudioFileOpenWithCallbacks` `AudioFile_GetSizeProc` callback is constantly called throughout the duration of the audio playback, and if the get size function is processing intensive (such as using `seek`), then it would be a performance issue.

This change fixes it by caching the size as soon as the file stream is opened.

In addition to this, the `AudioFile_ReadProc` now seeks the minimum distance, rather that seeking from the start of the stream.


rh101:
@halx99 I think we should add a `GetSize()` method to `cocos2d::FileStream`, which would improve performance overall, given that the only way to get the size of the file stream at the moment is to call `seek(0, SEEK_END)` then `tell()`, but the specific stream may have more efficient ways to get the stream/file size.  What do you think?  If you agree then I can implement it and put in a separate PR.

halx99:
> @halx99 I think we should add a `GetSize()` method to `cocos2d::FileStream`, which would improve performance overall, given that the only way to get the size of the file stream at the moment is to call `seek(0, SEEK_END)` then `tell()`, but the specific stream may have more efficient ways to get the stream/file size. What do you think? If you agree then I can implement it and put in a separate PR.

I agree with you.

rh101:
Excellent! I'll create a new PR soon with the `FileStream::GetSize()` changes.

halx99:
> Excellent! I'll create a new PR soon with the `FileStream::GetSize()` changes.

did this pr could be merged or not?

rh101:
> 
> did this pr could be merged or not?

Yes please merge this, since it won't be affected by the `FileStream::GetSize()` changes, as caching the size value is still required.

PR #410, opened 2021-06-30
Chipmunk2D TestBed finished

aismann:
I think my merge was wrong. Maybe you can do it better (overwrite is ok)


aismann:
I understand the merge process now. Hope I did all ok.

halx99:
Good, I also wrote a simple guide to contribute to adxe project: https://github.com/adxeproject/adxe/discussions/411

Issue #432, opened 2021-07-16
Cmake OpenSSL error on Win32 


rh101:
- adxe version: current
- devices test on: 
- developing environments
   - NDK version:
   - Xcode version:
   - VS/msvctools version: Visual Studio 2019/14.28.29910
   - cmake version: 3.20.2

If CMake is run to create the build folder, then either the incorrect version of OpenSSL is detected if there is an existing installation, or an error appears if no OpenSSL is installed.  In order to fix this issue, the OPENSSL_ROOT variable needs to be set with the correct path to the OpenSSL libraries. 

This is an example of the error:
```
-- TLS configured to use openssl
CMake Error at C:/Program Files/CMake/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY) (found
  version "1.1.1k")
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  C:/Program Files/CMake/share/cmake-3.20/Modules/FindOpenSSL.cmake:570 (find_package_handle_standard_args)
```

PR #433, opened 2021-07-16
Set OPENSSL_ROOT to ensure the correct OpenSSL version is located

rh101:
This addresses the issue described in #432


halx99:
Thanks

PR #459, opened 2021-08-16
Update sprite sheet API to allow custom sprite sheet texture filename to be used

rh101:
This is to re-introduce a function that was left out during the sprite sheet changes:
`void addSpriteFramesWithFile(const std::string& spriteSheetFileName, const std::string& textureFileName, const std::string& format = "PLIST");'

I now remember why I initially removed it and forgot to re-add it, and that was because the compiler would have issues trying to distinguish between calls to the following because of the default value being present in the function declaration:

`void addSpriteFramesWithFile(const std::string& spriteSheetFileName, const std::string& format = "PLIST");`

and this:

`void addSpriteFramesWithFile(const std::string& spriteSheetFileName, const std::string& textureFileName, const std::string& format = "PLIST");`

The only thing I can think of at the moment is to simply change the method names to `addSpriteFramesWithFileAndTexture`, which also better reflects the purpose of the functions.  This will affect 2 functions:

`void addSpriteFramesWithFileAndTexture(const std::string& spriteSheetFileName, const std::string& textureFileName, const std::string& format = "PLIST");`

and

`void addSpriteFramesWithFileAndTexture(const std::string& spriteSheetFileName, Texture2D *texture, const std::string& format = "PLIST");`

Is this reasonable?

### DIFF
--- a/cocos/2d/CCSpriteFrameCache.cpp
+++ b/cocos/2d/CCSpriteFrameCache.cpp
@@ -78,7 +78,16 @@ SpriteFrameCache::~SpriteFrameCache()
 {
 }
 
-void SpriteFrameCache::addSpriteFramesWithFile(const std::string& spriteSheetFileName, Texture2D* texture, const std::string& format)
+void SpriteFrameCache::addSpriteFramesWithFileAndTexture(const std::string& spriteSheetFileName, const std::string& textureFileName, const std::string& format)
+{
+    auto* loader = getSpriteSheetLoader(format);
+    if (loader)
+    {
+        loader->load(spriteSheetFileName, textureFileName, *this);
+    }
+}
+
+void SpriteFrameCache::addSpriteFramesWithFileAndTexture(const std::string& spriteSheetFileName, Texture2D* texture, const std::string& format)
 {
     auto* loader = getSpriteSheetLoader(format);
     if (loader)
--- a/cocos/2d/CCSpriteFrameCache.cpp
+++ b/cocos/2d/CCSpriteFrameCache.cpp
@@ -151,7 +151,7 @@ void SpriteFrameCache::removeUnusedSpriteFrames()
     for (auto& iter : frames)
     {
         auto* spriteFrame = iter.second;
-        if (spriteFrame->getReferenceCount() <= 2) // A frame has a ref count of 1 when created, and 2 when added to cocos2d::Map. It is unused if the ref count is 2.
+        if (spriteFrame->getReferenceCount() == 1)
         {
             toRemoveFrames.push_back(iter.first);
             spriteFrame->getTexture()->removeSpriteFrameCapInset(spriteFrame);
--- a/cocos/2d/CCSpriteFrameCache.cpp
+++ b/cocos/2d/CCSpriteFrameCache.cpp
@@ -151,7 +151,7 @@ void SpriteFrameCache::removeUnusedSpriteFrames()
     for (auto& iter : frames)
     {
         auto* spriteFrame = iter.second;
-        if (spriteFrame->getReferenceCount() == 1)
+        if (spriteFrame->getReferenceCount() <= 2) // A frame has a ref count of 1 when created, and 2 when added to cocos2d::Map. It is unused if the ref count is 2.
         {
             toRemoveFrames.push_back(iter.first);
             spriteFrame->getTexture()->removeSpriteFrameCapInset(spriteFrame);
--- a/cocos/2d/CCSpriteFrameCache.cpp
+++ b/cocos/2d/CCSpriteFrameCache.cpp
@@ -78,36 +78,36 @@ SpriteFrameCache::~SpriteFrameCache()
 {
 }
 
-void SpriteFrameCache::addSpriteFramesWithFileAndTexture(const std::string& spriteSheetFileName, const std::string& textureFileName, const std::string& format)
+void SpriteFrameCache::addSpriteFramesWithFile(const std::string& spriteSheetFileName, const std::string& textureFileName, uint32_t spriteSheetFormat)
 {
-    auto* loader = getSpriteSheetLoader(format);
+    auto* loader = getSpriteSheetLoader(spriteSheetFormat);
     if (loader)
     {
         loader->load(spriteSheetFileName, textureFileName, *this);
     }
 }
 
-void SpriteFrameCache::addSpriteFramesWithFileAndTexture(const std::string& spriteSheetFileName, Texture2D* texture, const std::string& format)
+void SpriteFrameCache::addSpriteFramesWithFile(const std::string& spriteSheetFileName, Texture2D* texture, uint32_t spriteSheetFormat)
 {
-    auto* loader = getSpriteSheetLoader(format);
+    auto* loader = getSpriteSheetLoader(spriteSheetFormat);
     if (loader)
     {
         loader->load(spriteSheetFileName, texture, *this);
     }
 }
 
-void SpriteFrameCache::addSpriteFramesWithFile(const std::string& spriteSheetFileName, const std::string& format)
+void SpriteFrameCache::addSpriteFramesWithFile(const std::string& spriteSheetFileName, uint32_t spriteSheetFormat)
 {
-    auto* loader = getSpriteSheetLoader(format);
+    auto* loader = getSpriteSheetLoader(spriteSheetFormat);
     if (loader)
     {
         loader->load(spriteSheetFileName, *this);
     }
 }
 
-void SpriteFrameCache::addSpriteFramesWithFileContent(const Data& content, Texture2D* texture, const std::string& format)
+void SpriteFrameCache::addSpriteFramesWithFileContent(const Data& content, Texture2D* texture, uint32_t spriteSheetFormat)
 {
-    auto* loader = getSpriteSheetLoader(format);
+    auto* loader = getSpriteSheetLoader(spriteSheetFormat);
     if (loader)
     {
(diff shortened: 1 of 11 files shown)


halx99:
> This is to re-introduce a function that was left out during the sprite sheet changes:
> `void addSpriteFramesWithFile(const std::string& spriteSheetFileName, const std::string& textureFileName, const std::string& format = "PLIST");'
> 
> I now remember why I initially removed it and forgot to re-add it, and that was because the compiler would have issues trying to distinguish between calls to the following because of the default value being present in the function declaration:
> 
> `void addSpriteFramesWithFile(const std::string& spriteSheetFileName, const std::string& format = "PLIST");`
> 
> and this:
> 
> `void addSpriteFramesWithFile(const std::string& spriteSheetFileName, const std::string& textureFileName, const std::string& format = "PLIST");`
> 
> The only thing I can think of at the moment is to simply change the method names to `addSpriteFramesWithFileAndTexture`, which also better reflects the purpose of the functions. This will affect 2 functions:
> 
> `void addSpriteFramesWithFileAndTexture(const std::string& spriteSheetFileName, const std::string& textureFileName, const std::string& format = "PLIST");`
> 
> and
> 
> `void addSpriteFramesWithFileAndTexture(const std::string& spriteSheetFileName, Texture2D *texture, const std::string& format = "PLIST");`
> 
> Is this reasonable?

Use Enum maybe better to avoid compiler mismatch invoke

rh101:
We can't pre-define the sprite sheet type IDs, so do you mean to just make it an uint32 value, and have the developer assign a number it to represent the format of the sprite sheet handler being registered?

For instance:
```
struct SpriteSheetFormatType {
    enum : uint32_t {
        PLIST = 0,
        CUSTOM = 1000
    }
};
```

That way a developer can register a custom format by just going:
```
format1 = SpriteSheetFormatId::CUSTOM + 1;
format2 = SpriteSheetFormatId::CUSTOM + 2;
```
etc..

halx99:
I prefer:
struct SpriteSheetFormat {
    enum : uint32_t {
        PLIST = 0,
        CUSTOM = 1000
    }
};

rh101:
Excellent, doing the changes now.

rh101:
I noticed some strange behaviour in the tests.  Each test case should be self-contained, and not affect another test, but it seems that when `TestCase::TestCase()` is run, it calls this:

```
Director::getInstance()->getTextureCache()->removeUnusedTextures();
SpriteFrameCache::getInstance()->removeUnusedSpriteFrames();
```

but those lines aren't removing all the associated resources of the previous test case, so you can get crashes in other test cases because of differences in expected values (such as pixel types not matching), since frames of the same name would still be in memory, and still linked to different textures.

Those calls should probably be reversed as well, to this:
```
SpriteFrameCache::getInstance()->removeUnusedSpriteFrames();
Director::getInstance()->getTextureCache()->removeUnusedTextures();
```

Since frames have references to textures.

I'll look more into it to see how best to clear the previous test case before the current one is initialised.

rh101:
OK, so I think the bug with not freeing unused frames is not new, and has been in there for years.

In this method:
`SpriteFrameCache::removeUnusedSpriteFrames()`

The check for an unused frame was this:
`if (spriteFrame->getReferenceCount() == 1)`

The problem with that is when a SpriteFrame is created, it has a ref count of 1, then when it's added to the `Map<std::string, SpriteFrame*> _spriteFrames;` collection, the reference count increases by 1, since Map is a special cocos2d container that retains the frames internally.  So, when frames are not used, but are still in the Map collection, they actually have a ref count of 2.  

The check needed to be changed to this to work correctly:
`if (spriteFrame->getReferenceCount() == 2)` 

or better yet just cover all possibilities:
`if (spriteFrame->getReferenceCount() <= 2)`

aismann:
The last fix (Fix incorrect reference count check in SpriteFrameCache::removeUnused) maybe will fix other errors too:
like this: https://github.com/adxeproject/adxe/issues/319

cocos/2d/CCSpriteFrameCache.cpp
```diff
@@ -142,7 +151,7 @@ void SpriteFrameCache::removeUnusedSpriteFrames()
     for (auto& iter : frames)
     {
         auto* spriteFrame = iter.second;
-        if (spriteFrame->getReferenceCount() == 1)
+        if (spriteFrame->getReferenceCount() <= 2) // A frame has a ref count of 1 when created, and 2 when added to cocos2d::Map. It is unused if the ref count is 2.
```
halx99, line 154: Fix original frame create code maybe better

halx99: The SpriteFrame::createWithTexture use autorelease, so it seems the referenceCount is correct and should be 1 when a spriteFrame unused, see:
https://github.com/adxeproject/adxe/blob/188345e42a68427e4ab347e3f7b4ba31dbedb007/cocos/2d/CCPlistSpriteSheetLoader.cpp#L224

rh101: That is so strange. I'll keep looking into it

rh101:
OK, so here is the original issue that I'd thought wasn't the problem, but it is:

The current sequence of events is this:
- TestScene1 created, calls Director::replaceScene
- TestScene1 becomes current scene
- Press button to go to next scene
- TestScene2 created, calls Director::replaceScene
- TestScene2 becomes new scene on next update
- TestScene1 destroyed

So, TestScene1 is still valid when TestScene2 is created, and has references to all nodes in that scene.  Those nodes (sprites and such) are holding on to the references to the frames, so they won't be removed.  Now, for example, if TestScene1 has a frame named "image123.png" from sprite sheet A, and TestScene2 has sprite sheet B with a frame called image123.png" as well, then that new frame will not be added to the frame list, because one still exists in it from sprite sheet A.

The primary issue is the fact that different sprite sheets are using the exact same frame names, and those sprite sheets are in memory at the same time, so there is no way to differentiate between the frames. 

In the original code, before the sprite sheet changes, these were the containers used:
```
Map<std::string, SpriteFrame*> _spriteFrames;
std::unordered_map<std::string, std::set<std::string>> _indexPlist2Frames;
std::unordered_map<std::string, std::string> _indexFrame2plist;
```
_indexFrame2plist would contain the frame->plist mapping, so same issue there.

In the current implementation, it's this:
```
Map<std::string, SpriteFrame*> _spriteFrames;
std::unordered_map<std::string, std::shared_ptr<SpriteSheet>> _spriteFrameToSpriteSheetMap;
```

I would assume developers explicitly load and release sprite sheets when not required, and that sprite frames are prefixed in some way, like sprite sheet 1 has `scene/image.png` and sprite sheet 2 has `scene2/image.png` etc, which avoids the name clashes, and is how I've personally been handling it.  I'm not sure though if other developers do this.  

As for the test scenes, there is no easy way to fix the issue pf the sprite sheets have exactly the same frame names inside and are both loaded at the same time.  Correct me if I'm wrong, but I'm certain the names would have clashed prior to the sprite sheet changes as well.

There is a way to fix the test scenes, but it has a side-effect for the Transition tests:

* Load TestScene1 and set as current scene
* Click next test
* Load placeholder scene (empty scene with a callback), and replace the current scene
* TestScene1 is unloaded
* Callback is called to create the new TestScene2
* TestScene2 replaces placeholder scene and becomes current scene

This order of events works, and would fix the issues, because TestScene1 is unloaded before TestScene2. The side-effect though is the impact it has on the scene transition tests, since they require the current scene to be valid when a new scene is being transitioned to.  For the scene Transition tests, it would have to be remain as-is, which is:

* Run TestScene1
* Click next
* Create TestScene2
* Transition to TestScene2 from TestScene1
* Free TestScene1

Note that this will only fix the Tests, since that is where the issue is.  The login in SpriteFrameCache is correct, and not the source of the problem.  The source of the problem is frame name clashes in sprite sheets, which can and should be avoided by ensuring either that sprite sheets with same name frames aren't loaded at the same time, or frame names are unique (prefixes are best, like `sheet1/image.png` and 'sheet2/image.png`).


Can anyone think of any other way to address this issue?

aismann:
Maybe a wiki entry is usefull?

rh101:
What information do you want covered in the wiki entry, and where should this be posted up?

aismann:
It only an idee:
As I understand is a test issue?
So the info why it ´'crash sporadically' can be part on Wiki (for developers maybe) 
I think autotest running on ccp-test will get the issue again, right?
So why not an info why?

rh101:
> As I understand is a test issue?
> So the info why it ´'crash sporadically' can be part on Wiki (for developers maybe)

Yes, it's how the tests work at the moment, because of the duplicated frame names in the sprite sheets.  

> I think autotest running on ccp-test will get the issue again, right?
Yes, that will most likely happen until the test resources are updated or the test code is tweaked to unload each test before loading the next (meaning no transition scenes).  It's very easy to reproduce it too. If you go into Sprite test, loop through the tests forward (pressing next etc), then go out and enter SpriteFrameCache test, and it'll definitely crash, because of non-unique frame names in loaded sprite sheets.

> So why not an info why?
I'm not sure that I understand what you mean.  We should have this documented somewhere, and I'm more than happy to do it, but which wiki are you referring to, because I've never written anything like that for this project?  Like where should the documentation be posted to?

aismann:
I understand:
The wiki is there:
https://github.com/adxeproject/adxe-wiki

rh101:
Thank you! I'll post a PR to it within the next few days with some documentation.

aismann:
Thanks. Thats great.

PR #462, opened 2021-08-18
cleanup and improvement

aismann:
[ADD] void drawSolidCircle(const Vec2& center, float radius, float angle, unsigned int segments, float scaleX, float scaleY, const Color4F& fillColor, float borderWidth, const Color4F& borderColor);

physics-nodes lib:
class CC_EX_DLL DebugDraw : public b2Draw  is "clean" now

Box2D - TestBed is improved again


aismann:
Small info:
How Box2D can used with adxe is done.
Latest cleanup will be done some time .

aismann:
Should be ready now (if all be green)

halx99:
Merged, thanks for your hard work for this improvment, Excellent

PR #466, opened 2021-08-19
Fixes for crashes in sprite and sprite cache tests

rh101:
This is a simple fix for the Sprite Tests and Sprite Cache Tests. The entire test suite can be run without the crashes resulting from pixel format differences etc.


aismann:
@rh101 
Maybe a hint about the correct use of sheets should be add at the wiki too.
The test doing it wrong...right?

rh101:
> Maybe a hint about the correct use of sheets should be add at the wiki too.

@aismann That's the plan, as I mentioned in the other post regarding this.  I'll write something up about it and post it to the wiki over the next few days.

> The test doing it wrong...right?

That is correct.

PR #468, opened 2021-08-19
Strange, build dll failed, visual studio bug?

halx99:
Static lib works, dynamic doesn't, caused by new custom spritesheet loader:
```cpp
std::map<uint32_t, std::uniqie_ptr<ISpriteSheetLoader>> loaders;
```

### DIFF
--- a/.github/workflows/pr-ci.yml
+++ b/.github/workflows/pr-ci.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Build
         shell: cmd
         run: |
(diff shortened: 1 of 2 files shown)


halx99:
@rh101 Could you take a look about this issue, the error caused by: https://github.com/adxeproject/adxe/blob/dev/cocos/2d/CCSpriteFrameCache.h#L291
And when I change it to ```std::map<uint32_t, std::shared_ptr<ISpriteSheetLoader>>```, works.

halx99:
Some other reference: https://www.py4u.net/discuss/89823

rh101:
Sorry, I didn't notice this issue earlier.  Yeah I get why `unique_ptr` is not working with a dynamic library, and there is no reason we can't use `shared_ptr` instead.

rh101:
Fix is in #471

PR #475, opened 2021-08-26
Win64 build support [DONE] 🚀🎉

halx99:
- [x] Node::sortNodes incorrect because endian check for win64 is not correct, will lead renderer problem, see fairygui-tests
- [x] Remove unused code from etc1 since we already use etc2 to decode etc1 RGB
- [x] Fix some warnings
- [x] win64 dll
- [x] Remove unused iconv.dll for windows
- [x] Add WIN64 detect at `thirdparty/cmake/CocosExternalConfig.cmake` and `cmake/Modules/CocosConfigDefine.cmake` like follow:
```cmake
if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
    set(WINDOWS TRUE)
    if("${CMAKE_GENERATOR_PLATFORM}" STREQUAL "Win64")
        set(WIN64 TRUE)
     else()
        set(WIN32 TRUE)
     endif()
# ...
endif()
```
- [x] Add `windows/x64` platform_spec_path at `thirdparty/cmake/CocosExternalConfig.cmake`
```cmake
if(IOS)
# ...
elseif(WINDOWS)
    set(platform_name windows)
    if(WIN64)
        set(platform_spec_path windows/x64)
    else()
        set(platform_spec_path windows/x86)
    endif()
# ...
endif()
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 adxe-1.0  ??
 
+[HIGHLIGHT] Windows x64 build support
 [HIGHLIGHT] Support custom texture atlas formats, thanks to @rh101
 [HIGHLIGHT] Downloader realtime md5 checksum calculation support
 [HIGHLIGHT] Decompress astc parallel support


halx99:
on windows default is x64, no need additional option

获取Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #481, opened 2021-08-28
Fix crash problem when load a non-truetype .ttf file

### LINKED ISSUE #480: Crash with some .ttf caused by #384
- adxe version:
- devices test on:
- developing environments
   - NDK version:
   - Xcode version:
   - VS/msvctools version: Visual Studio 2019/14.28.29910
   - cmake version: 
Steps to Reproduce:

1. 
2.
[Freeride.zip](https://github.com/adxeproject/adxe/files/7068295/Freeride.zip)
image

@rh101 Do you have time take a look about this issue?

halx99:
Fix #480
refer to: https://gitlab.freedesktop.org/freetype/freetype/-/blob/master/include/freetype/freetype.h#L1593

### DIFF
--- a/cocos/2d/CCFontFreeType.cpp
+++ b/cocos/2d/CCFontFreeType.cpp
@@ -2,8 +2,9 @@
 Copyright (c) 2013      Zynga Inc.
 Copyright (c) 2013-2016 Chukong Technologies Inc.
 Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
+Copyright (c) 2021 Bytedance Inc.
  
-http://www.cocos2d-x.org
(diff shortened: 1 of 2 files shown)


rh101:
The fix works fine, and the option to set the native byte code hinting to enabled or not is a good addition.  In the case of the app I'm working on, I had to make sure it's set to `true` in order to get the vertical spacing correct between lines using the TTF font.

halx99:
> The fix works fine, and the option to set the native byte code hinting to enabled or not is a good addition. In the case of the app I'm working on, I had to make sure it's set to `true` in order to get the vertical spacing correct between lines when using TTF fonts.

The default value is `true`

PR #501, opened 2021-09-17
Update README.md

aismann:


aismann:
Maybe the readme need no build check?

PR #514, opened 2021-10-04
Implement Windows WebView using WebView2 Edge Chromium

rh101:
Add download support to setup.py (code based on original cocos `download-deps.py`).
Add support for NuGet CLI.
Download NuGet Microsoft.Web.WebView2 package version 1.0.992.28.
Add support for WebView for Windows builds. Code based on project [https://github.com/webview/webview](https://github.com/webview/webview`) and old implementation of WebView in Cocos2d-x.

### DIFF
--- a/cocos/ui/UIWebView/UIWebViewImpl-win32.cpp
+++ b/cocos/ui/UIWebView/UIWebViewImpl-win32.cpp
@@ -1130,12 +1130,11 @@ bool Win32WebControl::createWebView(
         m_window = CreateWindow(L"webview", L"", WS_VISIBLE | WS_CHILD, 0, 0, 0, 0, hwnd, nullptr,
                                 hInstance, nullptr);
         SetWindowLongPtr(m_window, GWLP_USERDATA, (LONG_PTR)this);
-        SetWindowLongPtr(m_window, GWL_EXSTYLE, WS_EX_LAYERED);
 
         //SetProcessDpiAwarenessContext(DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE);
         ShowWindow(m_window, SW_SHOW);
         UpdateWindow(m_window);
-        //SetFocus(m_window);
+        SetFocus(m_window);
 
         auto cb = [this](const std::string msg) { on_message(msg); };
 
--- a/cocos/ui/UIWebView/UIWebViewImpl-win32.cpp
+++ b/cocos/ui/UIWebView/UIWebViewImpl-win32.cpp
@@ -686,12 +686,14 @@ private:
         ULONG STDMETHODCALLTYPE Release() { return 1; }
         HRESULT STDMETHODCALLTYPE QueryInterface(REFIID riid, LPVOID* ppv) { return S_OK; }
 
+        // ICoreWebView2CreateCoreWebView2EnvironmentCompletedHandler
         HRESULT STDMETHODCALLTYPE Invoke(HRESULT res, ICoreWebView2Environment* env)
         {
             env->CreateCoreWebView2Controller(m_window, this);
             return S_OK;
         }
 
+        // ICoreWebView2CreateCoreWebView2ControllerCompletedHandler
         HRESULT STDMETHODCALLTYPE Invoke(HRESULT res, ICoreWebView2Controller* controller)
         {
             controller->AddRef();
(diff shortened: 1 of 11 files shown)


rh101:
In order to use the WebView on Windows, the runtime must be installed: [https://developer.microsoft.com/en-us/microsoft-edge/webview2/](https://developer.microsoft.com/en-us/microsoft-edge/webview2/)

rh101:
This implementation is working for me, and it seems to pass all the tests in the `cpp-tests` project.  If anyone has a specific requirement that they can add a test for, then please do so.  The only thing this implementation does not support is opacity changes, since the underlying WebView2 doesn't seem to support it.

halx99:
Good job, does this PR complete?

rh101:
It should be all done now, unless anyone has any suggestions or concerns about this implementation.

rh101:
Regarding the NuGet download into the `tools/nuget/` folder, do you think it's worthwhile changing that to be something like `tools/external/nuget`, and add the `tools/external/` path to the `.gitignore` file? Any downloaded tools can then go into that folder, and at least then it won't show up when you're using git commit, so won't accidentally be added to the repository etc.

halx99:
> Regarding the NuGet download into the `tools/nuget/` folder, do you think it's worthwhile changing that to be something like `tools/external/nuget`, and add the `tools/external/` path to the `.gitignore` file? Any downloaded tools can then go into that folder, and at least then it won't show up when you're using git commit, so won't accidentally be added to the repository etc.

Agree

cocos/ui/UIWebView/UIWebViewImpl-win32.cpp
```diff
@@ -0,0 +1,1308 @@
...
+        }
+    }
+    return encoded;
+}
+
+inline std::string url_decode(const std::string st)
+{
+    std::string decoded;
+    const char* s = st.c_str();
+    size_t length = strlen(s);
```
halx99, line 108: Don't needs use strlen, and the  null character check can be loop condition with length check.

cocos/ui/UIWebView/UIWebViewImpl-win32.cpp
```diff
@@ -0,0 +1,1308 @@
...
+
+// Convert ASCII hex string (two characters) to byte.
+//
+// E.g., "0B" => 0x0B, "af" => 0xAF.
+static inline char hex2char(const char* p)
+{
+    return hex2nibble(p[0]) * 16 + hex2nibble(p[1]);
+}
+
+inline std::string url_encode(const std::string s)
```
halx99, line 84: The parameter shoild be const reference

halx99:
The url encode and decode are useful API, can be moved to ccUtils

cocos/ui/UIWebView/UIWebViewImpl-win32.cpp
rh101: I'll update that.

cocos/ui/UIWebView/UIWebViewImpl-win32.cpp
rh101: That code is came from the [webview](https://github.com/webview/webview) project, so I left it as is.  It may have something to do with the fact that `std::string` can hold null values, so `std::string::size()` will return an incorrect size,  As far as I know, a null isn't valid values to encode in a URL.  It's up to you, I can change it so it uses `std::string::size()`.

rh101:
> 
> 
> The url encode and decode are useful API, can be moved to ccUtils

Good idea.

cocos/ui/UIWebView/UIWebViewImpl-win32.cpp
halx99: More safe code:
```cpp
const char* s = st.c_str();
size_t length = st.length();
for (unsigned int i = 0; i < length; ++i)
{
   if(![i]) break;
   if (s[i] == '%')
    {
        decoded.push_back(hex2char(s + i + 1));
        i = i + 2;
    }
    else if (s[i] == '+')
    {
        decoded.push_back(' ');
    }
    else
    {
        decoded.push_back(s[i]);
    }
}
```

rh101: Yes, that would stop the decoding if a null value is hit, since it shouldn't be there in the encoded string.  I'll update it now.

PR #544, opened 2021-11-30
Change all setString(const std::string & ...) ...

aismann:
... to setString(std::string_view ...)

  Change all methods inherited from virtual void setString(const std::string & label) = 0;

### DIFF
--- a/cocos/ui/UITextFieldEx.cpp
+++ b/cocos/ui/UITextFieldEx.cpp
@@ -182,7 +182,7 @@ namespace ui {
         return label->getContentSize().width;
     }
 
-    static std::string internalUTF8MoveLeft(const std::string& utf8Text, int length/* default utf8Text.length() */)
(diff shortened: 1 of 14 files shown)


cocos/ui/UITextFieldEx.cpp
```diff
@@ -949,7 +949,7 @@ namespace ui {
             if (!secureTextEntry)
                 displayText = &this->getString();
             else if(!this->inputText.empty())
-                displayText = &renderLabel->getString();
+                displayText = (const std::string*) &renderLabel->getString();
```
halx99, line 952: Since c++17, the code could be optimize to use `std::string_view instead std::string*`, I will optimize it after this PR merged.

halx99 (approved):
LGTM, thanks

halx99:
Merged, thanks

PR #547, opened 2021-12-02
Update CCConsole.cpp

aismann:
log only  if COCOS2D_DEBUG != 0

### DIFF
--- a/cocos/base/CCConsole.cpp
+++ b/cocos/base/CCConsole.cpp
@@ -126,6 +126,10 @@ namespace {
 
 void log(const char * format, ...)
 {
+#if COCOS2D_DEBUG == 0
+    return;
+#endif  // COCOS2D_DEBUG == 0
+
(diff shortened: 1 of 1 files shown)


halx99:
If I want log message in release mode, how?

aismann:
> 



> If I want log message in release mode, how?

set COCO2D_DEBUG=1 (hardcoded on your 'want to test code')
another idea:
Add some new log types like:
ADXE_TRACE
ADXE_WARNING
ADXE_ERROR
ADXE_LOG   (always log also on release version)

aismann:
Once more:
log(...) should used on release mode "ONLY" if there is an issue and rermove it when 'fixed'.
Cocos2dx has used it to much also on release mode without removing it again (I think).

halx99:
Log level:
- TRACE/DEBUG
- INFO
- WARN
- ERROR

aismann:
I remove this PR and will add a better one

PR #560, opened 2021-12-28
Merge ahead to dev

halx99:
- All string_view
- Improve JNI call
- robin_map for all string map with Heterogeneous lookup support
- ci improvements
- Use std::filesystem implement listFiles on supported build system
- Optional windows webview2 with cmake variable `ADXE_ENABLE_WEBVIEW2`, by default: enabled


halx99:
Will merge later, now the default branch `release` instead `dev` for stable reason

Issue #565, opened 2022-01-01
CCLabel overflow shrink bug


DelinWorks:
There is a bug in cocos2d::Label that can be fixed easily, when making a Label overflow shrink the size of the label after being shrinked from a big text doesn't go back to it's original font size after setting the text to something smaller

In another wording if shrink shrinks the text the label stays small forever even if you put smaller text in it

- adxe version: 1.0.0b2 (Old version but bug still applies)
- devices test on: windows 10

I'm kind of noob at making GitHub pull requests but ill try, if you can help I'll appreciate it :D

Steps to Reproduce:

1. make a new adxe windows project
2. create a new label this way and add it to the scene
```
loadingText = Label::createWithTTF("Validating Resources...", "fonts/SourceCodePro-Regular.ttf", 22, Size(600, 40), TextHAlignment::CENTER);
loadingText->setAnchorPoint(Vec2(.5f, 1));
loadingText->setOverflow(Label::Overflow::SHRINK);
```

Keep manipulating the text when some keys are pressed and you'll notice the bug right off the bat

I fixed it by going to CCLabel.cpp in the engine's source and modifying setString() function and adding restoreFontSize() built-in function when the text changes so that it becomes normal again and for shrink function to modify it however it wants without "locking" it to a small font size

```
void Label::setString(const std::string& text)
{
    if (text.compare(_utf8Text))
    {
        restoreFontSize();

        _utf8Text = text;
        _contentDirty = true;

        std::u32string utf32String;
        if (StringUtils::UTF8ToUTF32(_utf8Text, utf32String))
        {
            _utf32Text  = utf32String;
        }
    }
}
```

rh101:
Isn't `restoreFontSize` an expensive call to make, especially in `setString()`, which may be called quite a lot in certain scenarios, for instance, where a label displays a counter etc.

When the overflow is set to `Overflow::SHRINK`, `shrinkLabelToContentSize` seems to be the section of code that handles everything to shrink the displayed text (I only had a quick look at it).  Have you looked into any alternatives to solving this issue without having to call `restoreFontSize` on every `setString`?

PR #566, opened 2022-01-01
Label overflow shrink bug fix

DelinWorks:
This is the fix for the issue #565

### DIFF
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -929,6 +929,7 @@ void Label::setString(const std::string& text)
 {
     if (text.compare(_utf8Text))
     {
+        // issue #565 Label overflow shrink bug
         restoreFontSize();
      
         _utf8Text = text;
--- a/cocos/2d/CCLabel.cpp
+++ b/cocos/2d/CCLabel.cpp
@@ -929,6 +929,8 @@ void Label::setString(const std::string& text)
 {
     if (text.compare(_utf8Text))
     {
+        restoreFontSize();
+     
         _utf8Text = text;
         _contentDirty = true;
 


halx99 (approved):

halx99:
Thanks, what about invoke `restoreFontSize()` when overflow changed?
The setString maybe invoked frequently

DelinWorks:
I think overflow changes all the time whenever you change the text it has to check at some point
and setString only gets invoked when string is different so there is no real issue to it being in setString

halx99:
Can you write a test case and post some screenshot to reproduce the issue?
Bye the way, you can fetch and update latest from `dev` and create a new PR to avoid conflicting.
You can refer to https://github.com/adxeproject/adxe/discussions/411

rh101:
> setString only gets invoked when string is different so there is no real issue to it being in setString

As I mentioned in the comment I posted in the issue thread, there are instances where a label is updated a lot, such as when using it to display something like a counter.  That would force a call to `restoreFontSize()` on every single `setString()`, and from the look of that method, it may have an impact on performance.

DelinWorks:
okay ill do the best in finding a better solution, if i have the time ill do a quick performance testing with and without the function

DelinWorks:
I did a test and the results were bad and I mean BAD

10000 setString calls with random text with restoreFontSize function takes is long is 1000ms
but doing the same without calling restoreFontSize takes as little as 24ms

I have thought of
```
void Label::setString(const std::string& text)
{
    if (text.compare(_utf8Text))
    {
        if (_overflow != Overflow::NONE)
            restoreFontSize();

        _utf8Text = text;
        _contentDirty = true;

        std::u32string utf32String;
        if (StringUtils::UTF8ToUTF32(_utf8Text, utf32String))
        {
            _utf32Text  = utf32String;
        }
    }
}
```

But i dont trust it.. ill think of something else

DelinWorks:
This is the closest I can get to

```
void Label::setString(const std::string& text)
{
    if (text.compare(_utf8Text))
    {
        if (_overflow == Overflow::SHRINK && getRenderingFontSize() < _originalFontSize)
            restoreFontSize();

        _utf8Text = text;
        _contentDirty = true;

        std::u32string utf32String;
        if (StringUtils::UTF8ToUTF32(_utf8Text, utf32String))
        {
            _utf32Text  = utf32String;
        }
    }
}
```

every time I try to restore or rescale when rendering font is lower than original in contentUpdate some how crashes

so I've put it in setString because it doesn't get called in visit or draw, first we check if the overflow is shrink if so then we see if the rendering font size is smaller than the original font size that the object was created on. (rendering font size being the font size of the label after shrink has modified it and it gets restored back to the original font size when ever you call restoreFontSize thus breaking this condition and giving us no performance impact). If so we restore the font size, I've ran the tests again and it yields pretty good results (24ms with and without this condition on 10000 tries). Now restoreFontSize is called rarely even if you call setString multiple times, Also since shrink modifies and recalculates the bounds again restoreFontSize puts it back, since you're losing performance one way or another from shrink.

So what do you think?

DelinWorks:
If somehow you don't know the issue in the first place here are some screen shots

Before Adding `restoreFontSize()`
image
image

This is of course wrong 'short :>' should look like this

image
I achieved this by adding the `restoreFontSize()` above of course

and yes `restoreFontSize()` doesn't ruin shrink the text still gets smaller is the text gets bigger like this:
image

rh101:
> So what do you think?

Calling `restoreFontSize()` from `setString()` is perhaps not the correct fix for several reasons, most importantly being the severe performance drop, as you can see from your own testing.  Another reason is that it seems more like it's being fixed in the wrong place, or rather in the wrong way.  

The important part to look at is either in `alignText` or in `shrinkLabelToContentSize()`, and focusing on what happens in there is most likely the key to fixing this issue.  

`scaleFontSizeDown` eventually gets called with a new font size, but it doesn't change the configured font size, but rather changes an internal value that is used to render out the text.  Side not, that method should probably be renamed to "scaleFontSize()`, since it doesn't only scale down.

Now, look at what happens in `setOverflow()`.  When that method is called, the label is forced to re-render if the overflow type changes, and it doesn't call `restoreFontSize()` to do this, instead it calls `rescaleWithOriginalFontSize()`:

```
void Label::rescaleWithOriginalFontSize()
{
    auto renderingFontSize = this->getRenderingFontSize();
    if (_originalFontSize - renderingFontSize >= 1)
    {
        this->scaleFontSizeDown(_originalFontSize);
    }
}
```

In this case, `scaleFontSizeDown()` is actually scaling the font up (the function name is completely misleading).  Calling `rescaleWithOriginalFontSize` may also not be the correct fix.


The reason the font is shrunk has to do with `_textDesiredHeight`:

```
bool Label::isVerticalClamp()
{
    if (_textDesiredHeight > _contentSize.height)
    {
        return true;
    }
    else
    {
        return false;
    }
}
```

This gets called from `Label::alignText()`, which is where it does the check for  `Overflow::SHRINK`:

```
if (_overflow == Overflow::SHRINK)
{
    float fontSize = this->getRenderingFontSize();

    if (fontSize > 0 && isVerticalClamp())
    {
        this->shrinkLabelToContentSize(CC_CALLBACK_0(Label::isVerticalClamp, this));
    }
}
```

The problem is how to ensure `_textDesiredHeight` contains a value based on the original font size, and not the current font size, which is what is happening at the moment.  The current font size is based on the last loop where it may have already been shrunk down, which results in the text display problem you reported.

I don't have time at the moment to deep dive into this, but I'll see what I can do over the next few days, unless you manage to figure out a proper fix for this issue in the meantime.    One thing we need to agree on is that `restoreFontSize` is not the correct fix for this issue, and more like a blanket call that just happens to fix the issue.

DelinWorks:
yeah you're right, I'll think of something else

rh101:
I just thought of something, and tried it out, and the results are promising.  In `Label::visit()`, you can add something like this:

```
if (_overflow == Overflow::SHRINK && this->getRenderingFontSize() < _originalFontSize)
{
    rescaleWithOriginalFontSize();
}
```

More specifically:

```
void Label::visit(Renderer* renderer, const Mat4& parentTransform, uint32_t parentFlags)
{
    if (!_visible || (_utf8Text.empty() && _children.empty()))
    {
        return;
    }

    if (_systemFontDirty || _contentDirty)
    {
        // Restore the font back to the original size if it is smaller (given we only ever shrink the output)
        if (_overflow == Overflow::SHRINK && this->getRenderingFontSize() < _originalFontSize)
        {
            rescaleWithOriginalFontSize();
        }

        updateContent();
    }
...
}
```

With 1000 loops of the following:
```
auto start = std::chrono::high_resolution_clock::now();
for (size_t i = 0; i < 1000; i++)
{
    label1->setString(std::to_string(i));
}
auto stop = std::chrono::high_resolution_clock::now();
auto duration = std::chrono::duration_cast<std::chrono::microseconds>(stop - start);

CCLOG("Duration: %d", duration.count());
```

(the following numbers based on my specific hardware setup, but the differences should be obvious with any hardware)
Duration of default code with problem: ~1173
Duration with the `restoreFontSize()` fix: ~59540
Duration with the new suggested fix in `Label::visit()`: ~1202

If you have time, check if this fix works for you.

DelinWorks:
weird.. I tried your exact solution but it kept crashing now it isn't, I guess I've done something wrong, it's working now flawlessly :D

also isn't it the same as putting these lines of code in setString() because setString() makes the content dirty anyways and you're going to rescale the text anyways, so performance isn't gonna be any different if you put it in setString or visit.

So should I do a new pull request now? I'm kinda new to this :D

rh101:
> also isn't it the same as putting these lines of code in setString() because setString() makes the content dirty anyways and you're going to rescale the text anyways, so performance isn't gonna be any different if you put it in setString or visit.

No, it's not the same, because it'll be called on every single `setString()`, and if it were in `setString()`, it results in this duration for the 1000 loop test:

```
Duration: 267572
```

Now that you mention it, if I move `restoreFontSize()` to `Label::visit()`, it also fixes the issue, and gives these results on multiple runs of the test:

```
Duration: 1291
Duration: 1233
Duration: 1149
Duration: 1177
```

So, the source of the performance issue is because of applying the fix in `setString()`.

rh101:
> So should I do a new pull request now? I'm kinda new to this :D

You don't need to create a new pull request, and you shouldn't, because it just confuses things.  Just update the current pull request.

halx99:
@DelinWorks Please update your branch first

- Goto your fork github page
- Click `Fetch upstream`

image

DelinWorks:
Okay I deleted my fork and forked it again lol

but dont worry its now updated i will make a new branch called "label-fix"

rh101:
> Okay I deleted my fork and forked it again lol

Oh no, you shouldn't have. Now it's going to create yet another new pull request.

DelinWorks:
Hold on wait

DelinWorks:
How do i create a new branch xD

DelinWorks:
cool 
image

rh101:
> How do i create a new branch xD

image

One way is to click the dropdown, and just type the name you want in the editbox, and press enter.

DelinWorks:
Like this?

image

rh101:
> Like this?

Yes.

DelinWorks:
Let's goo

PR #571, opened 2022-01-04
[Windows & Linux] Headless mode (Server mode) Feature

DelinWorks:
I tried to create a runnable server cli that clients can connect to on my linux machine but couldn't because the application needed an OpenGL window to work not a Cli window.

these changes check if there is a window and will not render if there is not (only do server side calculations like physics and hit validation and anti cheat and what not)

And will depend on a CLI whenever you launch the app in headless mode
And will also still call update() on scene and nodes

Normal:
image
 
CLI server mode:
if you dont create a window like in here

```
AllocConsole();
    freopen("CONIN$", "r", stdin);
    freopen("CONOUT$", "w", stdout);
    freopen("CONOUT$", "w", stderr);

    CCLOG("Headless mode (Server mode) Listening on port XXXXX");

    // initialize director
    auto director = Director::getInstance();
    auto glview   = director->getOpenGLView();
    if (!glview)
    {
#if (CC_TARGET_PLATFORM == CC_PLATFORM_WIN32) || (CC_TARGET_PLATFORM == CC_PLATFORM_MAC) || \
    (CC_TARGET_PLATFORM == CC_PLATFORM_LINUX)
        //glview = GLViewImpl::createWithRect("HelloCpp", cocos2d::Rect(0, 0, designResolutionSize.width, designResolutionSize.height));
#else
        glview = GLViewImpl::create("HelloCpp");
#endif
    }
....
```
then the application wont crash but just show a cute little server mode window because of the AllocConsole() command and without an OpenGL window

image

Of course these changes only work on windows and linux and were tested

### DIFF
--- a/core/platform/linux/CCApplication-linux.cpp
+++ b/core/platform/linux/CCApplication-linux.cpp
@@ -74,14 +74,16 @@ int Application::run()
     auto glview   = director->getOpenGLView();
 
     // Retain glview to avoid glview being released in the while loop
-    glview->retain();
+    if (glview)
+        glview->retain();
 
-    while (!glview->windowShouldClose())
+    bool flag = glview == nullptr ? false : glview->windowShouldClose();
+    while (!flag)
     {
         lastTime = getCurrentMillSecond();
 
         director->mainLoop();
-        glview->pollEvents();
+        if (glview) glview->pollEvents();
 
         curTime = getCurrentMillSecond();
         if (curTime - lastTime < _animationInterval)
@@ -94,13 +96,17 @@ int Application::run()
      *  when we want to close the window, we should call Director::end();
      *  then call Director::mainLoop to do release of internal resources
      */
-    if (glview->isOpenGLReady())
+    if (glview)
     {
-        director->end();
-        director->mainLoop();
-        director = nullptr;
+        if (glview->isOpenGLReady())
+        {
+            director->end();
+            director->mainLoop();
+            director = nullptr;
+        }
+        glview->release();
     }
-    glview->release();
+
     return EXIT_SUCCESS;
 }
 
--- a/core/platform/win32/CCApplication-win32.cpp
+++ b/core/platform/win32/CCApplication-win32.cpp
@@ -93,7 +93,8 @@ int Application::run()
     auto glview   = director->getOpenGLView();
 
     // Retain glview to avoid glview being released in the while loop
-    glview->retain();
+    if (glview)
+        glview->retain();
 
     LONGLONG interval = 0LL;
     LONG waitMS       = 0L;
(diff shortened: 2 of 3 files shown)


halx99:
Why needs this feature as game engine, just create simple console app for server maybe better

DelinWorks:
But then how can you make a robust game server that handles anti cheat and calculates physics accurately?

halx99:
Write game logic as shared module invoke by client & server

DelinWorks:
But then it becomes harder to maintain while you can just compile once and work as client or server on a single executables

halx99:
@rh101 What do you think about this feature?

rh101:
My opinion regarding this is that this game engine is intended for client devices, and adding functionality like this can create complexity and possible sources of problems (@DelinWorks I'm not implying that your changes cause these things).  

@DelinWorks Also, as you've discovered, nothing prevents you from modifying the code to do anything you want with it, but this is a very specific use-case, where the changes are unlikely to have any use for other users of this engine, so it's best to keep this out of this repository, and keep it in your own fork/branch.   

> But then it becomes harder to maintain while you can just compile once and work as client or server on a single executable

Merging those changes with the root project means transferring the maintenance over to the maintainers here for functionality that is unlikely to ever be used by anyone else.  Again, this functionality is very specific to how you want to do things, so the responsibility lies with you to maintain your own version of it.  

On a side note, from what I've worked on in the past, the game engine is not put on the server, but rather a more specialized and considerably more efficient server application is used to handle such functionality.  Just because something seems convenient, it doesn't mean that it's the right thing to do.

DelinWorks:
Yeah you're right, but since I'm changing my own fork to contain my code how can github know that these changes relate to me and not the original repo when I do fetch upstrem?

rh101:
> Yeah you're right, but since I'm changing my own fork to contain my code how can github know that these changes relate to me and not the original repo when I do fetch upstrem?

Well, when you fork, it creates a references to the upstream repository.  You then create a branch that represents your custom version (never make the changes in the existing branches used by the upstream repository, for many reasons).  When you fetch the changes from the upstream repository, it will update the existing branches in your fork.  You then create a pull request between your updated branch and the custom branch you have.  If git detects no merge issues, then you simply do a merge, and your own custom changes won't be affected.  The only time git will complain is if the same section of code that you changed has been also updated in the upstream branch, and git can't automatically figure out how to merge it.  Does that make sense?

DelinWorks:
Yes, thanks for telling me!

rh101:
> Yes, thanks for telling me!

It may be beneficial to get to know more about git.  You can do all of this on your own local system too via the CLI or any other GUI, not just github.  In the end Github is a pretty wrapper around git (well, with the bonus of issue tracking).

rh101:
> also I think game engines like unity do this. you can specify a server mode build which would only run a cli,

I had no idea Unity supported that, and I just took a quick look at how it works. It simply detaches the renderer, but all scene objects still exist, and still use up both resources and processing power, which would result in added costs if you're running this using some cloud infrastructure where you pay for such things.

DelinWorks:
Yep, but you can decide if you want to use a server like this that still uses objects and calculations or you can use a simple echo server which would just send player data to each other (no calculations nor anti cheat protection)

PR #583, opened 2022-02-09
Refine Readme English Grammar

DelinWorks:


README.md
```diff
@@ -22,39 +22,39 @@
...
 * IOS SDK 9.0 as minimal deployment
 * Use fast pugixml
 * Use [curl](https://github.com/curl/curl) for transferring data with URL syntax
-* Use SAX parser for all plist file
+* Use SAX parser for all plist files
 * Spine-3.8 support
 * Extension ```FairyGUI``` support
-* ASTC 4x4/6x6/8x8 support (if hardware decoder not present, use software decoder)
-* ETC2 RGB/RGBA support    (if hardware decoder not present, use software decoder)
+* ASTC 4x4/6x6/8x8 support (if hardware decoding is not present, software decoding will be used instead)
```
rh101, line 47: This isn't accurate, and the original is actually correct:

"decoding" is the action that the decoder will carry out. if one exists.  It's checking for a hardware decoder, and not the act of decoding.

So, `(if hardware decoder not present, use software decoder)` is spot on compared to `(if hardware decoding is not present, software decoding will be used instead)`.

DelinWorks: It's actually right if you look at it the right way, in english whenever you use ing it doesnt have to be 100% the action of doing something in the present.. somtimes it just refers to the thing itself or the action in general, I learned this somewhere but don't actually remember.

rh101: I understand, and it makes sense.  

Regarding the original text, it is correct.  The statement is to indicate what action the software will take if a condition is not met.  That condition is the availability of a hardware decoder, or alternatively, the support for hardware decoding.

DelinWorks: So the right thing to say is 
```If hardware decoder is not present, software decoder will be used instead```
felt that the change is a bit wordy

rh101: Oh I understand your concern, which is the the `use software decoder` part of the sentence.  

This also works and is perhaps less wordy: "if hardware decoding is not supported, then software decoding is used"

DelinWorks: Yeah this is cool

PR #621, opened 2022-04-24
[Fix] Node Camera Masking Design Flaw

DelinWorks:
`setCameraMask()` makes the parent only set it's current children to it's camera mask *ONCE*, but if more children were to be added, they won't follow their parent's mask and the function `setCameraMask()` has to be called again.

This PR fixes this problem by making `addChild()` assign the parent's camera mask to it's child upon entering/arrival.

This is helpful because when you set a node to a specific mask, children that might enter this node's tree in the future will join the node's chosen mask.

This functionality intervenes with the `setCameraMask(mask, applyChildren)` applyChildren property because it will ignore it all the time, so a custom function called `applyMaskOnEnter(bool applyChildren)` is added to make this functionality optional.

### DIFF
--- a/core/2d/CCNode.cpp
+++ b/core/2d/CCNode.cpp
@@ -117,6 +117,7 @@ Node::Node()
     , _realColor(Color3B::WHITE)
     , _cascadeColorEnabled(false)
     , _cascadeOpacityEnabled(false)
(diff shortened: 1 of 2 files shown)


halx99:
Thanks

PR #629, opened 2022-05-03
[FIX] Motion streak vsync off crash

DelinWorks:
There is a bug in MotionStreak Node which makes it so that when vsync is off the game crashes.

The crash is caused by memory getting allocated infinitely because 1 devided by 0 is infinity (or int32 max limit to be real) (vsync being off is 0 fps btw), If vsync is off (meaning 0) then we will assume that the refresh rate that that monitor can draw is 240hz even though vsync is off (worst case), This will cause minor artifacts that have no harm on a monitor that's beyond 240hz, But it's better than a memory leak or a crash.

### DIFF
--- a/core/2d/CCMotionStreak.cpp
+++ b/core/2d/CCMotionStreak.cpp
@@ -104,7 +104,7 @@ bool MotionStreak::initWithFade(float fade, float minSeg, float stroke, const Co
     // Fix animation interval 0 because of vsync being off
     double interval = _director->getAnimationInterval();
     double fps      = 1 / interval;
-    if (roundf(fps) == (float)(1e+300 * 1e+300))
+    if (roundf(fps) == /* INFINITY */ (float)(1e+300 * 1e+300))
         fps = 60.0;
 
     _maxPoints = (int)(fade * fps) + 2;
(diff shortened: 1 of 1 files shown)


halx99:
Maybe make sure the result fps less than a limit value is better, such as 1000

DelinWorks:
but it drops performance you know, maybe 240 is good?

halx99:
> but it drops performance you know, maybe 240 is good?

240 is ok

core/2d/CCMotionStreak.cpp
```diff
@@ -101,7 +101,12 @@ bool MotionStreak::initWithFade(float fade, float minSeg, float stroke, const Co
     _stroke    = stroke;
     _fadeDelta = 1.0f / fade;
 
-    double fps = 1 / _director->getAnimationInterval();
+    // Fix #629 Motion streak vsync off crash
+    double interval = _director->getAnimationInterval();
+    double fps      = 1 / interval;
+    if (roundf(fps) == INFINITY)
```
halx99, line 107:
```suggestion
    if (roundf(fps) > 240.0)
```

DelinWorks: that might be a good idea actually, making 240 the soft limit for the streak

PR #642, opened 2022-05-24
Add HSV/HSL color spaces to the engine.

DelinWorks:
This PR adds a useful function where you define colors in a different way, using Hue Saturation Value concept, it brings the ability to create fascinating colors with it.

The way it's built is indestructive and backwards compatible.

A structure is used that is called `HSV` and can be initialized with the constructor that takes 4 arguments, `h, s, v, a = 1.0` or it can be set using the `rgb()`  function which does some calculation to convert `Color3B`, `Color4B`, `Color4F` to `HSV` and store the results in the `HSV` object.

To use that `HSV` object you should first convert it back to `Color3B`, `Color4B`, `Color4F` and you can simply do that by just constructing a new color object and passing a const reference of `HSV` to it.

Create hsv:
```
auto hsv = HSV(146.0, 0.19, 0.66);
```
or with color values
```
auto hsv = HSV(Color3B(243, 39, 129));
```

Use hsv:
```
Node->setColor(Color3B(hsv));
```
this can also work with the other two variants `Color4B`, `Color4F` with their alpha values taken from the `HSV` object.

### DIFF
--- a/core/base/ccTypes.cpp
+++ b/core/base/ccTypes.cpp
@@ -46,15 +46,6 @@ Color3B::Color3B(const Color4B& color) : r(color.r), g(color.g), b(color.b) {}
 
 Color3B::Color3B(const Color4F& color) : r(color.r * 255.0f), g(color.g * 255.0f), b(color.b * 255.0f) {}
 
-Color3B::Color3B(const HSV& hsv)
-{
-    float fR, fG, fB;
-    hsv.color(fR, fG, fB);
-    r = fR * 255.0F;
-    g = fG * 255.0F;
-    b = fB * 255.0F;
-}
-
 bool Color3B::operator==(const Color3B& right) const
 {
     return (r == right.r && g == right.g && b == right.b);
@@ -96,16 +87,6 @@ Color4B::Color4B(uint8_t _r, uint8_t _g, uint8_t _b, uint8_t _a) : r(_r), g(_g),
 Color4B::Color4B(const Color3B& color, uint8_t _a) : r(color.r), g(color.g), b(color.b), a(_a) {}
 
 Color4B::Color4B(const Color4F& color) : r(color.r * 255), g(color.g * 255), b(color.b * 255), a(color.a * 255) {}
-
-Color4B::Color4B(const HSV& hsv)
-{
-    float fR, fG, fB, fA;
-    hsv.color(fR, fG, fB);
-    r = fR * 255.0F;
-    g = fG * 255.0F;
-    b = fB * 255.0F;
-    a = hsv.a * 255.0F;
-}
 
 bool Color4B::operator==(const Color4B& right) const
 {
@@ -151,16 +132,6 @@ Color4F::Color4F(const Color3B& color, float _a) : r(color.r / 255.0f), g(color.
 Color4F::Color4F(const Color4B& color)
     : r(color.r / 255.0f), g(color.g / 255.0f), b(color.b / 255.0f), a(color.a / 255.0f)
 {}
-
-Color4F::Color4F(const HSV& hsv)
-{
-    float fR, fG, fB, fA;
-    hsv.color(fR, fG, fB);
-    r = fR;
-    g = fG;
-    b = fB;
-    a = hsv.a;
-}
 
 bool Color4F::operator==(const Color4F& right) const
 {
@@ -312,7 +283,7 @@ HSV::HSV(const Color3B& c)
     float fR = c.r / 255.0F;
     float fG = c.g / 255.0F;
     float fB = c.b / 255.0F;
-    rgb(fR, fG, fB, 1.0F);
+    set(fR, fG, fB, 1.0F);
 }
 
 HSV::HSV(const Color4B& c)
@@ -321,7 +292,7 @@ HSV::HSV(const Color4B& c)
     float fG = c.g / 255.0F;
     float fB = c.b / 255.0F;
     float fA = c.a / 255.0F;
-    rgb(fR, fG, fB, fA);
+    set(fR, fG, fB, fA);
 }
 
 HSV::HSV(const Color4F& c)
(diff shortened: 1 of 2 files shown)


rh101:
Does it make sense for the HSV code to be part of the Color3/4 classes? Those classes form the base of all color usage in the engine, and they have a very specific meaning.  Adding methods into those classes that don't directly related to their functionality is polluting the API of those classes.

Also, are many other color spaces besides HSV, so HSV methods are for a very specific use case.  For instance, I personally make use of HSL and a few other color spaces instead of HSV.

It would even make more sense for the HSV class to convert to and from Color3/4, rather than putting HSV methods in the Color 3/4 classes.

The other option is to use specific conversion code that does all of this outside of those classes.

DelinWorks:
yeah I've thought about it this way, every feature that manipulates colors dont actually plant themselves in the color classes.

DelinWorks:
Yeah this seems like a more elegant solution:
```
ps->setStartColor(HSV(20, .3, 1, 1).toColor4F());
ps->setEndColor(HSV(160, .8, 1, 1).toColor4F());
```

Thanks @rh101 for the suggestion.

rh101:
That really isn't the responsibility of the HSV code.  When you pass values into it, they should already be within the expected range, and if not, it should trigger an error.

For instance, h is 0 - 360, s is 0-1.0, and v is 0 - 1.0.  Those are what your HSV class expects, and any code using them shouldn't be expected to fix the input.  If anything, it should output an error to the console, and cap the value to the expected range, but that capped value doesn't need to be the correct hue.  The value should just be in the valid range so that doesn't cause it to crash etc..

So, the calling code should be the that ensures the value is in the correct range.  

For conversion code, like RGB to HSV, there is a chance you get a negative h, and I can see from your code that you've already accounted for it:

    if (h < 0)
    {
        h = 360 + h;
    }

Now, that negative value will never be below -360, so that's why it works.  In terms of code passing values directly to HSV, there is no way you can factor in how large or small that value is, so it's really not the responsibility of the HSV implementation to adjust it, but rather the responsibility of the caller to ensure the values are valid.

Now, in terms of how to fix the value before you pass it to the HSV object, see if this works:

```
if (h < 0)
{
    h =  -(std::fabs(h) % 360); // Ensure number is not less than -360
    h += 360; // add 360 to make it positive
}
else if (h >= 360)
{
    h %= 360;
}
```

DelinWorks:
No totally the code is being taken care of inside `HSV` and `HSL` classes themselves, it's just nice to have the angle wrapped when you're doing animations and stuff so that you don't have to make sure that it's between 0 and 360

DelinWorks:
```
h = -(std::fabs(h) % 360); // result is h = -10
h += 360; // result is h = 350, which is what you expect
```

what if its -370?

rh101:
> what if its -370?

I already updated my post with that too a moment before you posted the question ;)

DelinWorks:
Sorry I didn't read it

DelinWorks:
Tho thanks for the reply! problem fixed :D

rh101:
No problem. Check what impact the call to `std::fabs` has if you're passing a lot of negative numbers to it.  I'm not sure how expensive that call is.

DelinWorks:
It works! but I had to modify the code because `%` doesn't work with decimals
```
float hue = -(remainder(std::fabs(h), 360));
hue += 360;
```

image

which shows a blue when passing `-120` is the hue

image

DelinWorks:
> No problem. Check what impact the call to `std::fabs` has if you're passing a lot of negative numbers to it. I'm not sure how expensive that call is.

No worries I'll test it.

rh101:
> It works! but I had to modify the code because `%` doesn't work with decimals

Oh oops! I completely forgot about the fact that the % only works for integers!

DelinWorks:
What is that name anyways `fabs` lol

rh101:
> What is that name anyways `fabs` lol

absolute for float values, so float abs = fabs :)

rh101:
> it takes the same time to compute negative and positive numbers is that what you mean?
> 

Yes, excellent!

DelinWorks:
Noice

core/base/ccTypes.cpp
```diff
@@ -388,16 +435,16 @@ void HSV::rgb(float fR, float fG, float fB, float fA)
...
 
-    // Make hue value wrap around 360 (i.e. -390 degrees would be 30 wrapped)
-    h = abs(fmod(h, 360.0F));
-
     a = fA;
 }
 
-void HSV::color(float& fR, float& fG, float& fB) const
+void HSV::get(float& fR, float& fG, float& fB) const
 {
```
rh101, line 442: You should avoid prefixing the parameter names with the type, as in "fR" etc.. There is no need for it, and it's actually bad practice to do so.

Just call them r, g, b, or red, green and blue.

DelinWorks:
Ok, ready! is there anything else missing?

core/base/ccTypes.cpp
DelinWorks: ok.

DelinWorks:
why would `std::fabs` be slow? I think what it does under the hood is this:
```
if (value < 0.0F)
    value *= -1.0F;
```

rh101:
> @rh101 why would `std::fabs` be slow? I think what it does under the hood is this:
> 
> ```
> if (value < 0.0F)
>     value *= -1.0F;
> ```

That's probably the case, and it's always good to test methods that may be a called a lot (in this case, many particles etc.) to find any bottlenecks.

DelinWorks:
True.

DelinWorks:
True, That's why I made HSV optional because it's expensive for particles.

core/base/ccTypes.h
```diff
@@ -173,6 +174,106 @@ Color4F operator/(Color4F lhs, const Color4F& rhs);
...
+};
+
+HSL& operator+=(HSL& lhs, const HSL& rhs);
+HSL operator+(HSL lhs, const HSL& rhs);
+
+HSL& operator-=(HSL& lhs, const HSL& rhs);
+HSL operator-(HSL lhs, const HSL& rhs);
+
+HSL& operator*=(HSL& lhs, const HSL& rhs);
+HSL operator*(HSL lhs, const HSL& rhs);
```
halx99, line 268: The lsh can be const HSL&

core/base/ccTypes.h
```diff
@@ -173,6 +174,106 @@ Color4F operator/(Color4F lhs, const Color4F& rhs);
 Color4F& operator/=(Color4F& lhs, float rhs);
 Color4F operator/(Color4F lhs, float rhs);
 
+/**
+ * Hue Saturation Value color space composed of 4 floats.
+ * @since adxe-1.0.0b6
```
halx99, line 179:
```suggestion
 * @since adxe-1.0.0b7
```

core/base/ccTypes.h
```diff
@@ -173,6 +174,106 @@ Color4F operator/(Color4F lhs, const Color4F& rhs);
...
+HSV operator*(HSV lhs, float rhs);
+
+HSV& operator/=(HSV& lhs, const HSV& rhs);
+HSV operator/(HSV lhs, const HSV& rhs);
+HSV& operator/=(HSV& lhs, float rhs);
+HSV operator/(HSV lhs, float rhs);
+
+/**
+ * Hue Saturation Luminance color space composed of 4 floats.
+ * @since adxe-1.0.0b6
```
halx99, line 228:
```suggestion
 * @since adxe-1.0.0b7
```

halx99 (approved):

halx99:
looks good to me, merged.

DelinWorks:
No cpp-tests needed?

halx99:
if you want, you can create a new PR for cpp tests

DelinWorks:
K i'll look for somewhere to fit it in, maybe like sprite color test

PR #653, opened 2022-06-13
[PROPOSAL] Add camera zoom, near, far, fov, aspect features & fix camera anchoring.

DelinWorks:
cameras have been hard to work with because they don't have functions that ease the workflow, incase of zooming in my game it was achieved using a wanky way. I was storing the z position of the camera and then multiplying it by some factor so why not make it a built-in camera function? same goes for near, far planes and field of view and aspect ratio. making these properties non-destructive and can be updated easily seems nice.

In case of 3D projection it was easy to implement, `setZoom()` sets the z position of the camera using
`Director::getInstance()->getZEye()` and then multiplying it by some factor on runtime.

In case of 2D projection it was easy to implement too, `setZoom()` sets the size of the view matrix of the camera, but that makes it zoom around bottom left. to fix this, the orthographic view matrix has to be shifted by half of the visible size to make the camera have it's origin at 0,0 on initialization. the user doesn't have to deal with this change as it's dealt with in `CCCamera.cpp` on camera creation, Unfortunately this breaks compatibility because an orthographic camera gets initialized at half of the visible size now rather than 0,0 origin, If previous games were to set the camera position to 0,0 then the camera will be at the center of the view area (which is the intended behaviour) rather than being at top right of the view area. this can easily be fixed by just shifting the camera half the visible size, this is how cameras should operate from the beginning, having 0,0 as their origin and center of the view area rather than shifting the view matrix.

That's even the way godot game engine does it:

https://user-images.githubusercontent.com/45469625/174081797-83bc9f98-315c-4e0c-8fbe-bb584bedf087.mp4

The camera's anchor being at 0,0 and not some arbitrary value based on screen size.

And when the window size or the screen size somehow change `applyCustomProperties()` deals with this behaviour automatically and anchors the camera accordingly and with respect to the resolution policy too.


DelinWorks:
For anyone that's confused why you would create a 3D `Perspective` camera and a 2D `Orthographic` camera for a 2D view it's because adxe uses 3D `Perspective` projection by default for 2D views which looks like this IF an object's z position or x,y rotations where to be changed:

https://user-images.githubusercontent.com/45469625/173447918-db4afa50-4199-4ec8-b987-3b13aa019778.mp4

It looks like it's in 3D even though we are viewing it from a 2D view

On the other hand, 2D `Perspective` Doesn't show stretching like 'real life' it's just orthographic:

https://user-images.githubusercontent.com/45469625/173448041-648e23c1-26b2-4556-9165-fa8b995581c7.mp4

halx99 (approved):

DelinWorks:
@halx99 @rh101 Do guys you think it would make much more since that I inherit the functions: `setPosition()` `getPosition()` `setPosition3D()` `getPosition3D()` `setRotation()` `getRotation()` `setRotation3D()` `getRotation3D()` into `Camera` from `Node` for custom behaviors?
It would be backwards compatible but I feel reluctant about it.

DelinWorks:
Done! The camera system is much easier to work with now.

DelinWorks:
Done! The camera system is much easier to work with now.

DelinWorks:
Done! The camera node is much easier to work with now.

core/2d/CCNode.cpp
```diff
@@ -525,8 +525,8 @@ void Node::setPosition(float x, float y)
 
 void Node::setPosition3D(const Vec3& position)
 {
-    setPositionZ(position.z);
-    setPosition(position.x, position.y);
+    Node::setPositionZ(position.z);
```
halx99, line 528: Why add prefix `Node::` for these function call?

DelinWorks: Because Node class calls the override of the camera class when infact it should call it's base function

DelinWorks:
bruh turns out I can just shift the orthographic view matrix by half rather than doing all this useless stuff ._.

DelinWorks:
Unfortunately this breaks compatibility because an orthographic camera gets initialized at half of the visible size now rather than 0,0 origin, If previous games were to set the camera position to 0,0 then the camera will be at the center of the view area rather than being at top right of the view area. which can easily be fixed by just shifting the camera half the visible size, this is how camera should operate from the beginning, having 0,0 as their origin rather than shifting the view matrix.

DelinWorks:
Okay @halx99 it's complete you can merge now

PR #657, opened 2022-06-14
SimpleSnake example

aismann:
First shot


aismann:
Of course. But there have to be examples created first.
Hope SimpleSnake and maybe spome more others example will makes adxe more attractiv for beginners

DelinWorks:
Finally, it's been tiring to see that same cocos2d sprite in the middle lol

PR #662, opened 2022-06-16
Fix enum class confliction & clean up types.

DelinWorks:


DelinWorks:
@halx99 I think `TEXTURE_ALPHA_MASK` is fair and self explanatory.

PR #690, opened 2022-07-01
[BUG] Fix `getDefaultCamera()` returning nullptr on scene initialization.

DelinWorks:
The function `Camera::getDefaultCamera()` returns a null pointer when calling it inside `init()` function of the scene, which defeats the whole purpose of scene initialization. so it's been fixed in this PR.

### DIFF
--- a/core/2d/CCCamera.cpp
+++ b/core/2d/CCCamera.cpp
@@ -78,16 +78,12 @@ Camera* Camera::createOrthographic(float zoomX, float zoomY, float nearPlane, fl
 Camera* Camera::getDefaultCamera()
 {
     // camera nullptr scene init fix #690
-    auto scene        = Director::getInstance()->getRunningScene();
-    auto sceneInitRef = Director::getInstance()->getRunningSceneInitRef();
-    if (scene && scene == sceneInitRef)
-    {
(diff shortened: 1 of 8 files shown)


halx99:
Why not get default camera from scene directly before it been run by director.

DelinWorks:
oh true I could've used `this->_cameras[0]`, but I think `getDefaultCamera()` should be safe and not actually through nullptrs around.

halx99:
Your code looks a little hacking, maybe, delete the function from directory, or add some comment to it:
```cpp
// User should ensure current running scene not null
```

DelinWorks:
maybe a `CCASSERT` could work in explaining why the game crashed when using getDefaultCamera() is called?

DelinWorks:
> Your code looks a little hacking, maybe, delete the function from directory, or add some comment to it:
> 
> ```c++
> // User should ensure current running scene not null
> ```

I see

halx99:
> maybe a `CCASSERT` could work in explaining why the game crashed when calling getDefaultCamera()?

Agree.

DelinWorks:
It's actually `this->_defaultCamera()` but don't you think that that function should be moved to the scene class rather than camera?

PR #691, opened 2022-07-01
Try fix fps

halx99:
Before fix, macOS:
image

After fix:
image


aismann:
Sweet to see snake on Mac OS (I haven't one) ;)

Issue #695, opened 2022-07-01
FPS display position should be set by API


aismann:
FPS display position should be set by API
Reason: The default FPS display position (LEFTBOTTOM) is not always the best to read the FPS values.

Example:
```
            Director::getInstance()->setFPSPos(Director::FPSPosition::RIGHTMIDDLE);
```

```
    enum FPSPosition
    {
        TOP_LEFT       = 7,
        TOP_CENTER     = 8,
        TOP_RIGHT      = 9,
        CENTER_LEFT    = 4,
        CENTER         = 5,
        CENTER_RIGHT   = 6,
        BOTTOM_LEFT    = 1,
        BOTTOM_CENTER  = 2,
        BOTTOM_RIGHT   = 3
    } _FPSPosition;
```
Hint: the position/numbers be like the position on the numpad for easier remind ;)

Test on test_cpp:  19:EventDispatcher: 5:Label Receives Keyboard Event
![image](https://user-images.githubusercontent.com/8652787/176916292-56912429-85dd-4f2c-9497-b0bad6fdca54.png)

aismann:
PR is ready to push.

aismann:
Implemented with #696

PR #696, opened 2022-07-01
FPS display position should be set by API 

aismann:
see feature whish: #695

### DIFF
--- a/core/base/CCDirector.cpp
+++ b/core/base/CCDirector.cpp
@@ -1318,7 +1318,7 @@ void Director::setFPSPos(FPSPosition FPSposition)
 
     if (_displayStats && _FPSLabel && _drawnBatchesLabel && _drawnVerticesLabel)
     {
-        auto safeOri22gin          = getSafeAreaRect().origin;
+        auto safeOrigin          = getSafeAreaRect().origin;
         auto safeSize          = getSafeAreaRect().size;
         const int height_spacing = (int)(22 / CC_CONTENT_SCALE_FACTOR());
 
@@ -1386,9 +1386,9 @@ void Director::setFPSPos(FPSPosition FPSposition)
             break;
         }
 
-        _drawnVerticesLabel->setPosition(Vec2(0, height_spacing * 2.0f) + _fpsPosition);
-        _drawnBatchesLabel->setPosition(Vec2(0, height_spacing * 1.0f) + _fpsPosition);
-        _FPSLabel->setPosition(Vec2(0, height_spacing * 0.0f) + _fpsPosition);
+        _drawnVerticesLabel->setPosition(Vec2(0, height_spacing * 2.0f) + _fpsPosition + safeOrigin);
+        _drawnBatchesLabel->setPosition(Vec2(0, height_spacing * 1.0f) + _fpsPosition + safeOrigin);
+        _FPSLabel->setPosition(Vec2(0, height_spacing * 0.0f) + _fpsPosition + safeOrigin);
     }
 }
 
--- a/core/base/CCDirector.cpp
+++ b/core/base/CCDirector.cpp
@@ -1303,11 +1303,89 @@ void Director::createStatsLabel()
     _drawnVerticesLabel->setIgnoreContentScaleFactor(true);
     _drawnVerticesLabel->setScale(scaleFactor);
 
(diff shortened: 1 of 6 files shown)


core/base/CCDirector.cpp
```diff
@@ -1303,11 +1303,89 @@ void Director::createStatsLabel()
...
+
+// Set the FPS position on disply like the numbers on numpad
+void Director::setFPSPos(FPSPosition FPSposition)
+{
+    Vec2 _fpsPosition;
+    _FPSPosition = FPSposition;
+
+    if (_displayStats && _FPSLabel && _drawnBatchesLabel && _drawnVerticesLabel)
+    {
+        auto safeOrigin          = getSafeAreaRect().origin;
```
halx99, line 1317: the safeOrign not use?

aismann: it was allways 0 (on my maschine)

aismann: > the safeOrign not use?



> the safeOrign not use?

I want not remove it at the moment (maybe it is using on other platforms?

halx99: You must consider use safeOrign, otherwise the fps label can't be display on some android high level devices. at least in my phone the safeOrigin not zero.

aismann: The problem is: I cant test it.
Maybe you can fix it if there is a mistake (missing it)?

aismann: It I add it it can be also wrong.

halx99: OK, I will

aismann: Please change this snippet also on it:  (the 'no change part' makes sense)
```
void Director::setFPSPos(FPSPosition FPSposition)
{
    if (_FPSPosition == FPSposition) // no changes!
    {
       return;
    }
    static Vec2 _fpsPosition;
    _FPSPosition = FPSposition;

    if (_displayStats && _FPSLabel && _drawnBatchesLabel && _drawnVerticesLabel)
```

aismann:
I have test it on the windows test (there are also full screen device, it's not zero)
next 'fix' should be work also on full screen device, it's not zero, such as MI MIX4

halx99:
> I have test it on the windows test (there are also full screen device, it's not zero) next 'fix' should be work also on full screen device, such as MI MIX4

Good, I will test it later

aismann:
I make latest test and rename the enum FPSPositio (use same naming convention seen on others)
    enum FPSPosition
    {
        TOP_LEFT       = 7,
        TOP_CENTER     = 8,
        TOP_RIGHT      = 9,
        CENTER_LEFT    = 4,
        CENTER         = 5,
        CENTER_RIGHT   = 6,
        BOTTOM_LEFT    = 1,
        BOTTOM_CENTER  = 2,
        BOTTOM_RIGHT   = 3
    } _FPSPosition;

aismann:
Should be ok now:
Test on test_cpp: 19:EventDispatcher: 5:Label Receives Keyboard Event
Solution for  #695

DelinWorks:
> 

Maybe this enum should be in a global class because many parts of the engine could use this enum like anchoring and what not, something like `AnchorCorner` enum or `EdgePositioning` enum

halx99:
> Maybe this enum should be in a global class because many parts of the engine could use this enum like anchoring and what not, something like `AnchorCorner` enum or `EdgePosition` enum

Or `FPSPosType`?

halx99:
> Maybe this enum should be in a global class because many parts of the engine could use this enum like anchoring and what not, something like `AnchorCorner` enum or `EdgePosition` enum

Let's just move to namespace `cocos2d`

halx99:
> > I have test it on the windows test (there are also full screen device, it's not zero) next 'fix' should be work also on full screen device, such as MI MIX4
> 
> Good, I will test it later

I have test on my device, works

aismann:
> Maybe this enum should be in a global class because many parts of the engine could use this enum like anchoring and what not, something like `AnchorCorner` enum or `EdgePosition` enum

I think a new enum for all kind like  FPSPosition is not really clear  for developer.
Maybe @halx99 can think about it later?

aismann:
PR is finished now.

halx99 (approved):

Issue #699, opened 2022-07-03
Should we generate an global enum using for all (new) enums which be use the "logic" like a numpad position? 


aismann:
Maybe this enum should be in a global class because many parts of the engine could use this enum like anchoring and what not, something like AnchorCorner enum or EdgePosition enum.

```
/**
 * @brief NumPadPosType can be used for all situations where to set the position on object like the numbers on the numpad
 */
enum NumPadPosType
{
    BOTTOM_LEFT    = 1,
    BOTTOM_CENTER  = 2,
    BOTTOM_RIGHT   = 3, 
    CENTER_LEFT    = 4,
    CENTER         = 5,
    CENTER_RIGHT   = 6,
    TOP_LEFT       = 7,
    TOP_CENTER     = 8,
    TOP_RIGHT      = 9
};
```

Where is the right place (file) for this 'global' enum?

DelinWorks:
i think `ccTypes.h` is a good place

aismann:
Thanks. I will send a PR

aismann:
> i think `ccTypes.h` is a good place

Or a new file like: 'ccEnum.h' is a better place for the future?

DelinWorks:
that's actually a good idea, wanna do it or do i do it? you'd have to move every enum that the engine uses there

aismann:
Maybe you are more familiar with enums on adxe.
Please add you it.
Better you are not change the existing enums at the moment because of backwards compatibility.
The only one enum should be 'FPSPosition' because it is new.

DelinWorks:
> MaBetter you are not change the existing enums at the moment because of backwards compatibility.

This won't hurt backwards compatibility because the enums that are previously used are going to be exposed anyways but in another file.

aismann:
I have start it (PR with no new file: ccEnum.h).
I will push it now. Ok?

DelinWorks:
This issue can be closed now as it's dealt with in #702

aismann:
closed

PR #700, opened 2022-07-03
 Global NumPadPosType can be used for all situations where to set the position on object like the numbers on the numpad

aismann:


aismann:
see also #699

DelinWorks:
Wait so you want to create `ccEnums.h`?

DelinWorks:
Maybe rename the `NumPadPosType` enum to `PositionAnchor` more comprehensible.

aismann:
> Wait so you want to create `ccEnums.h`?

Im  in a hurrry. would only add the changes because it would be reduce the work.
But if it should be rename again so the  pr can be deleted and you start it with new.

DelinWorks:
yeah delete the PR i'll do it

aismann:
> yeah close the PR i'll do it

Thanks.

PR #702, opened 2022-07-03
[PROPOSAL] Move global enums to their own file.

DelinWorks:
Any enums that exist outside of a class in any source file should be moved to `base/ccEnums.h`. Otherwise, If they are in a class or a cpp file they should stay there.


aismann:
Is it a good idea to change the  setFPSPos()   to   setStatsAnchor();
setFPSPos() was selfdescribe, and using the global 'enum class AnchorPreset' should not have the name of it.

DelinWorks:
it's actually fine becuase the AnchorPreset is called like that in other game engines like unity and godot because when you make datatypes you shouldn't tie them to anything but make them generic types, this a more pratical way.

as for setFPSPos it was changed to setStatsAnchor to make it close to the function setDisplayStats, notice how they both use the word stats.

that seemed logical to me! do you want me to revert it?

DelinWorks:
it's actually fine becuase the AnchorPreset is called like that in other game engines like unity and godot because when you make datatypes you shouldn't tie them to anything but make them generic types, this a more pratical way.

as for setFPSPos it was changed to setStatsAnchor to make it close to the function setDisplayStats, notice how they both use the word stats.

that seemed logical to me! do you want me to revert it?

aismann:
I removed my question before you answer. I think its ok.
But if I use setFPSPos on games is easier to read.
setStatsAnchor is only a learning issue. for me.

aismann:
> setStatsAnchor to make it close to the function setDisplayStats,
  
different names are bad:
better?
setStatsAnchor and setStatsDisplay ?

DelinWorks:
Wouldn't that break compatibility?

aismann:
> Wouldn't that break compatibility?

I'm not sure but for backwards compatibility we have to ADD the new function like this:
```
 bool isDisplayStats() { #pragma message(": Warning: Deprecated function: Use isStatsDisplay()");return isStatsDisplay(); }
 bool isStatsDisplay() { return _displayStats; }
```

aismann:
> > setStatsAnchor to make it close to the function setDisplayStats,
> 
> different names are bad: better? setStatsAnchor and setStatsDisplay ?

And of course. Im not sure the name be better.

DelinWorks:
> > > setStatsAnchor to make it close to the function setDisplayStats,
> > 
> > 
> > different names are bad: better? setStatsAnchor and setStatsDisplay ?
> 
> And of course. Im not sure the name be better.

No it's actually better! also I think backwards compatibility shouldn't be worried about when It's a simple function name.

halx99 (changes requested):
Please delete the build file:
image

DelinWorks:
Okay @halx99 I'm done you can merge

Issue #688, opened 2022-06-30
[PROPOSAL] Change class name `Sprite3D` to `MeshRenderer`


DelinWorks:
I think that the name `Sprite3D` is unclear because it means a sprite in 3d space and not actually a 3d model being rendered and that's what got me and multiple people in the cocos forums confused, Don't you think the name of the node should be `MeshRenderer` or `ModelRenderer` or something else as to make it clear?

halx99:
I prefer MeshRenderer

solan-solan:
If I may, some meanings about this fix.
Cocos2d has decade time of developing, with stable well known api. Potentially it could relatively light to go from cocos2d to adxe, for guys who used to use it, since it does not supported any more.
Of cause adxe is more deep fork, which brings more own features. But do you really have something about 3d space rendering support in your roadmap in the near future?
If not, then such changes gives no more then some cosmetic/logical iproves with much bigger pain to customize existing projects. These changes good after quality step, otherwise they looks like hasty.

halx99:
Use `typedef` to mark old name deprecated maybe better

solan-solan:
filenames were changed also

solan-solan:
User can include Sprite3D.h if he desire to extend its behavior for his project, it's c++ after all

halx99:
That also easy to solve it， create a empty file named Sprite3D.h in folder 3d and add one line code to include ccmeshrenderer.h

获取Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

solan-solan:
Ok, thanks for advise, I will do (material 3d is needed also for me)

I only wanted to notice, that everything is good when it is in time :) , now there were no visible reasons for such deviations from cocos api (imho of course)

halx99:
Do you work for `https://github.com/solan-solan/HeightMap`, I can help you

solan-solan:
This is example of using some technics, which I use in my more bigger project which supposed to be like rpg/action game in openworld (no so big like AAA projects of course, but it should include adventure/fight/quest/dialogs). The example includes postprocessing implementation base on render to texture which is provided by adxe/cocos and base building of 3d world space. It should represented that adxe/cocos are good choice for simple 3d games which are more interesting to my like.
And now, since cocos is not supported, I hope more on adxe engine.
I see how busy you are on engine support.
What help do you mean? I believe, that I can fix my things, but it happened a little unexpected

halx99:
[...]

PR #705, opened 2022-07-04
[IMPROVMENT] Change class name `Sprite3D` to `MeshRenderer`

DelinWorks:
Refer to #688


aismann:
We should generate a list of rename function, classes, etc
<old>  <new>
...

DelinWorks:
Good catch! I'll add it in this PR along with `Director` functions that were changed.

DelinWorks:
Okay @halx99 I'm done with my commits.

halx99 (approved):

aismann:
Maybe a list of new functions, classes, etc is also helpfull?
and add it to a doc like 'NEW_API.md'
Can be  a seperate build of course

DelinWorks:
> Maybe a list of new functions, classes, etc is also helpfull? and add it to a doc like 'NEW_API.md' Can be a seperate build of course

that's actually handled in `CHANGELOG.md` in the root directory of the repo

aismann:
> > Maybe a list of new functions, classes, etc is also helpfull? and add it to a doc like 'NEW_API.md' Can be a seperate build of course
> 
> that's actually handled in `CHANGELOG` file in the root directory of the repo

The new: 'enum AnchorPreset' will be found there?

halx99:
Merge first

PR #708, opened 2022-07-05
Update SimpleSnakeScene.cpp

aismann:
fix:   Director::getInstance()->setStatsAnchor(AnchorPreset::...);


halx99:
> BTW: I will maintenance the "game" on my own side now. Its easier to handle new stuff and pull it on githup. So no longer stress with SimpleSnake here on adxe repository. ;) Everybody can use this example here to start with adxe of course.

Yes, you can add a link refer to your own repo

Issue #706, opened 2022-07-05
Concerns regarding RngSeed implementation


rh101:
At the moment the implementation uses calculations like this:

    int32_t range(int32_t min, int32_t max)
    {
        return floor(min + static_cast<float>(rand()) / (static_cast<float>(RNG_RAND_MAX / (max - min))));
    }

No focus on this part of the code:
`static_cast<float>(RNG_RAND_MAX / (max - min))`

Which is effectively `RNG_RAND_MAX / (max - min)`, an integer division, before it is typecast to a float.  

If you pass a value of min = 10, and max = 1000, you will get this:

RNG_RAND_MAX / (max - min) = 4294967295 / (1000 - 10) = **4338350**

Converting the result of that to a float is not required here, since the numerator has already been typecast to a float, the `static_cast<float>(rand())`.

That brings us to the next concern.  Should it actually have been this instead:

` static_cast<float>(RNG_RAND_MAX) / (max - min)`

Where it would result in a floating point division, and the end result for min = 10 and max = 1000 would be this:

static_cast<float>(RNG_RAND_MAX) / (max - min) = 4294967295.0F / (1000 - 10) =  **4338351**

What should be the correct implementation, since each ends up with a different result?  As it is, the code doesn't make sense, since the current implementation is doing an integer division, which is an integer result, and typecasting it to a float makes no difference whatsoever.

rh101:
Also, the RngSeed.hpp is missing include statements, namely:

```
#include <cmath>
#include <limits.h>
```

It's compiling correctly as it is because of the order of compilation, but it should explicitly list the includes that it requires.

DelinWorks:
I've been thinking about it too, it needs a major overhaul including making `max` arguments exclusive NOT inclusive.

rh101:
If you're going to create a PR, please also typecast the result on the return statement to stop all the warnings coming up during compilation. For example:

    int32_t range(int32_t min, int32_t max)
    {
        return floor(min + static_cast<float>(rand()) / (static_cast<float>(RNG_RAND_MAX / (max - min))));
    }

Needs a cast added like this:

    int32_t range(int32_t min, int32_t max)
    {
        return static_cast<int32_t>(floor(min + static_cast<float>(rand()) / (static_cast<float>(RNG_RAND_MAX / (max - min)))));
    }

Otherwise, the compiler throws out a lot of warnings about the conversion.

halx99:
Does the c++11 random library enough?

[...]

PR #710, opened 2022-07-05
[IMPROVMENT] Rename `RngSeed` to `FastRNG`.

DelinWorks:
Refer to #706

### DIFF
--- a/RENAMED_TYPES.md
+++ b/RENAMED_TYPES.md
@@ -1,8 +1,10 @@
 adxe-1.0 ??
 
-### Classes & Types
-* ```Sprite3D``` => ```MeshRenderer```
+### Classes & Structs
+* ```RngSeed``` => ```FastRNG``` UNDER ```math/FastRNG.h```
+* ```Sprite3D``` => ```MeshRenderer``` UNDER ```3d/CCMeshRenderer.h```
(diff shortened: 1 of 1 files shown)


rh101:
@DelinWorks Avoid committing unrelated changes at the same time, and the same goes for the PR.  In this PR there are changes to the `RENAMED_TYPES.md`, which has nothing to do with the random number generator.

There are many reasons for why this is important, so for future commits and PRs, please do try to avoid this.

DelinWorks:
> @DelinWorks Avoid committing unrelated changes at the same time, and the same goes for the PR. In this PR there are changes to the `RENAMED_TYPES.md`, which has nothing to do with the random number generator.
> 
> There are many reasons for why this is important, so for future commits and PRs, please do try to avoid this.

It is related I don't know what you mean

rh101:
> It is related I don't know what you mean

The changes made to the text in the RENAMED_TYPES.md file have nothing to do with the FastRNG changes.  

You can see two commits here:

image

Both of them, I assume, went into the same branch, which is actually what caused the issue.

For each specific issue, you should create a separate branch, commit to it, and then create the PR from that. In this case, they should have been to two different branches.  Once the PR is merged (or no longer required for that specific issue), then the branch can be deleted.

DelinWorks:
Fair enough!

PR #730, opened 2022-07-07
Fix engine setup files.

DelinWorks:


halx99:
The template cmake needs modify at the same time:
https://github.com/axis-project/axis/blob/dev/templates/cpp-template-default/CMakeLists.txt#L43

halx99:
> The template cmake needs modify at the same time: https://github.com/axis-project/axis/blob/dev/templates/cpp-template-default/CMakeLists.txt#L43
> 
> and lua template cmake script also needs sync

No matter, I will do the cmake scripts replacement

DelinWorks:
I hate github desktop with passion

DelinWorks:
> > The template cmake needs modify at the same time: https://github.com/axis-project/axis/blob/dev/templates/cpp-template-default/CMakeLists.txt#L43
> > and lua template cmake script also needs sync
> 
> No matter, I will do the cmake scripts replacement

Don't stress it @halx99 I'll do it

halx99:
> > > The template cmake needs modify at the same time: https://github.com/axis-project/axis/blob/dev/templates/cpp-template-default/CMakeLists.txt#L43
> > > and lua template cmake script also needs sync
> > 
> > 
> > No matter, I will do the cmake scripts replacement
> 
> Don't stress it @halx99 I'll do it

ok.

CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 # Copyright (c) 2015-2017 Chukong Technologies Inc.
 # Copyright (c) 2021-2022 Bytedance Inc.
 #
-# https://adxeproject.github.io/
+# https://axisproject.github.io/
```
halx99: should be: axis-project.githib.io

DelinWorks: I've changed it before I don't know why it still says that:

DelinWorks:
the project creation process now works after these commits.

halx99:
> the project creation process now works after these commits. Also, cmake works just fine! I just tested it 👍

But the ci failed, some base cmake files in `axis/cmake/Modules` not modified

DelinWorks:
> > the project creation process now works after these commits. Also, cmake works just fine! I just tested it 👍
> 
> But the ci failed, some base cmake files in `axis/cmake/Modules` not modified

yeah there are many files that have to be changed and there are a lot of them, could you merge this PR just for a checkpoint or do i have to creat another branch?

halx99:
I can help modify them after lunch

halx99:
Now, I'm working on your branch

halx99:
Next step, could simply replace all `https://adxeproject.github.io` to `https://axis-project.github.io` in all sources

halx99:
If the latest ci check succed, I will merge this PR as a checkpoint

PR #731, opened 2022-07-07
Update setup.py

aismann:
rename ADXE* to AXIS*


aismann:
> @aismann you do same job as @DelinWorks #730, right?

I belive.  ;) Iwill delete it

PR #734, opened 2022-07-08
IndexArray to stl

solan-solan:
Added operator[] for IndexArray to align it to stl container to protect existing code which could use IndexArray with the corresponding api;

Added uint16_t as default template parameter for IndexArray::push_back(...)

### DIFF
--- a/core/3d/CCBundle3DData.h
+++ b/core/3d/CCBundle3DData.h
@@ -103,7 +103,7 @@ public:
     }
 
     /** Pushes back a value. */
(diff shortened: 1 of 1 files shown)


solan-solan:
Yes, thats it. 
It works for uint32_t. Looks not very convinient, but good for uint16_t

One of solution could be to implement operator[] like:
image

Then usage will be the following
image

halx99:
> Yes, thats it. It works for uint32_t. Looks not very convinient, but good for uint16_t
> 
> One of solution could be to implement operator[] like: image
> 
> Then usage will be the following image

looks good

Issue #724, opened 2022-07-07
Migrate engine name, huge changes


halx99:
## source code
- [x] New namespace full name `axmol`, shorten name `ax` @halx99
   - compatible: alias ax to cocos2d: namespace cocos2d = axys; works on old projects who use cocos2dx
- [x] macros: `CC_` to `AX_`
- [ ] Remove source code file name prefix `CC`, such as `CCFileUtils` to `FileUtils`
- [x] Change all `https://adxeproject.github.io` to `https://bytedance.com`
- [x] Migrate extension `axlua` 
- [x] Migrate lua script const prefix `CC_`

## project tools
- [x] cmdline tool: rename from `adxe` to `axmol`
- [x] cmake script: everything `cocos2d` to `axmol`, `CC_` to `AX_`

## docs

- [x] readme files
- [x] docs
- [ ]  logo

aismann:
It is the right way. Great!

aismann:
> * [ ]  Remove source code file name prefix `CC`, such as `CCFileUtils` to `FileUtils`

Maybe prefix AC is better?  uch as `CCFileUtils` to `AXFileUtils`

DelinWorks:
> * [ ]  Remove source code file name prefix `CC`, such as `CCFileUtils` to `FileUtils`

You can add the `breaks compatibility` label so that people know why their project won't compile.

halx99:
I will merge all before any migrate commit to branch `release` and change github default branch to it.

Certainly, any pull request sitll and always been want merge to branch `dev`

DelinWorks:
Should we cleanse every source file that contains `cocos2d` with the word `axis`?

aismann:
> Should we replace every source file that contains `cocos2d` with `axis`?

I think ax:: is the namespace?

DelinWorks:
No I mean like there are comments that contain the word `cocos2d` should we replace them?
The namespaces can be ax:: though

halx99:
> Should we replace every source file that contains `cocos2d` with `axis`?

Yes

aismann:
Maybe a new PR?

halx99:
> Maybe a new PR?

Yes, needs seperate PR

DelinWorks:
@halx99 should this be replaced with `ax` or `axis`?

![image](https://user-images.githubusercontent.com/45469625/178004602-356ea2ff-d482-44dc-8d18-eb7cee8c7711.png)

halx99:
[...]

PR #743, opened 2022-07-11
Migrate engine namespace to `axis`(abbv: `ax`)

halx99:
Refer to #724


DelinWorks:
for the sake of backwards compatibility you need to make an abbreviation for `USING_NS_CC`

Issue #759, opened 2022-07-20
Replace for(auto <var> or   for(auto& <var>  to   for(auto&& <var> …


aismann:
Working with reference and not with copies
=> Performance improvement

PR is ready and tested.

PR #760, opened 2022-07-20
replace for(auto <var> to for(auto& <var>

aismann:
working with reference and not with copies
=> Performance issue

solve feature: #759

### DIFF
--- a/core/2d/CCActionInterval.cpp
+++ b/core/2d/CCActionInterval.cpp
@@ -2495,7 +2495,7 @@ bool Animate::initWithAnimation(Animation* animation)
 
         auto& frames = animation->getFrames();
 
-        for (auto& frame : frames)
+        for (auto&& frame : frames)
         {
             float value = (accumUnitsOfTime * newUnitOfTimeValue) / singleDuration;
             accumUnitsOfTime += frame->getDelayUnits();
--- a/core/2d/CCAnimation.cpp
+++ b/core/2d/CCAnimation.cpp
@@ -121,7 +121,7 @@ bool Animation::initWithSpriteFrames(const Vector<SpriteFrame*>& frames,
     _delayPerUnit = delay;
     _loops        = loops;
 
-    for (auto& spriteFrame : frames)
+    for (auto&& spriteFrame : frames)
     {
         auto animFrame = AnimationFrame::create(spriteFrame, 1, ValueMap());
         _frames.pushBack(animFrame);
@@ -140,7 +140,7 @@ bool Animation::initWithAnimationFrames(const Vector<AnimationFrame*>& arrayOfAn
 
     setFrames(arrayOfAnimationFrames);
 
-    for (auto& animFrame : _frames)
+    for (auto&& animFrame : _frames)
     {
         _totalDelayUnits += animFrame->getDelayUnits();
     }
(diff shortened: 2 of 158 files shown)


DelinWorks:
Some objects cannot be directly referenced. they need to be copied instead, some iterators don't support referencing, and I think that's because of reference counting and that some objects might brake or leak if they're not copied.

halx99:
I think the raw iterator don't needs change to `auto&`, just c++11 for-loop needs for non-pointer element like:
```cpp
for (auto& item : items)
```

aismann:
> I think the raw iterator don't needs change to `auto&`, just c++11 for-loop needs for non-pointer element like:
> 
> ```c++
> for (auto& item : items)
>     item.callSomeMethod();
> ```

Correct! This one & was to much. Sorry. (removed it)

aismann:
> I think the raw iterator don't needs change to `auto&`, just c++11 for-loop needs for non-pointer element like:
> 
> ```c++
> for (auto& item : items)
>     item.callSomeMethod();
> ```

@halx99 
BTW: also lyasio.cpp has a:                   for (auto item : channel_eps) +> can be changed to:  for (auto& item : channel_eps)

At the moment a don't changing anything on thirtparty folders

halx99:
The `channel_ops` element is store as `pointer`, so don't need change

aismann:
> The `channel_ops` element is store as `pointer`, so don't need change

I know. for` (auto item : channel_eps)` is no different to `for (auto& item : channel_eps)`
BUT: as an developer reading the sources its easier for understanding (its a ref not a copy).

halx99:
Using c++11 `Universal Reference` maybe better like:
```cpp

std::vector<bool> items1 = {false};
for(auto&& item : items1) 
    item = true;

const std::vector<bool> items2 = {true};
for (auto && item : items2)
    printf("item=%d\n", (int)item);

```

aismann:
> Using c++11 `Universal Reference` maybe better like:
> 
> ```c++
> std::vector<bool> items1 = {false};
> for(auto&& item : items1) 
>     item = true;
> 
> const std::vector<bool> items2 = {true};
> for (auto&& item : items2)
>     printf("item=%d\n", (int)item);
> ```
Should I change ALL `for (auto& <var1>: <var2>)` to `for (auto&& <var1>: <var2>)`

halx99:
> > Using c++11 `Universal Reference` maybe better like:
> > ```c++
> > std::vector<bool> items1 = {false};
> > for(auto&& item : items1) 
> >     item = true;
> > 
> > const std::vector<bool> items2 = {true};
> > for (auto&& item : items2)
> >     printf("item=%d\n", (int)item);
> > ```
> 
> Should I change ALL `for (auto& <var1>: <var2>)` to `for (auto&& <var1>: <var2>)` ?

Yes, you can, the `Universal References` can handle both const or non-const containers.

aismann:
Also for the 'extension' folder?

The ThirtParty folder will be untouched

halx99:
> Also for the 'extension' folder?
> 
> The ThirtParty folder will be untouched

Yes, but should ignore `spine`, `fairygui`, `Live2D`

aismann:
this extensions be untouched  (spine, Live2D has no such C++11 for loop):
image

halx99 (approved):

PR #763, opened 2022-07-21
Support return type'ResourceData'

tkzcfc:

### DIFF
--- a/extensions/scripting/lua-bindings/auto/lua_axis_ui_auto.cpp
+++ b/extensions/scripting/lua-bindings/auto/lua_axis_ui_auto.cpp
@@ -4650,21 +4650,7 @@ int lua_axis_ui_Layout_getRenderFile(lua_State* tolua_S)
             return 0;
         }
         axis::ResourceData ret = cobj->getRenderFile();
-
-        lua_newtable(tolua_S);
-
-        lua_pushstring(tolua_S, "type");
-        lua_pushinteger(tolua_S, ret.type);
-        lua_rawset(tolua_S, -3);
-
-        lua_pushstring(tolua_S, "file");
-        lua_pushstring(tolua_S, ret.file.c_str());
-        lua_rawset(tolua_S, -3);
-
-        lua_pushstring(tolua_S, "plist");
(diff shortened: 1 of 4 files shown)


rh101:
Doesn't everything in the "auto" folder get overwritten with the automatic binding creation script?  If that is the case, then this PR is not valid, and any entries need to be done somewhere else, perhaps in the "manual" folder.

aismann:
BTW: /extensions/scripting/lua-bindings/auto should be disabled GitHub pull process. I have always uncheck it 😔

rh101:
> BTW: /extensions/scripting/lua-bindings/auto should be disabled on GitHub pull process. I have always uncheck it 😔

Is there any specific reason you prefer to not have them in the repository?

It would mean users would need to generate the bindings themselves, and that's just an extra setup step that may cause problems, especially for newer users of the engine.

aismann:
I mean the automatic generated stuff.

rh101:
> I mean the automatic generated stuff, the automatic binding stuff

I understand, but I mean what is the reason not wanting that generated files to be in the repository?  The reason I ask is that the only time the auto-generated files are affected is if the engine code changes, and that is only relevant for engine developers, and not the majority of users of the engine.  Forcing typical users of the engine to generate the bindings seems like an unnecessary step.

DelinWorks:
@rh101 I think what he meant is any pull requests that tries to modify lua-auto files should be denied, not actually removing the auto generated lua files from the repo

rh101:
> PR to "automatic created stuff (executables, apk's, etc) should be permit.
> So `PRs like this` is not possible.

Sorry, I completely misunderstood what you meant in your initial post, but now I get what you mean, and I agree with you.

@DelinWorks I think I need to avoid posting so late at night 🤦‍♂️

DelinWorks:
> > PR to "automatic created stuff (executables, apk's, etc) should be permit.
> > So `PRs like this` is not possible.
> 
> Sorry, I completely misunderstood what you meant in your initial post, but now I get what you mean, and I agree with you.
> 
> @DelinWorks I think I need to avoid posting so late at night 🤦‍♂️

Can you even do that in github?

PR #769, opened 2022-08-02
Crashfix in UserDefault

askamn:
Missed an edge case when resizing the memorymapped file in UserDefault that could cause a crash. We never really checked if the new size is enough! 

Not sure if we need an upper-limit here on the size.  @halx99 what do you think?


halx99:
Needs prvoide a API: UserDefault::setMaxMappingSize, and by default, we can set it to 4MB

halx99:
> Needs prvoide a API: UserDefault::setMaxMappingSize, and by default, we can set it to 4MB

Need a separate PR

askamn:
I will submit one soon.

PR #778, opened 2022-08-04
[PROPOSAL] Add wireframe rendering and function classification improvements.

DelinWorks:
## Describe your changes
Added a wireframe rendering mode for mesh renderers which would ease renderer debugging. 
`meshRenderer->setWireframe(true);` example:

https://user-images.githubusercontent.com/45469625/182822530-56567e37-e523-4f07-b794-7830f1ee5401.mp4

Exposed the mesh transparency functions and moved it to the material class.

Moved functions that should've been contained within the material class and NOT the mesh class, moved functions are `setTransparent` and `setForce2DQueue`.

They can now be accessed from the material class directly like:
```
material->setTransparent(true);
material->setForce2DQueue(true);
```
and if this material gets applied to multiple meshes then those functions will affect the meshes with this attached material just like any other game engine.


DelinWorks:
@halx99 ready! you can merge now.

Issue #788, opened 2022-08-09
Wrong libjpeg version for Mac included


pietpukkel:
The now included precompiled binary for Mac of libjpeg is version 90, whereas the headerfiles are for version 80 (leading to an error when performing the Texture2D test in cpp_test). I've recompiled with version 80 and will send a pull request.

aismann:
@halx99 
Can it closed?

halx99:
Refer to https://github.com/axys1/buildware/pull/6

pietpukkel:
Can be closed but please remove the Mac libraries I asked to be pulled earlier and replace them with ones build  from buildware after I've included glsl-optimizer too.

halx99:
> Can be closed but please remove the Mac libraries I asked to be pulled earlier and replace them with ones build from buildware after I've included glsl-optimizer too.

Sure

PR #773, opened 2022-08-02
Support for Apple Silicon / M1

pietpukkel:
THIS PROJECT IS IN DEVELOPMENT MODE. Any pull requests should merge to branch `dev`, otherwise will be rejected immediately.

## Describe your changes
Added universal libraries where needed and changed PNG CMake file to include ARM files when generating Mac XCode project.  After creating the Mac xcode project on an M1 Mac the tests build and run out of the box.

## Issue ticket number and link
#770 , https://github.com/axis-project/axis/issues/770

## Checklist before requesting a review
- [x] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests.

PR #789, opened 2022-08-09
Fixed libjpeg and libz for Mac

pietpukkel:
Fixed libjpeg and libz for Mac (x86_64 and arm64)

## Describe your changes
Recompiled libz and libjpeg to be the same version for Mac as for the other platforms, fixing my earlier error.

#788 
https://github.com/axys1/axys/issues/788

- [x ] I have performed a self-review of my code


rh101:
@pietpukkel Is there a link to the source code used to build these libraries, and also, how did you build them? This needs to be detailed step by step, in order to reproduce the exact libraries you generated, in case someone else needs to do it in the future.

@halx99 It's a little concerning that users can just add their own pre-built libraries to this project, especially since there is no way to validate them, because there is a possibility of the introduction of security issues (both intentional and unintentional).  The source code should be linked along with the instructions on how to build them for all platforms they are being used on.

halx99:
All prebuilt libraries should build from https://github.com/axys1/buildware

halx99:
We use libjpeg-turbo which only provide v6,v7,v8 ABI, and currently it auto build from github actions.

Please refer to https://github.com/libjpeg-turbo/libjpeg-turbo/tree/2.1.3

halx99:
@pietpukkel Do you have any idea to make axys/buildware support Apple M1?

pietpukkel:
> 

Let me kick myself in the head. I have been looking everywhere for the sources you used to build these 4 libraries. I compiled them from their respective GitHub repos after figuring out which version was included.  I or course completely agree with @rh101. Let me take a look at axys1/buildware and see if i can add the Arm64 for Mac to it. This also is an opportunity to add the missing x86_64 for Android to it (i am just a self taught hobbyist so be gentle with me ;-)

pietpukkel:
> @pietpukkel Do you have any idea to make axys1/buildware support Apple M1?

Well it appears buildware already is capable of building arm64 binaries for Mac. When I generate the Mac binaires with the current scripts, on my M1 Ultra Mac arm64 binaries are created without me having to change anything (no x86_64 though, so the scripts still need to be adapted to crosscompile).

pietpukkel:
Oh and when I run the scripts on Mac under Rosetta (the X86_64 emulation layer), they create the X86_64 binaries.

halx99:
For openssl, needs check `BUILD_ARCH` at here: https://github.com/axys1/buildware/blob/main/1k/build1.sh#L63

x64: darwin64-x86_64-cc
arm64: darwin64-arm64-cc

pietpukkel:
> For openssl, needs check `BUILD_ARCH` at here: https://github.com/axys1/buildware/blob/main/1k/build1.sh#L63
> 
> x64: darwin64-x86_64-cc arm64: darwin64-arm64-cc
> 
> refer to: https://github.com/openssl/openssl/blob/master/Configurations/10-main.conf#L1703

Done 👍. This also needed an architecture check as with this solution OpenSSL is basically always crosscompiled (also when native machine is available). Now on to the others :-)

pietpukkel:
I've updated the build scripts. For Mac i've included arm64 and for android I've added x64.  Everthing looks good, I am only struggeling to get libluajit to compile for any platform (compile fails on my mac for ios, android and linux ... make exits with error 127).

I've closed this pull request. Please also undo my earlier pull request (#773).  Once I'm able to build all and have been able to test all libraries I will share the updated build code in my fork so you can build them yourself and add them.

Issue #791, opened 2022-08-10
The Camera view incorrect when designResolutionSize changed


halx99:
- axys version:
- devices test on:
- developing environments
   - NDK version: r19c
   - Xcode version: 12.4
   - VS/msvctools version: Visual Studio 2019(16.11)/14.29.30133
   - cmake version: 
Steps to Reproduce:

1. 
2.

RT, this issue introduced since #653 merged.
@DelinWorks Could you take a look?

Projection is `3D`

DelinWorks:
could you elaborate more on what you mean by `designResolutionSize changed`?

halx99:
Seems there have no chance to update fov when winSize changed: https://github.com/axys1/axys/blob/dev/core/2d/CCCamera.cpp#L457

halx99:
### Before increase winSize.height:

![image](https://user-images.githubusercontent.com/6977319/183901409-a3af6049-8e33-4eac-aef9-f73971eb516f.png)


### After increase winSize.height

![image](https://user-images.githubusercontent.com/6977319/183901499-1d4c255f-4ee5-4da5-b13b-9017ac7f0001.png)

halx99:
> 

The reproduce example code could be:

1. set init frame and design size
```cpp
glview->setFrameSize(1280, 720);
glview->setDesignResolutionSize(1280, 720, ResolutionPolicy::NO_BORDER);
```

3. run a scene with some nodes, i.e. label or sprite
4. increase design resolution and frame size at next frame
```cpp
glview->setFrameSize(1920, 1080);
glview->setDesignResolutionSize(1920, 1080, ResolutionPolicy::NO_BORDER);
```

DelinWorks:
Yeah I was able to replicate that bug, I'll fix it no worries

PR #792, opened 2022-08-10
[BUG] Fix camera view calculation and tidy up system.

DelinWorks:
Refer to #791
The camera features that I added were bandaids over bandaids so I decided to rewrite it and it's significantly better now!

### DIFF
--- a/core/2d/CCCamera.cpp
+++ b/core/2d/CCCamera.cpp
@@ -42,7 +42,7 @@ Viewport Camera::_defaultViewport;
 Camera* Camera::create()
 {
     Camera* camera = new Camera();
-    camera->initCamera();
+    camera->init();
     camera->autorelease();
     camera->setDepth(0);
 
@@ -52,9 +52,6 @@ Camera* Camera::create()
 Camera* Camera::createPerspective(float fieldOfView, float aspectRatio, float nearPlane, float farPlane)
 {
     auto ret = new Camera();
-    ret->_fieldOfView = fieldOfView;
-    ret->_nearPlane   = nearPlane;
-    ret->_farPlane    = farPlane;
     ret->initPerspective(fieldOfView, aspectRatio, nearPlane, farPlane);
     ret->autorelease();
     return ret;
@@ -63,10 +60,6 @@ Camera* Camera::createPerspective(float fieldOfView, float aspectRatio, float ne
 Camera* Camera::createOrthographic(float zoomX, float zoomY, float nearPlane, float farPlane)
 {
     auto ret = new Camera();
-    ret->_zoom[0]   = zoomX;
-    ret->_zoom[1]   = zoomY;
-    ret->_nearPlane = nearPlane;
-    ret->_farPlane  = farPlane;
     ret->initOrthographic(zoomX, zoomY, nearPlane, farPlane);
     ret->autorelease();
     return ret;
@@ -191,7 +184,7 @@ void Camera::setAdditionalProjection(const Mat4& mat)
     getViewProjectionMatrix();
 }
 
-void Camera::initCamera()
+bool Camera::init()
 {
     auto& size = _director->getWinSize();
     switch (_director->getProjection())
@@ -223,6 +216,8 @@ void Camera::initCamera()
     }
 
     updateTransform();
+
+    return true;
 }
 
 void Camera::updateTransform()
--- a/core/2d/CCCamera.cpp
+++ b/core/2d/CCCamera.cpp
@@ -42,37 +42,7 @@ Viewport Camera::_defaultViewport;
 Camera* Camera::create()
 {
     Camera* camera = new Camera();
-
-    auto& size = Director::getInstance()->getWinSize();
-    switch (Director::getInstance()->getProjection())
-    {
-        case Director::Projection::_2D:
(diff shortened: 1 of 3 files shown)


core/2d/CCCamera.cpp
```diff
@@ -42,7 +42,37 @@ Viewport Camera::_defaultViewport;
 Camera* Camera::create()
 {
     Camera* camera = new Camera();
-    camera->initDefault();
+
+    auto& size = Director::getInstance()->getWinSize();
```
halx99, line 46:
```suggestion
    auto& size = _director->getWinSize();
```

core/2d/CCCamera.cpp
```diff
@@ -42,7 +42,37 @@ Viewport Camera::_defaultViewport;
...
+    switch (Director::getInstance()->getProjection())
```
halx99, line 47:
```suggestion
    switch (_director->getProjection())
```

core/2d/CCCamera.cpp
```diff
@@ -195,56 +221,33 @@ void Camera::setAdditionalProjection(const Mat4& mat)
...
-    {
-        applyCustomProperties();
-        return true;
-    }
-    _isCameraInitialized = true;
-
     auto& size = _director->getWinSize();
     // create default camera
-    switch (_projectionType)
+    switch (Director::getInstance()->getProjection())
```
halx99, line 228:
```suggestion
    switch (_director->getProjection())
```

core/2d/CCCamera.cpp
```diff
@@ -405,7 +408,7 @@ void Camera::setZoom(float factor)
 
 void Camera::applyZoom()
 {
-    switch (_projectionType)
+    switch (Director::getInstance()->getProjection())
```
halx99, line 411:
```suggestion
    switch (_director->getProjection())
```

halx99 (changes requested):

core/2d/CCCamera.cpp
DelinWorks: `_director` cannot be used in a static function.

core/2d/CCCamera.cpp
```diff
@@ -42,7 +42,37 @@ Viewport Camera::_defaultViewport;
...
+                up(0.0f, 1.0f, 0.0f);
+            camera->setPosition3D(eye);
+            camera->lookAt(center, up);
+            break;
+        }
+    }
+
+    camera->updateTransform();
     camera->autorelease();
     camera->setDepth(0);
```
halx99, line 77: The code block still can wraped as a member function of Camera.

halx99 (approved):

core/2d/CCCamera.cpp
```diff
@@ -42,7 +42,7 @@ Viewport Camera::_defaultViewport;
 Camera* Camera::create()
 {
     Camera* camera = new Camera();
-    camera->initDefault();
+    camera->initCamera();
```
halx99, line 45: Can be still name `initDedault`?

core/2d/CCCamera.cpp
```diff
@@ -42,7 +42,7 @@ Viewport Camera::_defaultViewport;
...
     camera->autorelease();
     camera->setDepth(0);
```
halx99, line 47: Can be move to `initDefault`?

core/2d/CCCamera.cpp
```diff
@@ -53,11 +53,9 @@ Camera* Camera::createPerspective(float fieldOfView, float aspectRatio, float ne
 {
     auto ret = new Camera();
     ret->_fieldOfView = fieldOfView;
```
halx99, line 55: Does the 3 assign statements can be delete?

core/2d/CCCamera.cpp
```diff
@@ -70,7 +68,6 @@ Camera* Camera::createOrthographic(float zoomX, float zoomY, float nearPlane, fl
     ret->_nearPlane = nearPlane;
     ret->_farPlane  = farPlane;
```
halx99, line 69: Does the 4 assign statements can be delete?

core/2d/CCCamera.cpp
DelinWorks: it's gonna be misleading if `initDefault` is used, previous code bases will think it's an old API function and will lead to bugs.

DelinWorks: Actually, it needs to be `init` only overriding the node's `init` function, that way the code is cleaner

DelinWorks:
The ci error is caused by `lua auto generated bindings` I don't think they need to be fixed because they're auto, that's what I think

halx99:
These changes can't fix #791, I fix with the commit https://github.com/axys1/axys/commit/3cf19787ce1bb4945e42c73e013c735c7a046bd2, and I rollback default camera init from `init` to `initDefault`.

halx99:
> These changes can't fix #791, I fix with the commit [3cf1978](https://github.com/axys1/axys/commit/3cf19787ce1bb4945e42c73e013c735c7a046bd2), and I rollback default camera init from `init` to `initDefault`.

I'm not sure does this will break your features, I think you can check it

PR #798, opened 2022-08-11
[DRAFT] Use fopen for isFileExistsInternal on android

theunwisewolf:
I was just profiling my game and realized that isFileExists was appearing in my profile for some reason. Upon digging further, it seems we are using the stat struct to know about the file's existence on android. I changed it back to `fopen` and the profiling showed a great improvement. 

I tested this on an android 10 device (snapdragon 865+) with the project built in release with profiling on. 
Maybe you can also test on other devices.

`Performance with stat`  
stat

`Performance with fopen`  
fopen

### DIFF
--- a/core/platform/android/CCFileUtils-android.cpp
+++ b/core/platform/android/CCFileUtils-android.cpp
@@ -107,7 +107,6 @@ bool FileUtilsAndroid::init()
 
 bool FileUtilsAndroid::isFileExistInternal(std::string_view strFilePath) const
 {
-
     DECLARE_GUARD;
 
     if (strFilePath.empty())
@@ -118,10 +117,9 @@ bool FileUtilsAndroid::isFileExistInternal(std::string_view strFilePath) const
     bool bFound = false;
 
     // Check whether file exists in apk.
+    const char* s = strFilePath.data();
     if (strFilePath[0] != '/')
     {
-        const char* s = strFilePath.data();
-
         // Found "assets/" at the beginning of the path and we don't want it
         if (strFilePath.find(_defaultResRootPath) == 0)
             s += _defaultResRootPath.length();
@@ -146,9 +144,12 @@ bool FileUtilsAndroid::isFileExistInternal(std::string_view strFilePath) const
     }
     else
     {
-        struct stat st;
-        if (::stat(strFilePath.data(), &st) == 0)
-            bFound = S_ISREG(st.st_mode);
+        FILE *fp = fopen(s, "r");
+        if (fp)
+        {
+            bFound = true;
+            fclose(fp);
+        }
     }
     return bFound;
 }


halx99:
Could you share your performance test code?

halx99:
Since our build system fully support c++17 filesystem, what about use filesystem `is_regular_file`?

askamn:
> Could you share your performance test code?

I tested this on my production game's code, so I can't share it :( 
You can just load run the tests and profile them.

> Since our build system fully support c++17 filesystem, what about use filesystem `is_regular_file`?
`is_regular_file` or `exists` ?
Let me try to profile this.

halx99:
OK, I will test on my device `MI MIX4 (Android 12)`

theunwisewolf:
@halx99 

I wrote a very simple test and slapped it in FileUtilsTests in cpp_tests project and std::filesystem api is the slowest.

These are the results - 
fopen - 6.5ms
struct stat - 6.2ms
std::filesystem::exists - 10ms
std::filesystem::is_regular_file - 9.5ms

`stat` seems to be performing better here as expected. Not really sure why my initial tests showed it to be slow.

```
// TestIsFileExistPerformance

void TestIsFileExistPerformance::onEnter()
{
    FileUtilsDemo::onEnter();
    auto fs = FileUtils::getInstance();

    int subDirectories;
    for (int i = 0; i < MAX_FILES; i++)
    {
        _filePaths[i] = fs->getWritablePath();
        subDirectories = rand() % 5;
        for (int j = 0; j < subDirectories; j++)
            _filePaths[i] += std::to_string(j) + "/";

        if (subDirectories > 0)
            fs->createDirectory(_filePaths[i]);
        _filePaths[i] += "_" + std::to_string(i);
    }

    // The data that we will write to test files
    std::string data(2048, '0');
    for (int i = 0; i < MAX_FILES; i+=2)
    {
        fs->writeStringToFile(data, _filePaths[i]);
    }

    int existsCount = 0;
    int notExists = 0;
    auto start = std::chrono::high_resolution_clock::now();
    for (int i = 0; i < MAX_FILES; ++i)
    {
        if (fs->isFileExist(_filePaths[i]))
        {
            existsCount++;
        }
        else
        {
            notExists++;
        }
    }
    auto end = std::chrono::high_resolution_clock::now();
    auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count();

    AXLOG("[TestIsFileExistPerformance]: Took %d ms", ms);

    AXASSERT(existsCount == MAX_FILES / 2, "existsCount invalid!");
    AXASSERT(notExists == MAX_FILES / 2, "notExists invalid!");
}

void TestIsFileExistPerformance::onExit()
{
    FileUtils* fs = FileUtils::getInstance();
    for (int i = 0; i < MAX_FILES; ++i)
    {
        if (fs->isFileExist(_filePaths[i]))
        {
             fs->removeFile(_filePaths[i]);
        }
    }

    FileUtilsDemo::onExit();
}

std::string TestIsFileExistPerformance::title() const
{
    return "FileUtils: check performance of isFileExistInternal";
}

std::string TestIsFileExistPerformance::subtitle() const
{
    return "";
}
```

```
class TestIsFileExistPerformance : public FileUtilsDemo
{
private:
    static const int MAX_FILES = 2000;
    std::string _filePaths[MAX_FILES];

public:
    CREATE_FUNC(TestIsFileExistPerformance);

    virtual void onEnter() override;
    virtual void onExit() override;
    virtual std::string title() const override;
    virtual std::string subtitle() const override;
};
```

Issue #775, opened 2022-08-03
Localization Support


theunwisewolf:
**Is your feature request related to a problem? Please describe.**
Games sometimes require localization support. Currently, UIText and CCLabel do not have any support for this.

**Describe the solution you'd like**
First, cocos-studio already has a localization manager that is unused (No support in cocos studio to enable localization, although it is part of the flatbuffer schema, see options->isLocalized()). I was thinking that maybe we can edit the reader code to localize strings if some global flag is set. Also, localization should be done in such a way that it also enables to set the font on the UIText, because some languages like arabic might require a different font. Maybe something like this:

```
// TextReader.cpp
ILocalizationManager* lm = LocalizationHelper::getCurrentManager();
if (lm) // or something like if (g_FlagEnableLocalization)
{
    lm->setLocalizedString(label, text);
}
else
{
    label->setString(text);
}

// CustomLocalizationManager.cpp
void CustomLocalizationManager::setLocalizedString(Text* node, std::string_view text)
{
      auto font = this->getFontBasedOnLanguage(_currentLanguage);
      node->setFontName(font);
      node->setString(this->localizeString(text));
}
```

**Describe alternatives you've considered**
Another alternative is to just mark UIText::setString() and UIText::setFontName() as virtual so users can write their custom UIText nodes inherited from UIText and then they can override setString() and setFontName().
The TextReader in cocos studio can already be overriden by registering the type in ObjectFactory before the app loads, and there instead of doing Text::create() we can create our custom TextCustom::create() node.

And, one more approach could be to just add localization support engine wide, so that setString() function localizes the text before it sets it.

Please suggest a good approach so I can work on it and submit a PR!

DelinWorks:
I've elegantly written my own localization string and here is how it goes:

[...]

PR #799, opened 2022-08-11
Add custom localization support.

DelinWorks:
Refer to: #775

### DIFF
--- a/core/2d/CCLabel.cpp
+++ b/core/2d/CCLabel.cpp
@@ -568,7 +568,8 @@ Label::~Label()
     AX_SAFE_RELEASE_NULL(_textSprite);
     AX_SAFE_RELEASE_NULL(_shadowNode);
 
-    _localizer->release();
+    if (_localizer)
+        _localizer->release();
 }
 
 void Label::reset()
@@ -788,8 +789,11 @@ void Label::updateBatchCommand(Label::BatchCommand& batch)
 
 void Label::setLocalizationHandler(LocalizationHandler* handler)
 {
-    _localizer = handler;
-    handler->retain();
+    if (handler)
+    {
+        _localizer = handler;
+        handler->retain();
+    }
 }
 
 void Label::updateUniformLocations()
--- a/core/2d/CCLabel.cpp
+++ b/core/2d/CCLabel.cpp
@@ -789,11 +789,9 @@ void Label::updateBatchCommand(Label::BatchCommand& batch)
 
 void Label::setLocalizationHandler(LocalizationHandler* handler)
 {
+    _localizer = handler;
     if (handler)
-    {
-        _localizer = handler;
         handler->retain();
-    }
 }
 
 void Label::updateUniformLocations()
--- a/core/2d/CCLabel.cpp
+++ b/core/2d/CCLabel.cpp
@@ -567,6 +567,8 @@ Label::~Label()
 
     AX_SAFE_RELEASE_NULL(_textSprite);
     AX_SAFE_RELEASE_NULL(_shadowNode);
+
+    _localizer->release();
 }
 
 void Label::reset()
@@ -784,6 +786,12 @@ void Label::updateBatchCommand(Label::BatchCommand& batch)
     batch.setProgramState(_programState);
 }
 
+void Label::setLocalizationHandler(LocalizationHandler* handler)
+{
+    _localizer = handler;
+    handler->retain();
+}
+
 void Label::updateUniformLocations()
 {
     _mvpMatrixLocation   = _programState->getUniformLocation(backend::Uniform::MVP_MATRIX);
@@ -953,6 +961,9 @@ bool Label::setBMFontFilePath(std::string_view bmfontFilePath, const Vec2& image
 
 void Label::setString(std::string_view text)
 {
+    if (_localizer)
+        text = _localizer->_filter(this, text);
+
     if (text.compare(_utf8Text))
     {
         _utf8Text     = text;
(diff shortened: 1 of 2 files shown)


askamn:
Looking good! Only thing that concerns me, is that now we have to store another 8 bytes per Label for the callback. But then again, if I look at CCLabel it has a thousand other fields lol, like all those bools which could have all been save in a single 64 bit integer (if that was that case, we could have also fit a new bool called _doLocalize there). 

Gosh, writing messages here to & fro is so weird on github. We need a discord server!

rh101:
This implementation is not following conventional methods of how localization is handled.  

Correct me if I'm wrong in my understanding of how this implementation works:

`label->setString("Hello World");`

This goes through some filter to convert it to another language, so you're effectively using that string as the conversion key, which makes no sense, since that string you're passing in should be the actual text to be displayed.  The strings could be of any size, and you definitely don't want to be using that string in whatever the base language is (ie. EN) as the key.

The purpose of the label is to display text, not to modify it in any way, and that string should not be treated as a key to lookup text for other languages.

If that is how this implementation works, then I do not agree with it at all.

The localization implementation should be completely separate from the label, and and it's as simple as this (where `l10n` is the localization method or similar):

`label->setString(l10n('HELLO_WORD"));`

This would also have the effect of allowing you to look up the text in all languages by simply searching for that key.

DelinWorks:
This implementation is correct, and as I stated above ^ it's just a test to show that the filter works, it only gets called if the pointer to that filter handler is not null, also the filter function just gets called if setString is called, that just gives the ability to automatically make the function deal with the passed values instead, and to ease the workflow of setting strings. and that's why I think this implementation is correct.

rh101:
> There are instances where projects have a lot of lines of code where `setString` is used.. 

The frequency of `setString` should not be the deciding factor on how to implement this functionality, and in this instance, all setString code would need to be updated regardless in order to add the filter.   This implementation is effectively changing the meaning of `setString`.  

> You don't want to go to every line of every label and set it to `setStringLocalized` inorder to modify it

You would definitely would want to go through every call to `setString` to add localization support, because then you're making your intention clear in the source code.

Even if a method like `setLocalizedString` is implemented, it would need to be different to `setString`, for instance:

`setLocalizedString(std::string key)` which would make its intention clear, that you're passing a key, not the string to be displayed, and then that setLocalizedString would do the conversion, and then call `setString` with the localized text.

Now, while something like `setLocalizedString` would be helpful, it's still polluting the label class with code that doesn't actually have anything to do with the label.  This is a text conversion issue, not a label issue.

DelinWorks:
how about adding a comment to `setString` that it has 2 abilities? besides that function is useful as it's just a filterer and that's it, it's like you're overriding `setString`

rh101:
> how about adding a comment to `setString` that it has 2 abilities? besides, that function is useful as it's just a filterer and that's it, it's like you're overriding `setString`

I do really understand where you're coming from regarding this issue, in that you want to make it easier on developers, but take a step back for a second and think about this.   

In this case, this implementation violates the single responsibility principle, because you're trying to have `setString` do two different and unrelated things.  It's intention is just to set the text that is passed in, not to modify it.  Adding a comment won't help at all.

Current code in the engine may already have similar issues with methods or classes doing more than they should be; if anything, we should be cleaning up the code to reduce the problem, not add to it.

DelinWorks:
The label pointer seems nice to have...

DelinWorks:
> Current code in the engine may already have similar issues with methods or classes doing more than they should be; if anything, we should be cleaning up the code to reduce the problem, not add to it.

Well if someone is fairly dumb to not realize that the function does two things, then I don't think they should be using a C++ game engine and instead switch to godot gdscript (I have nothing against godot nor gdscript), also we don't want classes to be fat, meaning to have mutiple things that can be packed to a single function. just like `DoThatA`, `DoThatB`, DoThatC` that doesn't feel right. also other game engines like godot stuff multiple functions into one A LOT, so yeah that's what I think of it

rh101:
> Well if someone is fairly dumb to not realize that the function does two things, then I don't think they should be using a C++ game engine and instead switch to godot gdscript

Please avoid making remarks like that, because they never help in any way.

> also we don't want classes to be fat, meaning to have mutiple things that can be packed to a single function. just like `DoThatA`, `DoThatB`, `DoThatC` that doesn't feel elegant, when they're all can be `DoThat` and that's it

I'm definitely not suggesting a class be "fat", completely the opposite in fact.  A class has a specific purpose, and in this suggested implementation, you're quite literally adding something to the class that has nothing to do with it, in essence, making  it "fat".  To be more specific, the label class has one purpose, which is to display text, not to modify it.

I know we think or have feelings about how things are, but that doesn't necessarily mean that our point of view is correct.  Again, I get what you're trying to achieve, but this isn't the way.

DelinWorks:
So what do you think the solution is? make it a singleton and leaving the label class alone? because it's important that having it in cocostudio isn't right

rh101:
> So what do you think the solution is? make it a singleton and leave the label class alone? because it's important that having it in cocostudio isn't right

Cocos Studio is a relic that is still in the code base for, I assume, backwards compatibility.  It's one of the things that eventually need to be addressed, but that doesn't mean it needs to be now.  

Just because Cocos Studio has some code related to localization doesn't mean that code needs to be used, or moved.  It can be left as is, and a new localization support be added elsewhere.



> damn we really need a discord server for this

Discord is the worst place to deal with this.  Having it here means there is a history of the discussion and decisions made.  Discord would be great for asking questions for help with the engine or similar things, but not for stuff like this.

DelinWorks:
> Discord is the worst place to deal with this. Having it here means there is a history of the discussion and decisions made. Discord would be great for asking questions for help with the engine or similar things, but not for stuff like this.

Discord does store discussions you now, and is way better than github from my opinion 

I'll close this PR for now

PR #801, opened 2022-08-11
RichElementImage - Add support for scaling

theunwisewolf:

## Describe your changes
Added support for "scaleX", "scaleY" and "scale" tags for rich text  tag.
Added support for width/height in percentage. E.g. ``
Also fixed a bug with the url attribute. Even if you didn't put a url in your img, it will still call openUrl() for some reason.

- [x] If it is a core feature, I have added thorough tests.

### DIFF
--- a/core/ui/UIRichText.cpp
+++ b/core/ui/UIRichText.cpp
@@ -471,13 +471,29 @@ MyXMLVisitor::MyXMLVisitor(RichText* richText) : _fontElements(20), _richText(ri
         it = tagAttrValueMap.find("height");
         if (it != tagAttrValueMap.end())
         {
-            height = it->second.asInt();
+            auto str = it->second.asStringRef();
+            if (str[str.length() - 1] == '%')
+            {
+                scaleY = std::atoi(str.data()) / 100.f;
+            }
+            else
+            {
+                height = it->second.asInt();
+            }
         }
 
         it = tagAttrValueMap.find("width");
         if (it != tagAttrValueMap.end())
         {
-            width = it->second.asInt();
+            auto str = it->second.asStringRef();
+            if (str[str.length() - 1] == '%')
+            {
+                scaleX = std::atoi(str.data()) / 100.f;
+            }
+            else
+            {
+                width = it->second.asInt();
+            }
         }
 
         it = tagAttrValueMap.find("scaleX");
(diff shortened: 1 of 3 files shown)


halx99:
What about implement html standard `width="150%"

theunwisewolf:
Check @halx99

core/ui/UIRichText.cpp
```diff
@@ -471,13 +471,29 @@ MyXMLVisitor::MyXMLVisitor(RichText* richText) : _fontElements(20), _richText(ri
         it = tagAttrValueMap.find("height");
         if (it != tagAttrValueMap.end())
         {
-            height = it->second.asInt();
+            auto str = it->second.asStringRef();
+            if (str[str.length() - 1] == '%')
```
rh101, line 475: It's best to check if the string is empty before attempting to do a `[length - 1]`, because this will definitely cause a crash if the string is like this: `height=""`. Same goes for the `width`.

You could also `if (str.back() == '%')`, but you would still need to check if the string is empty beforehand.

rh101: Then again, there doesn't seem to be much (if any) validation code in the RichText implementation for all the other fields either, so perhaps the assumption is that they will always be valid.  This could lead to problems if the text is added dynamically (such as based on user-input), but if the text is hard-coded, then any issues will most likely show up in development.

halx99: All find and try get value could use: `axys::optValue(valueMap, "key"sv).asXXX` instead.

theunwisewolf: @rh101, Yeah we can add empty check but as you said, there is no validation anyways so I didn't think it was needed. I will add it.

@halx99 optValue is extremely slow. You should profile it. I have already removed it from my fork as it was slowing down plist loading for particle effects as well as spriteframes. I suspect it has something to do with the custom string_hash & equal_to functions you have written for string_view for use with robin_map (to support both string & string_view interchangeably, right?), although I was not able to figure out the exact cause.

halx99: Check does the latest commit solve the performance issue:
https://github.com/axys1/axys/commit/d8c9a8826ac8744bd718fc1170144c562921d9bf

theunwisewolf: Cool. I will profile and get back to you.

aismann: > Cool. I will profile and get back to you.

Maybe the Optick profiler #845 is helping you?

aismann: > Check does the latest commit solve the performance issue: [d8c9a88](https://github.com/axmol1/axys/commit/d8c9a8826ac8744bd718fc1170144c562921d9bf)

@halx99 
Is this the solution?
So merge can be done or what is missing?

Issue #819, opened 2022-08-27
Replace clipper1 vs Clipper2 


aismann:
I have replaced Clipper1 vs Clipper2:
Also adapt AutoPolygon class to Clipper2

All tests have the same results but mostly faster ;) , that's the goal!

![image](https://user-images.githubusercontent.com/8652787/187030030-316286dc-b17e-4c00-8d45-e88d830e8ea2.png)

aismann:
can be closed

PR #820, opened 2022-08-27
Replace clipper1 vs clipper2 completely

aismann:
## Describe your changes
Replace clipper1 vs clipper2 completely
Adapt AutoPolygon class

#819 

- [ x] I have performed a self-review of my code
- [ x] If it is a core feature, I have added thorough tests (same as with clipper1).


aismann:
I have to check the lua part also (didnt it)

PR #835, opened 2022-08-31
Update CCQuadCommand.cpp (cocos2dx #20744)

aismann:
see also https://github.com/cocos2d/cocos2d-x/pull/20744/files


halx99:
It's not a correct fix, will cause crash. just run cpp-tests

aismann:
> It's not a correct fix, will cause crash. just run cpp-tests

You are right. closed it. wrong turn.

Issue #866, opened 2022-09-27
Component::onEnter() not called if component added after parent node is running


rh101:
- axys version: All, with latest checked being https://github.com/axys1/axys/tree/f8c3d95176422960a9aa210bd61425432851c948
- devices test on: N/A
- developing environments
   - NDK version: r19c
   - Xcode version: 12.4
   - VS/msvctools version: Visual Studio 2019(16.11)/14.29.30133
   - cmake version: 

This issue exists in Cocos2d-x as well.

Steps to Reproduce:

1. Add a Component to a node at any point after `Node::onEnter()` is run (so outside of `Node::init()` etc), such as in `Node::update()`.
2. `Component::onEnter()` is never called.

I can't seem to find any documentation related to this, so I'm not sure if this is a bug or intended behaviour, but it's strange that it would be implemented this way.

The `Node::addComponent()` is implemented like this:
```
bool Node::addComponent(Component* component)
{
    // lazy alloc
    if (!_componentContainer)
        _componentContainer = new ComponentContainer(this);

    // should enable schedule update, then all components can receive this call back
    scheduleUpdate();

    return _componentContainer->add(component);
}
```

The only place `Component::onEnter()` is called is here, where the component must already exist in the list prior to `Node::onEnter()` being called:
```
void Node::onEnter()
{
...
    if (_componentContainer && !_componentContainer->isEmpty())
    {
        _componentContainer->onEnter();
    }

    _isTransitionFinished = false;

    for (const auto& child : _children)
        child->onEnter();

    this->resume();

    _running = true;
}
```

This is completely different to when using `Node::addChild()`, as the `addChild()` method will actually call `Node::onEnter()` if the parent node is already running (it checks the `_running` flag):

```
void Node::addChild(Node* child, int localZOrder, int tag)
{
...
    addChildHelper(child, localZOrder, tag, "", true);
}

void Node::addChildHelper(Node* child, int localZOrder, int tag, std::string_view name, bool setTag)
{
...
    if (_running)
    {
        child->onEnter();
...
    }
...
}
```

Can anyone think of a reason to not call `Component::onEnter()` as well if the node is already running?

DelinWorks:
From what i think, components are considered entered when they enter a node, not a node containing them entering a node, because that's what we care about in the end, we're just components of that node so we don't care where that node ends up as long as we're inside it and our onEnter has been called, but nodes are at the same level so when they exit or enter in each other, their functions get triggered. hope that helps 🙏 

[...]

PR #867, opened 2022-09-27
Fix to ensure Component::onEnter() is called

rh101:
## Describe your changes
If a component is added to a parent node after the parent `Node::onEnter()` has been called, then the `Component::onEnter()` would never be called. 

This fix ensures that if the component is added while the parent node is already running, then the `Component::onEnter()` is called.

#866 


DelinWorks:
maybe remove the `Component::onEnter` from `Node::onEnter` since your're calling `Component::onEnter` anyways when it gets added

for reference:
```
void Node::onEnter() {
    if (_componentContainer && !_componentContainer->isEmpty()) {
        _componentContainer->onEnter();
    }
}
```
portion should be removed.

rh101:
> maybe remove the `Component::onEnter` from `Node::onEnter` since your're calling `Component::onEnter` anyways when it gets added

That code should not be removed, since it is still required.  Review the code carefully and you will see why.

The `Node::_running` flag is `false` until the end of the Node's onEnter() method, so if a component is added prior to `Node::onEnter()` being called, the only place `Component::onEnter()` will be called is in `Node::onEnter()`.  That's why the code is correct and should not be removed at all.

DelinWorks:
I see..

aismann:
Maybe a test (cpp-test) with different scenarios will be usefull too?

rh101:
> Maybe a test (cpp-test) with different scenarios will be usefull too?

I don't see any way to test this, as the side effect of the original code is that `onEnter()` is never called, which would cause undesired behaviour in a component.  This fix simply ensures that `Component::onEnter()` is called regardless of when the component is added to the parent.

The sub-classes that inherit from `Component` and override `Component::onEnter()` are `Physics3DComponent`, `NavMeshObstacle` and `NavMeshAgent`, all of which would be buggy if they weren't added to the parent node prior to the parent node `Node::onEnter()`.  The current test cases cover them all, and the changes here have no impact on their operation.

PR #905, opened 2022-10-08
Improve AutoPolygon::generateTriangles performance

halx99:
a. Avoid new many small p2t::Point
b. Add pod_vector for performance purpose


core/base/pod_vector.h
```diff
@@ -0,0 +1,164 @@
+/****************************************************************************
```
crazyhappygame, line 1: This class is super dangerous to use. Please do not commit this change.
Instead of we should change TrianglesCommand::Triangles to have std::vector inside

core/2d/CCAutoPolygon.cpp
```diff
@@ -616,38 +625,24 @@ TrianglesCommand::Triangles AutoPolygon::triangulate(const std::vector<Vec2>& po
...
-    unsigned short* indicesBuf = new unsigned short[indices.size()];
-    memcpy(indicesBuf, indices.data(), indices.size() * sizeof(short));
 
-    // Triangles should really use std::vector and not arrays for verts and indices.
-    // Then the above memcpy would not be necessary
-    TrianglesCommand::Triangles triangles = {vertsBuf, indicesBuf, (unsigned int)verts.size(),
-                                             (unsigned int)indices.size()};
+    // Good concept: the pod_vector support release internal buffer ownership
+    TrianglesCommand::Triangles triangles = {vertsBuf.release_pointer(), indicesBuf, (unsigned int)vdx,
+                                             (unsigned int)idx};
```
crazyhappygame, line 645: Please let's do what is written here

> Triangles should really use std::vector and not arrays for verts and indices

I am afraid that if we will introduce pod_vector then we will find many places in the code where it can be used. 
But the proper way here is to remove c style array and use std::vector everywhere

crazyhappygame (changes requested):

core/base/pod_vector.h
crazyhappygame: Please remove this class or at least create many unit tests to prove that it works as expected

core/2d/CCAutoPolygon.cpp
halx99: OK, will remove pod_vector use at here

core/base/pod_vector.h
```diff
@@ -0,0 +1,213 @@
...
```
crazyhappygame, line 1: Can  we remove this unused file?

core/base/pod_vector.h
halx99: At least, for some specific use cases, pod_vector faster than std::vector:
```txt
Runing perftest ...
ff######## Test round 1 ######### 
Testing xx_pod_vector...--> count: 49950000000, cost: 0.176476(s), realloc_hits: 0
Testing ax_pod_vector...--> count: 49950000000, cost: 0.103016(s), realloc_hits: 0
Testing std_vector...--> count: 49950000000, cost: 0.141327(s)


######## Test round 2 ######### 
Testing xx_pod_vector...--> count: 49950000000, cost: 0.168548(s), realloc_hits: 0
Testing ax_pod_vector...--> count: 49950000000, cost: 0.105139(s), realloc_hits: 0
Testing std_vector...--> count: 49950000000, cost: 0.140052(s)


######## Test round 3 ######### 
Testing xx_pod_vector...--> count: 49950000000, cost: 0.170028(s), realloc_hits: 0
Testing ax_pod_vector...--> count: 49950000000, cost: 0.103696(s), realloc_hits: 0
Testing std_vector...--> count: 49950000000, cost: 0.140359(s)
```

refer to: https://github.com/yasio/yasio/actions/runs/3213606435/jobs/5253397887

core/base/pod_vector.h
halx99: Yes, removed

crazyhappygame (approved):
Thank you

PR #907, opened 2022-10-09
Update CCTrianglesCommand.h to use std::vector

crazyhappygame:

## Describe your changes

Update CCTrianglesCommand.h to use std::vector - in progress

### DIFF
--- a/core/renderer/CCTrianglesCommand.h
+++ b/core/renderer/CCTrianglesCommand.h
@@ -52,20 +52,8 @@ public:
     /**The structure of Triangles. */
     struct Triangles
     {
-        Triangles(V3F_C4B_T2F* _verts, unsigned short* _indices, unsigned int _vertCount, unsigned int _indexCount)
-            : verts(_verts), indices(_indices), vertCount(_vertCount), indexCount(_indexCount)
-        {}
-
-        Triangles() {}
-
-        /**Vertex data pointer.*/
-        V3F_C4B_T2F* verts = nullptr;
-        /**Index data pointer.*/
-        unsigned short* indices = nullptr;
-        /**The number of vertices.*/
-        unsigned int vertCount = 0;
-        /**The number of indices.*/
-        unsigned int indexCount = 0;
+        std::vector<V3F_C4B_T2F> verts;
+        std::vector<short> indices;
     };
 
     /**Constructor.*/


halx99:
You needs do more work to replace with `std::vector`

rh101:
Won't this change have a negative impact on performance?

halx99:
Not sure, maybe need test?

rh101:
> Not sure, maybe need test?

Yes, it definitely needs to be tested thoroughly, because that section of code would impact a lot in the engine.

aismann:
@crazyhappygame 
Before this pr can be merged a Performance Test ist needed. (Own PR)

halx99 (changes requested):
Needs sync

crazyhappygame:
Changes it self for core looks easy.

But there are others non trivial parts like e.g.:
In spine extension there is similar structure.
```
namespace spine {
	struct V3F_C4B_C4B_T2F {
		ax::Vec3 position;
		ax::Color4B color;
		ax::Color4B color2;
		ax::Tex2F texCoords;
	};
	
	struct TwoColorTriangles {
		V3F_C4B_C4B_T2F* verts;
		unsigned short* indices;
		int vertCount;
		int indexCount;
	};
```

crazyhappygame:
Note: It looks like in many places PolygonInfo are duplicated by some helper variables. Writing directly to PolygonInfo will make code simpler and probably faster - less coping
```
    // vertex coords, texture coords and color info
    V3F_C4B_T2F_Quad _quad;
    V3F_C4B_T2F* _trianglesVertex   = nullptr;
    unsigned short* _trianglesIndex = nullptr;
    PolygonInfo _polyInfo;
```

crazyhappygame:
This change is bigger than anticipated. Closing for now. 
I think it would be good to back to this at some point

halx99:
so, pod vector is useful?

halx99:
there is a performance test about it: https://github.com/yasio/yasio/actions/runs/3213606435/jobs/5253397887

halx99:
the test code is here: https://github.com/yasio/yasio/blob/perftest/tests/perf/main.cpp

halx99:
test result:
```txt
1m 34s
D:\a\yasio\yasio\tests\perf\main.cpp(38): message : see reference to function template instantiation 'xx::PodVector<int,true>::~PodVector(void)' being compiled [D:\a\yasio\yasio\build\tests\perf\perftest.vcxproj]
D:\a\yasio\yasio\tests\perf\xx_podvector.h(45,4): warning C4984: 'if constexpr' is a C++17 language extension [D:\a\yasio\yasio\build\tests\perf\perftest.vcxproj]
D:\a\yasio\yasio\tests\perf\xx_podvector.h(44): message : while compiling class template member function 'T *xx::PodVector<T,true>::Alloc(const size_t &)' [D:\a\yasio\yasio\build\tests\perf\perftest.vcxproj]
          with
          [
              T=int
          ]
D:\a\yasio\yasio\tests\perf\xx_podvector.h(17): message : see reference to function template instantiation 'T *xx::PodVector<T,true>::Alloc(const size_t &)' being compiled [D:\a\yasio\yasio\build\tests\perf\perftest.vcxproj]
          with
          [
              T=int
          ]
  perftest.vcxproj -> D:\a\yasio\yasio\build\tests\perf\Release\perftest.exe
Runing perftest ...
ff######## Test round 1 ######### 
Testing xx_pod_vector...--> count: 49950000000, cost: 0.176476(s), realloc_hits: 0
Testing ax_pod_vector...--> count: 49950000000, cost: 0.103016(s), realloc_hits: 0
Testing std_vector...--> count: 49950000000, cost: 0.141327(s)


######## Test round 2 ######### 
Testing xx_pod_vector...--> count: 49950000000, cost: 0.168548(s), realloc_hits: 0
Testing ax_pod_vector...--> count: 49950000000, cost: 0.105139(s), realloc_hits: 0
Testing std_vector...--> count: 49950000000, cost: 0.140052(s)


######## Test round 3 ######### 
Testing xx_pod_vector...--> count: 49950000000, cost: 0.170028(s), realloc_hits: 0
Testing ax_pod_vector...--> count: 49950000000, cost: 0.103696(s), realloc_hits: 0
Testing std_vector...--> count: 49950000000, cost: 0.140359(s)
```

halx99:
checking new perftest result: https://github.com/yasio/yasio/actions/runs/3214356785/jobs/5254690328 after merge your changes.

crazyhappygame:
@halx99 Thank you for this interesting benchmark.

Indeed it looks like for this particular github PC configuration ax::pod_vector is ~50% faster. But I am not sure if that really important here.

For me the most important is code correctness. pod_vector with method release_pointer everything is more complicated. Additionally as name said in works only with POD types. For big projects sooner or later someone will add polymorphic class into this container ;(

I checked original code on my Windows 10 PC, msvc 2022 std::vector is ~10%.

After changing std::vector implementation to below on my PC std::vector is **50% faster** then ax::pod_vector.
```
    {
      /* std vector */
      printf("Testing std_vector...");
      count = 0;
      start = yasio::highp_clock();
      for (int i = 0; i < NUM_TIMES; ++i)
      {
        std::vector<int> pv(1000);
        std::iota(pv.begin(), pv.end(), 0);
        count += std::accumulate(pv.begin(), pv.end(), 0);
      }

      printf("--> count: %lld, cost: %lf(s)\n", count,
             (yasio::highp_clock() - start) / (double)std::micro::den);
    }
```

I'm curious what is the difference on other platforms (android, mac, ios, linux). Could you run original benchmark on other configurations?

halx99:
Actually, the release_pointer of pod_vector is specific use to avod gc alloc at some situations like: https://github.com/axmolengine/axmol/blob/dev/core/2d/CCAutoPolygon.cpp#L644

That's why we needs a tiny container with features: a. incredible buffer: std vector support
b. buffer can be transfer: std vector not support

halx99:
For 
```cpp
std::vector<int> pv(1000);
        std::iota(pv.begin(), pv.end(), 0);
        count += std::accumulate(pv.begin(), pv.end(), 0);
```

do you implement resize(with zero fill) and begin, end for pod_vector?

halx99:
I have add begin, resizefor pod_vector, and the latest test result:
https://github.com/yasio/yasio/actions/runs/3215269698/jobs/5256253115

crazyhappygame:
> 2\. buffer can be transfer: std vector not support

With std::vector you can transfer buffer from one to another std::vector (std::move, swap, etc). The problem is that in many places pointers are used. It is difficult to reason about ownership in this case. And std:: vector does not release ownership. 
I think it should be possible to use std:: vector here. I will make another attempt later this month.

Do we have any perf test to check regressions/improvements?

halx99:
Let's wait for your attempt fitst.

PR #910, opened 2022-10-09
Fix to UI touch location and add keyboard support for ImGui on Android

rh101:
## Describe your changes
Reduce code duplication and fix X touch location calculation.

Also added soft-keyboard support for ImGui text input fields.


rh101:
I don't have any more changes to make to this PR, unless there is any issue you need me to address.

halx99:
Nice work

Issue #915, opened 2022-10-10
Generic C++ code quality improvements (Long Terms Dev Plan)


halx99:
- [ ] vector use, and alloc for overwrite(alloc without zero fill), #916

aismann:
maybe label 'pinned' is useful

aismann:
@halx99,
C++ code quality improvements is an always/permanent issue.

Whith this knowledge, this issue can be closed (its not an "its done" issue one one date). 
Improvements will be always activ.

PR #916, opened 2022-10-10
Improve code quality (step1. Oct)

halx99:

## Describe your changes

- resize without filled
- resize fit

### Example code: 
```cpp
auto buf0 = std::vector<char>(128); // with zero filled and the real memory will large than required `128bytes`
auto buf1 = std::make_unique<char[]>(128); // alloc fit memory with zero filled
auto buf2 = std::make_unique_for_overwrite<char[]>(128); // alloc memory fit without zero filled.
```

### Reference links
- resize_and_overwrite:
  - https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1072r10.html
- boost vector resize with init behavior parameter
  - https://github.com/boostorg/container/blob/develop/include/boost/container/vector.hpp
- pod_vector
  - https://stackoverflow.com/questions/15219984/using-vectorchar-as-a-buffer-without-initializing-it-on-resize/15220853#15220853
  - https://github.com/yasio/yasio/blob/perftest/tests/perf/pod_vector.h

#915


core/3d/CCAnimation3D.cpp
```diff
@@ -113,8 +114,11 @@ bool Animation3D::init(const Animation3DData& data)
 
         if (iter.second.empty())
             continue;
-        std::vector<float> keys;
-        std::vector<float> values;
+
+        keys.clear();
+        values.clear();
+        keys.reserve(iter.second.size());
+        values.reserve(iter.second.size() * 3);
```
crazyhappygame, line 121: This change make code less readable. 
What do you think about following change
```
        std::vector<float> keys;
        std::vector<float> values;
        keys.reserve(iter.second.size());
        values.reserve(iter.second.size() * 3);
```
If this performance improvement probably ~2x faster then reserve+emplace_back will be something like
```
std::vector<float> keys(iter.second.size());
std::transform(iter.second.begin(), iter.second.end(), keys.begin(), [](const auto& keyIter){return keyIter._time;});
```

core/2d/CCMotionStreak.cpp
```diff
@@ -120,9 +121,7 @@ bool MotionStreak::initWithFade(float fade, float minSeg, float stroke, const Co
     _colorPointer = (uint8_t*)malloc(sizeof(uint8_t) * 4 * _vertexCount);
     _customCommand.createVertexBuffer(VERTEX_SIZE, _vertexCount, CustomCommand::BufferUsage::DYNAMIC);
 
-    std::vector<uint8_t> zeros;
-    zeros.resize(VERTEX_SIZE * _vertexCount);
-    std::fill(zeros.begin(), zeros.end(), 0);
+    yasio::byte_buffer zeros(VERTEX_SIZE * _vertexCount, (uint8_t)0, std::true_type{});
```
crazyhappygame, line 124:
```
std::vector<uint8_t> zeros(zeros.resize(VERTEX_SIZE * _vertexCount);
```
?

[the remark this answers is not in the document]
halx99: I prefer `std::make_unique<char[]>`

core/3d/CCAnimation3D.cpp
halx99: > This change make code less readable. What do you think about following change
> 
> ```
>         std::vector<float> keys;
>         std::vector<float> values;
>         keys.reserve(iter.second.size());
>         values.reserve(iter.second.size() * 3);
> ```
> 
> If this performance improvement probably ~2x faster then reserve+emplace_back will be something like
> 
> ```
> std::vector<float> keys(iter.second.size());
> std::transform(iter.second.begin(), iter.second.end(), keys.begin(), [](const auto& keyIter){return keyIter._time;});
> ```

And keys, values still could be reuse for 3 for-loops to avoid GC alloc

core/2d/CCMotionStreak.cpp
halx99: make_unique is good for alloc momory fit with zero filled, better than std vector

crazyhappygame (approved):

crazyhappygame:
Do we have any benchmark to prove that those changes improve performance?

halx99:
> > Do we have any benchmark to prove that those changes improve performance?
> 
> Not have yet, maybe at next step

I have test at local:

Before: 

```txt
benchmark `Animation3D_init_1` of cost: 1.827(ms)
benchmark `Animation3D_init_2` of cost: 1.759(ms)
benchmark `Animation3D_init_3` of cost: 1.499(ms)
```

After:

```txt
benchmark `Animation3D_init_1` of cost: 0.402(ms)
benchmark `Animation3D_init_2` of cost: 0.295(ms)
benchmark `Animation3D_init_3` of cost: 0.235(ms)
```

PR #918, opened 2022-10-10
Fix noisy message "Unable to resolve full path of PCH-header"

crazyhappygame:
fix noisy message "Unable to resolve full path of PCH-header"

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -158,7 +158,10 @@ if(XCODE OR VS)
 endif()
 
 message("CMake ${_AX_CORE_LIB} target_precompile_headers")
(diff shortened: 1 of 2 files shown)


core/CMakeLists.txt
```diff
@@ -158,7 +158,9 @@ if(XCODE OR VS)
 endif()
 
 message("CMake ${_AX_CORE_LIB} target_precompile_headers")
-target_precompile_headers(${_AX_CORE_LIB} PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:cocos2d.h>")
+target_precompile_headers(${_AX_CORE_LIB} PRIVATE 
+  "$<$<COMPILE_LANGUAGE:CXX>:cocos2d.h>"
```
rh101, line 162: @crazyhappygame Should `axmol.h` be added to this list given that, moving forward, it's the one we should be using?

PR #978, opened 2022-12-06
Move getWorldPosition() from Widget to Node class

rh101:
## Describe your changes

The method `Widget::getWorldPosition()` is convenient in calculating the position of an object in the scene, but this method should not be limited to UI nodes, since there is no UI specific data that it requires.

This change simply moves the method from the Widget class to the Node class, since it applies to any node object in the scene.


DelinWorks:
maybe you can add this change to `axmol/RENAMED_TYPES.md` so that people are aware.

PR #988, opened 2022-12-17
Update AXBuildSet.cmake

crazyhappygame:

## Describe your changes
Needed to build for linux platform from outside of axmol folder 

### DIFF
--- a/cmake/Modules/AXBuildSet.cmake
+++ b/cmake/Modules/AXBuildSet.cmake
@@ -15,6 +15,7 @@ if(CMAKE_TOOLCHAIN_FILE)
 endif()
 
 find_program(PYTHON_COMMAND NAMES python3 python2 python)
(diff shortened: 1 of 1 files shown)


rh101:
What do you mean by building from outside of the axmol folder?

crazyhappygame:
This change will set _AX_ROOT_PATH to axmol folder from AXBuildSet.cmake is included.

I thought that proper way to use axmol is to include AXBuildSet
```
    set(CMAKE_MODULE_PATH PATH_TO_AXMOL/cmake/Modules/)
    include(AXBuildSet)
```
Closing

rh101:
> thought that proper way to use axmol is to include AXBuildSet
> 
> ```
>     set(CMAKE_MODULE_PATH PATH_TO_AXMOL/cmake/Modules/)
>     include(AXBuildSet)
> ```

No, that's not the case. The `_AX_ROOT_PATH` must always be set by the game project, since its location is dependent on where the developer requires it to be.  It is always set prior to including AXBuildSet.

PR #999, opened 2022-12-29
[IMPROVEMENT] Add option to shift vertices rather than node when anchoring.

DelinWorks:

## Describe your changes
Refer to discussion #998
This change adds an extra parameter to the `setAnchorPoint` method in the sprite class, `useVertexAnchor` parameter which allows anchor points to apply at vertex level, not node level when set to true.

when `useVertexAnchor` is set to false i.e:
image
then the anchor will move the node itself making the sprite in the center of the scene (previously how it worked)
image

when `useVertexAnchor` is set to true i.e:
image
then the anchor will shift the sprite's vertices which allows us to add a child in the center of that sprite
image

you can also scale/rotate sprites without parenting them to another node because they're not being shifted at the node but rather at the vertex level, This change also makes it flexible to build UI elements while maintaining backward compatibility.

### DIFF
--- a/core/2d/CCSprite.cpp
+++ b/core/2d/CCSprite.cpp
@@ -867,8 +867,8 @@ void Sprite::setTextureCoords(const Rect& rectInPoints, V3F_C4B_T2F_Quad* outQua
 
 void Sprite::setVertexCoords(const Rect& rect, V3F_C4B_T2F_Quad* outQuad)
 {
-    float relativeOffsetX = _unflippedOffsetPositionFromCenter.x - getContentSize().x * _spriteVertexAnchor.x;
-    float relativeOffsetY = _unflippedOffsetPositionFromCenter.y - getContentSize().y * _spriteVertexAnchor.y;
+    float relativeOffsetX = _unflippedOffsetPositionFromCenter.x;
+    float relativeOffsetY = _unflippedOffsetPositionFromCenter.y;
 
     // issue #732
     if (_flippedX)
@@ -1327,20 +1327,10 @@ void Sprite::setPositionZ(float fVertexZ)
     SET_DIRTY_RECURSIVELY();
 }
 
-void Sprite::setAnchorPoint(const Vec2& anchor, bool useVertexAnchor)
+void Sprite::setAnchorPoint(const Vec2& anchor)
 {
-    if (useVertexAnchor)
-    {
-        _spriteVertexAnchor = anchor;
-        Node::setAnchorPoint({0, 0});
-        SET_DIRTY_RECURSIVELY();
-        updatePoly();
-    }
-    else
-    {
-        Node::setAnchorPoint(_spriteVertexAnchor = anchor);
-        SET_DIRTY_RECURSIVELY();
-    }
+    Node::setAnchorPoint(anchor);
+    SET_DIRTY_RECURSIVELY();
 }
 
 void Sprite::setIgnoreAnchorPointForPosition(bool value)
@@ -1696,9 +1686,10 @@ std::string Sprite::getDescription() const
     char textureDescriptor[100];
     if (_renderMode == RenderMode::QUAD_BATCHNODE)
         snprintf(textureDescriptor, sizeof(textureDescriptor), "<Sprite | Tag = %d, TextureID = %p>", _tag,
-                _batchNode->getTextureAtlas()->getTexture()->getBackendTexture());
+                 _batchNode->getTextureAtlas()->getTexture()->getBackendTexture());
     else
-        snprintf(textureDescriptor, sizeof(textureDescriptor), "<Sprite | Tag = %d, TextureID = %p>", _tag, _texture->getBackendTexture());
+        snprintf(textureDescriptor, sizeof(textureDescriptor), "<Sprite | Tag = %d, TextureID = %p>", _tag,
+                 _texture->getBackendTexture());
 
     return textureDescriptor;
 }
--- a/core/2d/CCSprite.cpp
+++ b/core/2d/CCSprite.cpp
@@ -867,8 +867,8 @@ void Sprite::setTextureCoords(const Rect& rectInPoints, V3F_C4B_T2F_Quad* outQua
 
 void Sprite::setVertexCoords(const Rect& rect, V3F_C4B_T2F_Quad* outQuad)
 {
-    float relativeOffsetX = _unflippedOffsetPositionFromCenter.x;
-    float relativeOffsetY = _unflippedOffsetPositionFromCenter.y;
+    float relativeOffsetX = _unflippedOffsetPositionFromCenter.x - getContentSize().x * _spriteVertexAnchor.x;
+    float relativeOffsetY = _unflippedOffsetPositionFromCenter.y - getContentSize().y * _spriteVertexAnchor.y;
 
     // issue #732
     if (_flippedX)
@@ -1327,10 +1327,21 @@ void Sprite::setPositionZ(float fVertexZ)
     SET_DIRTY_RECURSIVELY();
 }
 
-void Sprite::setAnchorPoint(const Vec2& anchor)
+void Sprite::setAnchorPoint(const Vec2& anchor, bool useVertexAnchor)
 {
-    Node::setAnchorPoint(anchor);
-    SET_DIRTY_RECURSIVELY();
+    if (useVertexAnchor)
+    {
+        _spriteVertexAnchor = anchor;
+        Node::setAnchorPoint({0, 0});
+        SET_DIRTY_RECURSIVELY();
+        updatePoly();
+    }
+    else
+    {
+        _spriteVertexAnchor = {0, 0};
+        Node::setAnchorPoint(anchor);
+        SET_DIRTY_RECURSIVELY();
+    }
 }
 
 void Sprite::setIgnoreAnchorPointForPosition(bool value)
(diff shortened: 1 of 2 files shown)


rh101:
As I mentioned in the discussion regarding this issue, what happens when with things like `getWorldPosition()` and touch related code when setting the last parameter to true? Do the coordinates that are calculated match up with the visual position of the sprite?

Thorough test cases should be added for this feature.

DelinWorks:
That's why I haven't marked the `thorough tests` check box

rh101:
> That's why I haven't marked the `thorough tests` check box

In that case, it is best not to create a PR until the code is entire code is ready in a separate branch, otherwise each commit will keep triggering the CI build and any other notifications related to changes to the PR.

DelinWorks:
> things like `getWorldPosition()` and touch related code when setting the last parameter to true? Do the coordinates that are calculated match up with the visual position of the sprite?

this only applies to `Sprite` objects and not any other object like `MenuItemImage` so the touch-related stuff is good also I checked that UI objects use custom rendering and do not inherit the `Sprite` class directly.
also the `getWorldPosition()` is good too because if you set that value to true you should expect the world position to be the position of the node even when the vertices are shifted.

rh101:
> this only applies to `Sprite` objects and not any other object like `MenuItemImage` so the touch-related stuff is good also I checked that UI objects use custom rendering and do not inherit the `Sprite` class directly.

I don't understand what you mean by that, or what relevance menu items have to it.  Any kind of node, Sprite or not, can have a touch event handler connected to it, so when that is done, does it still work as intended?

DelinWorks:
`Sprite`s as far as I know do not have touch handlers is there an example for it? or do you mean anything that inherits `Widget`

rh101:
Sorry, I'm not being clear.  UI elements have nothing to do with this at all, and not relevant to this in any way.  A `Sprite`, or any other sub-class of `Node`, can have a touch event handler connected to it so it can receive touch input.  Touch input is not only for UI elements.

For example:
```
auto* listener = EventListenerTouchOneByOne::create();
listener->onTouchBegan = [this](Touch* touch, Event* event) {
    const auto location = touch->getLocation();
    // handle it here
    return true;
};

_eventDispatcher->addEventListenerWithSceneGraphPriority(listener, targetNode);
```

`targetNode` can be anything; a `Node`, a `Sprite` etc etc.

DelinWorks:
are you sure the touch handler works for nodes? because wherever I click it registers a press is there a way to check if it hit the sprite? 

https://user-images.githubusercontent.com/45469625/210002350-250d2f26-fa85-4123-afa4-18477f6ec87b.mp4

rh101:
> are you sure the touch handler works for nodes? because wherever I click it registers a press is there a way to check if it hit the sprite?

This is used extensively in my released games using Cocos2d-x, and by extension, this engine, so yes, it does work.

By default all touch is passed through to the parent node, and you need to ensure you return true in the `onTouchBegan()` if the touch coordinate falls within the coordinates of the node it's connected to.

Call this on it to avoid passing input to anything behind that node:
`listener->setSwallowTouches(true);`

DelinWorks:
Thanks! I really didn't know this existed in the engine I was just using `hitTest` rather than touch handlers, Also the touch handling worked but the hit area was shifted by half its size I had to modify the code like this to fix it:

```
    auto* listener = EventListenerTouchOneByOne::create();
    listener->onTouchBegan = [this](Touch* touch, Event* event) {
        auto target = static_cast<Sprite*>(event->getCurrentTarget());

        Vec2 locationInNode = target->convertToNodeSpace(touch->getLocation());
        Size s = target->getContentSize();
        Rect rect = Rect(s.x * -0.5, s.y * -0.5, s.width, s.height);

        if (rect.containsPoint(locationInNode))
        {
            log("sprite began... x = %f, y = %f", locationInNode.x, locationInNode.y);

            return true;
        }
        return false;
    };

    _eventDispatcher->addEventListenerWithSceneGraphPriority(listener, sprite);
```

rh101:
> Also the touch handling worked but the hit area was shifted by half its size I had to modify the code like this to fix it:

Do you mean you needed to do that when setting the last parameter of `setAnchorPoint` to `true`, so the visual position of the sprite is not at its true coordinates?

DelinWorks:
I came to the conclusion that a function like `getContentSize()` should return a rect for content size rather than vec2 to specify the position of the anchor and its size.

DelinWorks:
> > Also the touch handling worked but the hit area was shifted by half its size I had to modify the code like this to fix it:
> 
> Do you mean you needed to do that when setting the last parameter of `setAnchorPoint` to `true`, so the visual position of the sprite is not at its true coordinates?

yes absolutely

rh101:
> I came to the conclusion that a function like `getContentSize()` should return a rect for content size rather than vec2 to specify the position of the anchor and its size.

The `getContentSize` method has a very specific meaning, and changing it would not make any sense at all.  Size of a 2D object is width and height, and nothing else; position and anchor point has nothing to do with it.

If anything you should be using `getBoundingBox()`, which does return a `Rect`.  If you need to read the anchor point of an object, then use `getAnchorPoint`, as it's there for that reason.

DelinWorks:
nice! this can work out for this change:
```
auto* listener = EventListenerTouchOneByOne::create();
listener->onTouchBegan = [this](Touch* touch, Event* event) {
    auto target = static_cast<Sprite*>(event->getCurrentTarget());

    Vec2 locationInNode = target->convertToNodeSpace(touch->getLocation());
    Size s = target->getContentSize();
    Size a = target->getAnchorPoint() * s;
    Rect rect = Rect(-a.x, -a.y, s.width, s.height);

    if (rect.containsPoint(locationInNode))
    {
        log("sprite began... x = %f, y = %f", locationInNode.x, locationInNode.y);

        return true;
    }
    return false;
};

_eventDispatcher->addEventListenerWithSceneGraphPriority(listener, sprite);
```

rh101:
I understand what you're trying to do here, but I genuinely feel that this is not the correct way to go about it.

If you have a specific need for sprites that are centered in their parent at position `0,0`, then the most appropriate way is to sub-class it and add all functionality in there via overrides of the relevant methods.  This would avoid messing with the `Sprite` class.  You would no longer require changing the signature of `setAnchorPoint()` or modify the calculations in `Sprite::setVertexCoords()`, since all of that would be in the overrides within the sub-class.

For instance, create a new class named `CenteredSprite`, which inherits from `Sprite`, and override the required methods.  This way, a developer can explicitly use a node of this type when it is required.  The added benefit to this is that it would automatically return correct values for positioning, which, for example, would avoid mucking about with the touch coordinates, as you've shown in the post above.

DelinWorks:
good idea! and I can create a bounding box function specific to that class right?

rh101:
> good idea! and I can create a bounding box function specific to that class right?

Nothing is stopping you from changing functionality in the sub-class, since it does not affect anything but itself.  It's the same way with the `Widget` and other sub-classes; they were created for specific reasons.  Do not be afraid of creating new sub-classes for things that work differently.

DelinWorks:
is this how you create a `create` method inside a class that inherits sprite? 

```
AnchoredSprite* AnchoredSprite::create(std::string_view filename)
{
    return (AnchoredSprite*)Sprite::create(filename);
}
```

rh101:
> is this how you create a `create` method inside a class that inherits sprite?
> 
> ```
> AnchoredSprite* AnchoredSprite::create(std::string_view filename)
> {
>     return (AnchoredSprite*)Sprite::create(filename);
> }
> ```

No, that is absolutely not the way.   In your sample above you're creating a base class into a subclass, which is not correct C++, and is going to result in an error.  You cannot typecast a base class into a subclass, and it doesn't even make sense to do so, but you can you can however cast a sub-class to a base class, although it's not relevant for what you're trying to do here.

All you need to do is look at `Sprite::create` to see what you need. They are static methods that you need to copy into your new class, and change all the types to your new `AnchoredSprite`.

Also, commits like [f4f2287](https://github.com/axmolengine/axmol/pull/999/commits/f4f228792fa94ef97520c82ef4857b3714b1da3f) should not be done.  Do not change formatting, since it's been explicitly done that way by the formatting settings included with the source in the `.editorconfig` file.  Do not override these settings and force them to your own.  You should set your Visual Studio to read the correct settings from `.editorconfig` for that source tree.

rh101:
As an example of a create method for your new class:

```
AnchoredSprite* AnchoredSprite::createWithTexture(Texture2D* texture)
{
    AnchoredSprite* sprite = new AnchoredSprite();
    if (sprite->initWithTexture(texture))
    {
        sprite->autorelease();
        return sprite;
    }
    AX_SAFE_DELETE(sprite);
    return nullptr;
}
```

Once again, copy **all** the `createXXXXXX` methods and re-implement them, and do not call the `Sprite::createXXXXX` from `AnchoredSprite`, because they're creating instances of the wrong type of class.

DelinWorks:
> Also, commits like [f4f2287](https://github.com/axmolengine/axmol/pull/999/commits/f4f228792fa94ef97520c82ef4857b3714b1da3f) should not be done. Do not change formatting, since it's been explicitly done that way by the formatting settings included with the source in the [.editorconfig](https://github.com/axmolengine/axmol/blob/dev/.editorconfig) file. Do not override these settings and force them to your own. You should set your Visual Studio to read the correct settings from `.editorconfig` for that source tree.

oh sorry, I thought it was my editor that made these changes 😅

aismann:
@DelinWorks should this PR not better chancel and an own sub class written?

DelinWorks:
yeah true this pr should close.

PR #1003, opened 2023-01-02
Update CCFileUtils-win32.cpp

crazyhappygame:

## Describe your changes

By default check "{s_exeDir}/Resources" if not exist fallback to s_workingDir

Before that change "{s_exeDir}/Resources" was used only when application was started from exe dir (s_workingDir == s_exeDir)

### DIFF
--- a/core/platform/win32/CCFileUtils-win32.cpp
+++ b/core/platform/win32/CCFileUtils-win32.cpp
@@ -111,13 +111,14 @@ bool FileUtilsWin32::init()
     _checkWorkingPath();
     _checkExePath();
 
-    bool startedFromSelfLocation = s_workingDir == s_exeDir;
-    if (!startedFromSelfLocation || !isDirectoryExistInternal(AX_PC_RESOURCES_DIR))
-        _defaultResRootPath = s_workingDir;
-    else
+    // By default check "{s_exeDir}/Resources" if not exist fallback to s_workingDir
+    _defaultResRootPath.reserve(s_exeDir.size() + AX_PC_RESOURCES_DIR_LEN);
+    _defaultResRootPath.append(s_exeDir).append(AX_PC_RESOURCES_DIR, AX_PC_RESOURCES_DIR_LEN);
+    bool startedFromSelfLocation = true;
+    if (!isDirectoryExistInternal(_defaultResRootPath))
     {
-        _defaultResRootPath.reserve(s_exeDir.size() + AX_PC_RESOURCES_DIR_LEN);
-        _defaultResRootPath.append(s_exeDir).append(AX_PC_RESOURCES_DIR, AX_PC_RESOURCES_DIR_LEN);
+        startedFromSelfLocation = false;
+        _defaultResRootPath     = s_workingDir;
     }
 
     bool bRet = FileUtils::init();


crazyhappygame:
@halx99 could you merge it?

rh101:
@crazyhappygame Are you sure this change is correct?

If I recall correctly, the working directory should always have priority over the executable directory, but with the change you have made, that is no longer the case.  

Why exactly have you made this change, and are you sure it's following the correct and expected behaviour?

rh101:
Consider the example with the change made in this PR:

1 - User installs application, which has a "Resources" sub-directory in the executable directory
2 - User changes the working directory of the executable
3 - User starts up executable, expecting it to be reading from the working directory, but instead with this change, it will ignore the working directory, since it's hard-coded to use the executable directory if the "Resources" sub-directory exists.

This is definitely incorrect behaviour.

halx99:
> Consider the example with the change made in this PR:
> 
> 1 - User installs application, which has a "Resources" sub-directory in the executable directory 2 - User changes the working directory of the executable 3 - User starts up executable, expecting it to be reading from the working directory, but instead with this change, it will ignore the working directory, since it's hard-coded to use the executable directory if the "Resources" sub-directory exists.
> 
> This PR definitely introduces incorrect behaviour.

Correct

crazyhappygame:
@rh101 @halx99 
In the case the Resources folder is copied to exe location I have following problems
1. Can not run app from different directory
Dir/my_app.exe
2. Can not run app with using shortcuts (e.g.: desktop shortcut )
3. Can not run my tests using ctest

In all above case exeDir!=workDir but Resources folder is in exeDir. I think that proper behavior is that by default Resources should be taken from folder relative to exeDir.

----
Generally, I do not understand few things around Win32 and current/working dir behavior 
1. why user would like to use resources from current directory at all if there are resources close to exe file? 
2. Why this behavior change (compare to cocos2d-x) is done for Win32 only and not for Linux (Linux copies Resources folder as well)?
3. Why we need this current/working dir behavior at all if always (by cmake) Resources are copied to exe directory?
4. What if user working/current dir is some random dir location without resources?
5. Generally, any logic around current/working dir could be flaky. E.g.: what if someone (e.g.: 3rd part lib ) changed current dir between process is created and current dir access?

rh101:
@halx99 will have the answer most of those questions, but here are my responses to a few:

> 4\. What if user working/current dir is some random dir location without resources?

That's fine, it's something the user would have changed, so not something you should be concerned about.  If a user does this, then what happens is their problem, not yours.  The best thing to do in that situation is to display a pop-up notifying the user that resources are missing, and exit the app.

> 5\. Generally, any logic around current/working dir could be flaky. E.g.: what if someone (e.g.: 3rd part lib ) changed current dir between process is created and current dir access?

What is concerning about this point is a 3rd party library modifying the working directory when none should ever do that, and if one does, then the more important question is why.  If something like that is happening, then something is seriously wrong, and that 3rd party library should not be used.

If a user decides to move the game resources into another folder, and set the working directory to that folder, then that is their choice.  They are explicitly choosing to do so, and restricting them from doing that doesn't seem like a good idea.

My only concern regarding this PR was to do with the changes to the working directory, which should always have priority if it's been set by the user on the executable properties.

crazyhappygame:
@halx99 any comments?

crazyhappygame:
@rh101 Thank you for reply. 

> 1 - User installs application, which has a "Resources" sub-directory in the executable directory
> 2 - User changes the working directory of the executable
> 3 - User starts up executable, expecting it to be reading from the working directory, but instead with this change, it will ignore the working directory, since it's hard-coded to use the executable directory if the "Resources" sub-directory exists.

Why user should change working directory to use Resources from that directory? Why not copy "Resources" or cheaper change symlink to Resource folder?

rh101:
> @rh101 Thank you for reply.
> 
> > 1 - User installs application, which has a "Resources" sub-directory in the executable directory
> > 2 - User changes the working directory of the executable
> > 3 - User starts up executable, expecting it to be reading from the working directory, but instead with this change, it will ignore the working directory, since it's hard-coded to use the executable directory if the "Resources" sub-directory exists.
> 
> Why user should change working directory to use Resources from that directory? Why not copy "Resources" or cheaper change symlink to Resource folder?

The current working directory has nothing to do with Axmol or the resources. In Windows, a user can modify the working directory of an executable for whatever reason they may have, and the application is expected to use that directory.  What the user does or doesn't do it with working directory or resources is not really an issue we need to concern ourselves with.

rh101:
> Is it possible to change priority of dir to exeDir?

This is not the fix, but I actually agree with you regarding the fact that there is an issue that needs to be addressed here.

On the Windows platform, axmol seems to be forcing the resources to be mapped to the same folder as the executable, which I don't feel is correct.  

I had not given this issue much thought since my primary target platforms are Android and iOS, but the majority of my development is done using the Windows build in Visual Studio.  The app works when running it from inside the debugger, but when running the exe directly from the output folder, it doesn't work, since it was trying to look for the resources in the current folder, and they don't exist because of the changes done to axmol a while back.  

The file [run.bat.in](https://github.com/axmolengine/axmol/blob/dev/templates/cpp-template-default/run.bat.in) was added to generate a batch script that allowed the application to run correctly, but this seems to be covering up the real issue, which is that the developer doesn't have a choice in where the engine is reading the resources from. 

So, the solution to this would be to add functionality to the engine to allow the developer to set exactly where the resources are read from, whether it be current exe directory, or a sub-directory named "Resources", "data" or whatever else.  What is not correct is forcing the exe directory to have priority over the working directory; that is not the fix.

crazyhappygame:
@rh101 yes, one of the solution would be to add functionality to provide Resources path. Current solution based Working/current dir has problems describe above.

Another option would be to always use Resources from exeDir (it works in that way for cocos2d-x). To use different Resources you can copy files or use symlink

halx99:
All platforms woking directory is priority for system find file, so we don't change system behavior, just add exeDir to search path when work dir not equal exe dir.

rh101:
Would this work:

Say that we still copy the "Resources" sub-directory to the executable directory, and "Resources" is added to the search path via `FileUtils::addSearchPath`.  Would this solve the issue?

crazyhappygame:
@halx99 not Tru for Linux. Please check https://github.com/axmolengine/axmol/blob/dev/core/platform/linux/CCFileUtils-linux.cpp#L65 always exe path.

But ok. I will explicitly set search path in all my apps for windows

crazyhappygame:
> Would this work:
> 
> Say that we still copy the "Resources" sub-directory to the executable directory, and "Resources" is added to the search path via `FileUtils::addSearchPath`. Would this solve the issue?

@rh101 currently we (cmake) always copy Resources folder to exe dir

rh101:
> @rh101 currently we (cmake) always copy Resources folder to exe dir

I understand, so it would be like this:
```
game.exe
lib1.dll
lib2.dll
Resources/resources1.xyz
               /resource2.xyz etc etc
```

If you call this:

`FileUtils::getInstance()->addSearchPath("Resources", true);`

Would that work?

crazyhappygame:
I have added something similar to
```FileUtils::getInstance()->addSearchPath(exeDir+"/Resources", true);``` to all my apps.

Question is if we would like to have similar default behavior for axmol?

rh101:
> I have added something similar to `FileUtils::getInstance()->addSearchPath(exeDir+"/Resources", true);` to all my apps.
> 
> Question is if we would like to have similar default behavior for axmol?

It shouldn't be default behaviour, otherwise it would be forcing developers to use the "Resources" folder, which is what we're trying to avoid here as well.

For example, a developer may want resources to be in a different folder, such as "data".  It would be up to the developer to make that call with the correct folder name.

crazyhappygame:
> It shouldn't be default behaviour, otherwise it would be forcing developers to use the "Resources" folder, which is what we're trying to avoid here as well.

That what I do not understand. Using Working/current dir is tricky - 

But ok. I will close this PR and leave with settings search path in my application.

halx99:
> @halx99 not Tru for Linux. Please check https://github.com/axmolengine/axmol/blob/dev/core/platform/linux/CCFileUtils-linux.cpp#L65 always exe path.
> 
> But ok. I will explicitly set search path in all my apps for windows

Yes, current, axmol only avoid resources copy in development time on windows dev workflow, should cover PC platforms

PR #1014, opened 2023-01-05
[DRAFT DO NOT MERGE] Check BUILD_SHARED_LIBS ON for win32

crazyhappygame:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,9 @@
 cmake_minimum_required(VERSION 3.10)
 
 project(axmol)
+if(WIN32)
+    set(BUILD_SHARED_LIBS ON)
+ENDIF()
 
 if(XCODE)
     set(CMAKE_XCODE_GENERATE_TOP_LEVEL_PROJECT_ONLY TRUE)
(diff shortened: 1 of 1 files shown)


halx99:
Will fixed by https://github.com/axmolengine/axmol/commit/640dda61ec89ef56fb12abb03948ae0ffbdfd598

crazyhappygame:
@halx99 Thank you. You are super fast!

Please check https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html?highlight=target_file_dir#genex:TARGET_RUNTIME_DLLS

It looks like coping dependent DLL could be done with 1 line

crazyhappygame:
BTW could you remove as well STATIC and SHARED from all add_library calls? In that way we can control all libraries with BUILD_SHARED_LIBS option

halx99:
Some thirdparty libs doesn't support dll, so we can't do that

halx99:
> @halx99 Thank you. You are super fast!
> 
> Please check https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html?highlight=target_file_dir#genex:TARGET_RUNTIME_DLLS
> 
> It looks like coping dependent DLL could be done with 1 line

done https://github.com/axmolengine/axmol/commit/e9c1726c31b2b2c394510b3019d36abd41552787

crazyhappygame:
@halx99 Thank you!

PR #1022, opened 2023-01-08
Disable ALSOFT

crazyhappygame:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,8 @@ cmake_minimum_required(VERSION 3.10)
 
 project(axmol)
 
+SET(AX_USE_ALSOFT OFF CACHE BOOL "")
+
 if(XCODE)
     set(CMAKE_XCODE_GENERATE_TOP_LEVEL_PROJECT_ONLY TRUE)
 endif()


halx99:
You can't disable ALSOFT, only apple platform can disable it, then use system OpenAL.framework, other platforms, the ALSOFT is mandatory

crazyhappygame:
Thank you for clarifying.
But it looks like Android compiled.

What can be disabled to reduce android binary size. After switch to axmol size of my game increases from 15MB to 20MB

halx99:
image

crazyhappygame:
@halx99 It looks like some option should not be option

AX_USE_ALSOFT - can not be disabled
AX_WITH_BULLET - build fails

crazyhappygame:
Any hints what can be turnoff to reduce Android size?

halx99:
https://github.com/axmolengine/axmol/blob/dev/CMakeOptions.md

crazyhappygame:
Thank you for answers.

I try to understand changes axmol vs cocos2d-x that can contribute to my game size increases from ~15MB to ~20MB.

What are the other libraries added to axmol (except alsoft) compare to cocos2d-x for Android platform?
Do I get this right that curl is used for Android instead of native implementation? Is it possible to use native Android instead of curl?

rh101:
Try this utility: [https://github.com/google/bloaty](https://github.com/google/bloaty)

Other than that, Axmol has diverged quite a lot from Cocos2d-x, so comparing the final binary size of both makes little sense.

You could easily compare the extensions folder and such between Axmol and Cocos2d-x to know what libraries have been changed, but even then, knowing what has changed won't make any difference in the end.

crazyhappygame:
> Try this utility: https://github.com/google/bloaty
> 
> Other than that, Axmol has diverged quite a lot from Cocos2d-x, so comparing the final binary size of both makes little sense.
> 
> You could easily compare the extensions and thirdparty folders between Axmol and Cocos2d-x to know what libraries have been changed, but even then, knowing what has changed won't make any difference in the end. Just disable the extensions/libraries that you don't need, and move on from there.

Question is what and how can be disabled

PR #1031, opened 2023-01-14
Update CMakeLists.txt to be more dynamic and take all cpp and h files from the classes folder

Obg1:
Update CMakeLists.txt to be more dynamic and take all cpp and h files from the classes folder.
This will make it easier when adding new files, you won't need to update the cmakelists.txt file

### DIFF
--- a/templates/cpp-template-default/CMakeLists.txt
+++ b/templates/cpp-template-default/CMakeLists.txt
@@ -75,13 +75,12 @@ if(APPLE OR WINDOWS)
 endif()
 
 # add cross-platforms source files and header files
-list(APPEND GAME_SOURCE
-     Classes/AppDelegate.cpp
-     Classes/HelloWorldScene.cpp
+file(GLOB_RECURSE GAME_SOURCE
+     Classes/*.cpp
      )
-list(APPEND GAME_HEADER
-     Classes/AppDelegate.h
-     Classes/HelloWorldScene.h
+
+file(GLOB_RECURSE GAME_HEADER
+     Classes/*.h
      )
 
 if(ANDROID)


rh101:
GLOB usage is not recommended, and the reasons are stated in the CMake documentation [here](https://cmake.org/cmake/help/latest/command/file.html#glob-recurse).  Even if you decided to use `CONFIGURE_DEPENDS` to solve the issues introduced by using GLOB, `CONFIGURE_DEPENDS` comes with a significant performance penalty for large projects.

While it may seem *easier*, it's not a good enough reason to use it in projects.

Using it in the cpp-template is fine, but there should be a comment made so that new users understand the issues with GLOB.

Obg1:
@rh101 

For the dot underscore we can filter the list, something like this:
list(FILTER GAME_SOURCE EXCLUDE REGEX "._*")

or just not include them at all, by doing:

file(GLOB GAME_SOURCE [a-z]*.cpp)

So it will only pick up files with a-z characters

rh101:
That was just an example of the issues introduced by the use of GLOB, but not the only ones.  

> For the dot underscore we can filter the list, something like this: list(FILTER GAME_SOURCE EXCLUDE REGEX "._*")
> 
> or just not include them at all, by doing:
> 
> file(GLOB GAME_SOURCE [a-z]*.cpp)
> 
> So it will only pick up files with a-z characters


Adding any kind of work around for something like GLOB, usage of which is recommended against by the very developers that create cmake, seems to be adding another layer of complexity to what may have seemed "easy" to start with.

Once again, having GLOB in the cpp-template may be okay, as more experienced developers can decide on their own whether to use GLOB or not, but any new developers who are not experienced with cmake may end up running into issues because of GLOB that they won't know how to resolve.

Personally, I'm completely against the usage of GLOB in my own projects, regardless of how convenient its usage may seem to be.

PR #1040, opened 2023-01-30
Add Anchored Sprite Class

DelinWorks:

## Describe your changes
added anchored sprite, gives a much better anchoring alternative compared to the traditional method of anchoring the node

Example Code:
```
// add "HelloWorld" splash screen"
AnchoredSprite* sprite = AnchoredSprite::create("HelloWorld.png"sv);
if (sprite == nullptr)
{
    problemLoading("'HelloWorld.png'");
}
else
{
    // position and anchor the sprite on the center of the screen, this is an override function
    sprite->setAnchorPoint({ 1,0.5 });

    sprite->setPosition(Vec2(visibleSize.width / 2 + origin.x, visibleSize.height / 2 + origin.y));
    sprite->addChild(label, 1);
    label->setPosition(0, 0);
    // add the sprite as a child to this layer
    this->addChild(sprite, 0);
    auto drawNode = DrawNode::create();
    drawNode->setPosition(Vec2(0, 0));
    addChild(drawNode);

    drawNode->drawRect(safeArea.origin, safeArea.origin + safeArea.size, Color4F::BLUE);
}
```
Example with Touch Listener:
```
auto* listener = EventListenerTouchOneByOne::create();
listener->onTouchBegan = [this](Touch* touch, Event* event) {
    auto target = static_cast<AnchoredSprite*>(event->getCurrentTarget());

    Vec2 locationInNode = target->convertToNodeSpace(touch->getLocation());
    Rect rect = target->getTouchRect();

    if (rect.containsPoint(locationInNode))
    {
        log("sprite began... x = %f, y = %f", locationInNode.x, locationInNode.y);

        return true;
    }
    return false;
    };

_eventDispatcher->addEventListenerWithSceneGraphPriority(listener, sprite);
```

I added the function `getTouchRect` which deals with content size and rect offsets to make touch listener work.
 
### DIFF
--- a/core/2d/CCAnchoredSprite.cpp
+++ b/core/2d/CCAnchoredSprite.cpp
@@ -1,0 +1,210 @@
+/****************************************************************************
+Copyright (c) 2008-2010 Ricardo Quesada
+Copyright (c) 2010-2012 cocos2d-x.org
+Copyright (c) 2011      Zynga Inc.
+Copyright (c) 2013-2016 Chukong Technologies Inc.
+Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
+Copyright (c) 2020 C4games Ltd.
+Copyright (c) 2021-2022 Bytedance Inc.
+
+https://axmolengine.github.io/
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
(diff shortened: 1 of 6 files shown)


aismann:
@DelinWorks 
Why not add your example code to cpp_tests?

DelinWorks:
Don't worry! working on it

aismann:
@halx99 ,
Will it be part of the native Axmol engine?
It's a very special feature, but maybe someone can use it.

crazyhappygame:
@DelinWorks Could you explain how AnchoredSprite is different from Sprite and what problems we would like to solve with this class?

DelinWorks:
@crazyhappygame If you try to add children to a sprite with an anchor different than 0.5 than it will be frustrating because the node system moves the entire node instead of its vertices, so this class anchors the vertices and doesn't do anything to the node class, refer to #998 to learn more or you can take a look at the sprite test and see what I mean.

crazyhappygame:
OK. Thank you. So for AnchoredSprite child nodes (0,0) is anchor point of parent. 
With Sprite we can achieve the same by adjusting position ```child->setPosition(parent->getContentSize())```

aismann:
@DelinWorks 
Maybe the title of the tester is not correct?
image
"Testing AnchoredSprite" is more self explaining

Issue #1046, opened 2023-02-07
 Visible artifacts on DrawNode::drawCircle() with a corresponding lineWidth > 5 


aismann:
Steps to Reproduce:
![image](https://user-images.githubusercontent.com/8652787/217216261-b9ebe4b0-7829-4f7d-80de-f978afca850c.png)

aismann:
With this fix:
cpp_test  has a scale_factor bug (i belive) -> if window is resizing.
https://user-images.githubusercontent.com/8652787/217228819-9371f0cd-e935-4418-8a4e-6bc714d8421c.png

aismann:
Or another possible reason:
cpp_tests Node::Draw: 3. Rendering drawCircle itself has a bug ;)

aismann:
works corrcet now:
![image](https://user-images.githubusercontent.com/8652787/217739178-7aacb0c2-4096-4bf8-a4da-a82b6a0b2876.png)

halx99:
seems correct in current version: 
![image](https://user-images.githubusercontent.com/6977319/218024895-7084862d-0fa9-48bf-963f-fc054d95c60c.png)

aismann:
> seems correct in current version: ![image](https://user-images.githubusercontent.com/6977319/218024895-7084862d-0fa9-48bf-963f-fc054d95c60c.png)

The green rings looks bigger (nearer on the subtitle)
I can clicking the mousebutton on the window title and the "rendering is stopping for better comparing)

halx99:
Seems it's the drawPolygon behavior, when the border large, the circle looks large

aismann:
the current implementaion:
![image](https://user-images.githubusercontent.com/8652787/218034771-d90cdc07-e9a0-4501-b799-59009185c017.png)

the PR
![image](https://user-images.githubusercontent.com/8652787/218034415-48b289bb-91d7-472d-b347-5afc1043dfdc.png)

aismann:
@halx99  and @rh101 look here for description of the "internal: drawCircle render improvement:

As long the "line width < 5" both using the same faster render method
![image](https://user-images.githubusercontent.com/8652787/218143196-36a8c719-ee28-4c6d-9370-9f05bc72dcab.png)
If the "line width >= 5" left (GREEN) using the smoother internal method (automaticfly) RED use the faster but ugly method. 
![image](https://user-images.githubusercontent.com/8652787/218143980-d62562b1-3292-40a8-92a5-3895aecb3cbb.png)

[...]

PR #1047, opened 2023-02-07
Update CCGLView.cpp

aismann:

## Describe your changes
// Set the "correct" 'content scale factor' 

#1046

Can be testet with 31. Node: Draw => 3: Rendering drawCircle


aismann:
Seems cpp_test has an "scalefactor" bug.
I have to check it first

image

aismann:
I have to check something

rh101:
I realise this has been closed, but just an additional remark.  The `setContentScaleFactor` should not be called by the engine, since it's based on other factors that only the game knows about.  It's perhaps best to only send out an event when the window size changes (I think it already does?), and let the game set the content scale factor to whatever value it needs to be when it receives that event.

PR #1050, opened 2023-02-09
 Visible artifacts on DrawNode::drawCircle() with a corresponding lineWidth > 5 (window resizing issue)

aismann:

## Describe your changes
[Fix DrawNode::drawCircle() behavior itf the window is resizing.]
I tested it with cpp-test again => no effects on other nodes, objects or other stuff.

#1046


core/2d/CCDrawNode.cpp
```diff
@@ -369,8 +369,8 @@ void DrawNode::drawCircle(const Vec2& center,
     }
     if (_lineWidth > minThickness)
     {
-        auto scaleFactor = AX_CONTENT_SCALE_FACTOR() * 2 ;
-        drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth / scaleFactor, color);
+        static GLView* glView = _director->getOpenGLView();
```
halx99, line 372: please don't use static

core/2d/CCDrawNode.cpp
```diff
@@ -369,8 +369,8 @@ void DrawNode::drawCircle(const Vec2& center,
...
+        drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth / (glView->getScaleX() + glView->getScaleY()), color);
```
halx99, line 373: I think, you can use _lineWidth directly, don't use any scale from glview

halx99 (changes requested):

core/2d/CCDrawNode.cpp
aismann: I used it for this reason:
glView will be "only one" initialized (not every call) => faster and it will be done only if the _lineWidth > minThickness.

core/2d/CCDrawNode.cpp
aismann: I have improve the tester on cpp_test for better see the "improvment".
Its an own PR.

aismann: > I think, you can use _lineWidth directly, don't use any scale from glview

It is not working correct only using the _lineWidth without scaling it.

halx99: In your test case, you set the border so much large, right?

aismann: > In your test case, you set the border so much large, right?

Yes. its a test case to see it works also with a "big" _linewith too.
Maybe not a real situation...but a worst case which have to test.

halx99: Your code based on cpp-tests with frameSize(960, 640) design size(480, 320), it's very specific, when you change both of them to 1280x720, then looks good without scale, and you scale code looks a little magic.

halx99: image

aismann: Its magic ;)

halx99: image

This maybe better.

aismann:
```
        GLView* glView = _director->getOpenGLView();
        auto borderScale = (glView->getScaleX() + glView->getScaleY());
        drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth / borderScale , color);
```

Only this is working with different "window settings like:
```
static Size resourceSize(960, 640);
static Size resourceSize(1280, 720);

static Size designSize(480, 320);
static Size designSize(400, 320);
```

aismann: @halx99 
I change it like this and PR will be merged?
```
GLView* glView = _director->getOpenGLView();
        auto borderScale = (glView->getScaleX() + glView->getScaleY());
        drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth / borderScale , color);
```

halx99: Yes

aismann:
Hope all questions be closed

core/2d/CCDrawNode.cpp
```diff
@@ -369,8 +369,9 @@ void DrawNode::drawCircle(const Vec2& center,
     }
     if (_lineWidth > minThickness)
     {
-        auto scaleFactor = AX_CONTENT_SCALE_FACTOR() * 2 ;
-        drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth / scaleFactor, color);
+        GLView* glView = _director->getOpenGLView();
+        auto borderScale = (glView->getScaleX() + glView->getScaleY());
```
halx99, line 373:
```suggestion
        auto borderScale = (glView->getScaleX() + glView->getScaleY()) / 2;
```

core/2d/CCDrawNode.cpp
```diff
@@ -369,8 +369,9 @@ void DrawNode::drawCircle(const Vec2& center,
...
+        drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth / borderScale , color);
```
halx99, line 374:
```suggestion
        drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth * borderScale , color);
```

halx99 (changes requested):

core/2d/CCDrawNode.cpp
aismann: With `_lineWidth * borderScale`i getting this result:

image

halx99: Because the resourceSize is 960x640, and designSize is 480, 320 then the scale is 2.0, it's correct behavior

aismann: The green ring should have the same behavior like the red ring but without the rendering issue.
The green ring "is the red ring" without the bad rendering issue.

aismann: @halx99 
You anderstand what I mean?
The green is only needed to make a better rendering if the _lineWidth to big and normaly the "red circle show renderuing issues.

halx99: If I set both resourceSize and designSize to 1280x720, and limit the `thick` < 10, then looks like:
image

halx99: And 960x640, 480x320 looks like:
image

aismann: The green api is only a workaourond if the red api not rendering as the developer want (in most cases the thick will be "1-2' pixel)

halx99: Both 960x640:
image

All my screeshot use follow code:
```cpp
void DrawNode::drawCircle(const Vec2& center,
                          float radius,
                          float angle,
                          unsigned int segments,
                          bool drawLineToCenter,
                          float scaleX,
                          float scaleY,
                          const Color4B& color,
                          float minThickness)
{
    const float coef = 2.0f * (float)M_PI / segments;

    auto vertices = _abuf.get<Vec2>(segments + 2);

    auto scaleFactor = AX_CONTENT_SCALE_FACTOR();

    for (unsigned int i = 0; i <= segments; i++)
    {
        float rads    = i * coef;
        vertices[i].x = radius * cosf(rads + angle) * scaleX * scaleFactor + center.x;
        vertices[i].y = radius * sinf(rads + angle) * scaleY * scaleFactor + center.y;
    }
    if (_lineWidth > minThickness)
    {
        GLView* glView = _director->getOpenGLView();
        auto borderScale = (glView->getScaleX() + glView->getScaleX()) / 2.0f;
        drawPolygon(vertices, segments, Color4B(0.0f, 0.0f, 0.0f, 0.0f), _lineWidth * borderScale, color);
    }
    else
    {
        if (drawLineToCenter)
        {
            vertices[segments + 1] = center;
            drawPoly(vertices, segments + 2, true, color);
        }
        else
            drawPoly(vertices, segments + 1, true, color);
    }
}
```

aismann: > And 960x640, 480x320 looks like: image

The red looks good. so the green is not needed ;)

aismann: The example on th cpp-test using 2 as parameter maybe thats too smal better I use 4 or 5

halx99: Try without any scale: 960x640, resizing window:
image

image

image

halx99: code is:
```cpp
void DrawNode::drawCircle(const Vec2& center,
                          float radius,
                          float angle,
                          unsigned int segments,
                          bool drawLineToCenter,
                          float scaleX,
                          float scaleY,
                          const Color4B& color,
                          float minThickness)
{
    const float coef = 2.0f * (float)M_PI / segments;

    auto vertices = _abuf.get<Vec2>(segments + 2);

    auto scaleFactor = AX_CONTENT_SCALE_FACTOR();

    for (unsigned int i = 0; i <= segments; i++)
    {
        float rads    = i * coef;
        vertices[i].x = radius * cosf(rads + angle) * scaleX * scaleFactor + center.x;
        vertices[i].y = radius * sinf(rads + angle) * scaleY * scaleFactor + center.y;
    }
    if (_lineWidth > minThickness)
    {
        drawPolygon(vertices, segments, Color4B(0.0f, 0.0f, 0.0f, 0.0f), _lineWidth, color);
    }
    else
    {
        if (drawLineToCenter)
        {
            vertices[segments + 1] = center;
            drawPoly(vertices, segments + 2, true, color);
        }
        else
            drawPoly(vertices, segments + 1, true, color);
    }
}
```

halx99: So, finally, remove any scale of it?

aismann: Oh dear,
You misunderstand the use case for the green API.

Its only needed is the thick so many as the red api will be wrong rendering.
The developer have to check it and change the api (if it needed).
The example is bad for this: => "must" have to do which is part of the developer

aismann: In your last three examples the red api is working good enough and it is faster as the green api too.

aismann: > So, finally, remove any scale of it?

without any scale is look really often bad  (i belive)

halx99:
Let's @rh101  help us.

core/2d/CCDrawNode.cpp
aismann: Last but not least: In most cases is the red api working well.
Only if the "programmwer" want a thick > 10 or so and using circles the rendering goes badly.
Thast is not a no go...here can the green api making a better "look" of the ring. thats all.

aismann: The example is bad for "comparing" because the developer is using only the green OR the red api

rh101:
I'm having a lot of trouble following what this is all about.

Can you please summarize, concisely, what the purpose of this change is?  

> If the thick allways smaller as the red api rendering going bad so the green api is never needed ;)
> 
> Look on the result of the "red" API if the thickness >10 or more. Thats the reason implementing the green API. (And I want not write a new Api because the drawCircle is connected with the global thicknees. Of course developer can use always a new DrawNode object but this make the rendering slower.

This is confusing.  Are you stating that we need to use one API if thickness is less or equal to 10, and a different API for thickness >10?

aismann:
I will answer later

aismann:
please look here: #1046

rh101:
@aismann A few questions related to your changes:

1 - Will your changes result in a correct representation of a circle being drawn (at the correct scale), with no visible artifacts, regardless of line thickness?

2 - If the answer to the previous question is "Yes", then is the cause of all the confusion in this thread related to this specific test implementation?  

Now that I'm looking at this specific issue, I do see a problem, which is that the API may also be an issue.  It may have been better to leave `drawCircle` as it was previously (without the `minThickness` parameter), and a new method with a different name added for the optimized circle drawing (like `drawCircleOptimized` etc).  The `minThickness` parameter also doesn't mean what a developer may think it means without reading the comments in the code, so a better name for that would be beneficial to reflect its true purpose.

aismann:
> 1 - Will your changes result in a correct representation of a circle being drawn (at the correct scale), with no visible artifacts, regardless of line thickness?
> 
Yes, but its not so fast as the 'original' drawCircle() method which works good if the thickness <5.

> 2 - If the answer to the previous question is "Yes", then is the cause of all the confusion in this thread related to this specific test implementation?
> 
Yes and No. This PR is optimizing the situation if the window is resizing.
Yes: 
The visual comparing is the reason for so much "confusion". If you comment the 'original' drawCircle() line the 'understand will be better and its having not so much questions here ;)
No: 
This PR is optimizing the situation if the window is resizing. And one of the goal is having the same "thickness" as the original drawCircle() method.  That's the reason to show 'both' drawCircle methods on the same scene.


> Now that I'm looking at this specific issue, I do see a problem, which is that the API may also be an issue. It may have been better to leave `drawCircle` as it was previously (without the `minThickness` parameter), and a new method with a different name added for the optimized circle drawing (like `drawCircleOptimized` etc). The `minThickness` parameter also doesn't mean what a developer may think it means without reading the comments in the code, so a better name for that would be beneficial to reflect its true purpose.

I think having a `drawCircleOptimized`  method will write source code like this:
```
if (_lineWidth < 5) 
    drawCircle(...);
else
    drawCircleOptimized(...);
```
That's the same as using `drawCircle(... , /*minThickness */ 5);`
Of course the `minThickness` name can be missunderstanding but this example above is helping to understand what it mean
The tester helps too: Shows how its works.

And last but not least: `minThickness` name can be changed also.
I think cocos2dx/axmol having more such APIs which need to read the 'API description for a better understanding what it is doing.

rh101:
I understand what you mean now, and yes, it may be that a lot of the confusion is coming from that specific test.

Now, related to the API, you make a fair point regarding the optimization in that method, and what do you think about changing `minThickness` to `minOptimizedLineWidth` or something similar would make more sense?

aismann:
> I understand what you mean now, and yes, it may be that a lot of the confusion is coming from that specific test.
> 
> Now, related to the API, you make a fair point regarding the optimization in that method, and what do you think about changing `minThickness` to `minOptimizedLineWidth` or something similar would make more sense?

I think `threshold` is the best name for this 'optional' parameter.

rh101:
> I think `threshold` is the best name for this 'optional' parameter.

That's actually a good idea.  Perhaps you can add more to it so to give it some context, such as "optimizationThreshold" etc, which in a way would indicate what that parameter is being used for.

aismann:
> Now, related to the API, you make a fair point regarding the optimization in that method, and what do you think about changing `minThickness` to `minOptimizedLineWidth` or something similar would make more sense?

I will write some more about this parameter on the APPENDIX.md

halx99:
The scale code still confuse me:

```cpp
 auto borderScale = (glView->getScaleX() + glView->getScaleY());
drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth / borderScale , color);
```

If glView scale smaller then the borderWidth will be larger?

halx99:
> The scale code still confuse me:
> 
> ```c++
>  auto borderScale = (glView->getScaleX() + glView->getScaleY());
> drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth / borderScale , color);
> ```
> 
> If glView scale smaller then the borderWidth will be larger?

My suggestion is don't relay on glView scale just use follow code:

```cpp
drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth, color);
```

aismann:
OK. I will make a changes first.

aismann:
> > The scale code still confuse me:
> > ```c++
> >  auto borderScale = (glView->getScaleX() + glView->getScaleY());
> > drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth / borderScale , color);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If glView scale smaller then the borderWidth will be larger?
> 
> My suggestion is don't relay on glView scale just use follow code:
> 
> ```c++
> drawPolygon(vertices, segments, Color4B(1.0f, 0.0f, 0.0f, 1.0f), _lineWidth, color);
> ```

Yes i will change the Line...no longer need to have the same bevior as the old api.

aismann:
I have no right access :(

aismann:
@halx99 
For me its finished.

halx99 (approved):

halx99 (approved):

PR #1067, opened 2023-02-22
Scrollbar visibility fix for when auto hide is disabled

rh101:
## Describe your changes

Currently, when `ScrollViewBar::setAutoHideEnabled(false)` is called to disable the auto-hide of the scroll bar, the scrollbar opacity is set to 0 under certain conditions, when that should not be the case.  The logic in that method seems to be incorrect.

Current implementation:
```
void ScrollViewBar::setAutoHideEnabled(bool autoHideEnabled)
{
    _autoHideEnabled = autoHideEnabled;
    if (!_autoHideEnabled && !_touching && _autoHideRemainingTime <= 0)
    {
        ProtectedNode::setOpacity(_opacity);
    }
    else
        ProtectedNode::setOpacity(0);
}
```

If `_autoHideEnabled` is `false`, then it should not be checking for `_touching` or `_autoHideRemainingTime`, as they would only be relevant if `_autoHideEnabled` is set to true.

This should be the correct implementation:

```
void ScrollViewBar::setAutoHideEnabled(bool autoHideEnabled)
{
    if (_autoHideEnabled == autoHideEnabled)
        return;
    
    _autoHideEnabled = autoHideEnabled;
    if (!_autoHideEnabled)
    {
        ProtectedNode::setOpacity(255); // No point passing _opacity since it is an unknown value, so set it to 255 by default
    }
    else if (!_touching && _autoHideRemainingTime <= 0)
    {
        ProtectedNode::setOpacity(0);
    }
}
```

If the user requires a different opacity than 255, then it would simply be this call order:
```
ScrollViewBar::setAutoHideEnabled(false);
ScrollViewBar::setOpacity(requiredOpacity);
```


rh101:
Actually, would a default opacity of 128 be better instead of 255 for when auto-hide is disabled?

halx99:
> Would a default opacity of 128 be better instead of 255 for when auto-hide is disabled?

I don't think so.

rh101:
All good, then this PR is complete, and fixes the bug that has been in there since Cocos2d-x v3, unless you would want me to add a test case for it.

PR #1072, opened 2023-02-23
Tidy project tree and full  symlink support

halx99:


aismann:
Maybe "adding" one ore more  "issue" project(s) (everybody can add sources for esear testing issues)?
It is often a hard work to reproduce the issue (copy sources and so on...)

halx99:
This is a improvements PR, not only for sepcific issue

core/platform/CCFileUtils.h
```diff
@@ -43,8 +43,8 @@ THE SOFTWARE.
 #include "base/CCScheduler.h"
 #include "base/CCDirector.h"
 
-#define AX_PC_RESOURCES_DIR "Resources/"
-#define AX_PC_RESOURCES_DIR_LEN (sizeof("Resources/") - 1)
+#define AX_CONTENT_DIR "Content/"
+#define AX_CONTENT_DIR_LEN (sizeof("Content/") - 1)
```
crazyhappygame, line 47: @halx99 
Why we changed Resources to Content?
Does it mean that all games have to be changed and we should rename Resources to Content?

FYI @rh101

rh101: @crazyhappygame This change caused no issues for my projects, and was pretty quick to do.

crazyhappygame: @halx99 Thank you for answer.
Those kind of changes are not compatible with cocos2d-x and make harder to switch from cocos2d-x to axmol.
I would vote for keeping migration from cocos2d-x to axmol as simple as possible

rh101: @crazyhappygame I understand where you're coming from, but respectfully, I disagree.  The Axmol engine has moved well away from Cocos2d-x, even disregarding these latest changes, and preventing it from evolving further purely to further ease the transition from Cocos2d-x to Axmol is counterproductive, at the very least.  

Switching a project from Cocos2d-x to Axmol only has to be done once, a process that is done in a less than a day, and in the majority of cases would be a few hours at most.  The majority of changes are related to cmake and project files, and any code changes are more a matter of search and replace, so hamstringing the Axmol engine just to keep the transition simple is of no real benefit.  

Even the directory name changes in this PR took me, quite literally, a few minutes to apply for a project with over a thousand game source files and hundreds of resource, so I can't imagine it causing any issues for anyone else.

PR #1085, opened 2023-02-27
[BUG] TileMap OpenGL pixel-tolerance fix

DelinWorks:

## Describe your changes
This PR fixes a problem with TileMap rendering that causes irritating red lines to appear. i.e:

image

Notice how there are lines between the tiles in the image above? Sometimes these horizontal and vertical lines appear.

**Why does it happen?** This happens because we're performing sub-pixel rendering without using texture borders. This causes OpenGL to read/take a pixel from the texture that doesn't fall into the boundaries of the tex-coords for a given tile. In other words, when a sub-pixel is rendered on the boundary of the tile, OpenGL assumes that you'll want to retrieve a pixel on the boundary of your tex-coords. In texture atlases this can be a problem, since that one pixel can mess up your tile map -as seen above.

In this case with the vertical lines we're getting the color red because there are red pixels leaking from the atlas of the TileMap
image

`float pt = FLT_EPSILON;` shifts the tex-coords by `FLT_EPSILON` so that margin of error can be mitigated.

image

### DIFF
--- a/core/2d/CCFastTMXLayer.cpp
+++ b/core/2d/CCFastTMXLayer.cpp
@@ -594,7 +594,7 @@ void FastTMXLayer::updateTotalQuads()
                 bottom           = (tileTexture.origin.y / texSize.height);
                 top              = bottom + (tileTexture.size.height / texSize.height);
 
-                // issue# OpenGL sub-pixel horizontal-vertical lines pixel-talorance fix.
+                // issue#1085 OpenGL sub-pixel horizontal-vertical lines pixel-tolerance fix.
                 float pt = FLT_EPSILON;
 
                 quad.bl.texCoords.u = left + pt;
--- a/core/2d/CCFastTMXLayer.cpp
+++ b/core/2d/CCFastTMXLayer.cpp
@@ -594,14 +594,17 @@ void FastTMXLayer::updateTotalQuads()
                 bottom           = (tileTexture.origin.y / texSize.height);
                 top              = bottom + (tileTexture.size.height / texSize.height);
 
-                quad.bl.texCoords.u = left;
-                quad.bl.texCoords.v = bottom;
-                quad.br.texCoords.u = right;
-                quad.br.texCoords.v = bottom;
-                quad.tl.texCoords.u = left;
-                quad.tl.texCoords.v = top;
-                quad.tr.texCoords.u = right;
-                quad.tr.texCoords.v = top;
+                // issue# OpenGL sub-pixel horizontal-vertical lines pixel-talorance fix.
+                float pt = FLT_EPSILON;
+
+                quad.bl.texCoords.u = left + pt;
+                quad.bl.texCoords.v = bottom + pt;
+                quad.br.texCoords.u = right - pt;
+                quad.br.texCoords.v = bottom + pt;
+                quad.tl.texCoords.u = left + pt;
+                quad.tl.texCoords.v = top - pt;
+                quad.tr.texCoords.u = right - pt;
+                quad.tr.texCoords.v = top - pt;
 
                 quad.bl.colors = color;
                 quad.br.colors = color;
--- a/core/2d/CCFastTMXLayer.cpp
+++ b/core/2d/CCFastTMXLayer.cpp
@@ -596,7 +596,7 @@ void FastTMXLayer::updateTotalQuads()
 
                 // issue#1085 OpenGL sub-pixel horizontal-vertical lines pixel-tolerance fix.
                 float ptx = 1.0 / (_layerSize.x * tileSize.x);
-                float pty = 1.0 / (_layerSize.y * tileSize.x);
+                float pty = 1.0 / (_layerSize.y * tileSize.y);
 
                 quad.bl.texCoords.u = left + ptx;
                 quad.bl.texCoords.v = bottom + pty;
--- a/core/2d/CCFastTMXLayer.cpp
+++ b/core/2d/CCFastTMXLayer.cpp
@@ -595,8 +595,8 @@ void FastTMXLayer::updateTotalQuads()
                 top              = bottom + (tileTexture.size.height / texSize.height);
 
                 // issue#1085 OpenGL sub-pixel horizontal-vertical lines pixel-tolerance fix.
-                float ptx = 1.0 / (_layerSize.x * tileSize.x);
-                float pty = 1.0 / (_layerSize.y * tileSize.y);
+                float ptx = 1.0 / (_tileSet->_imageSize.x * tileSize.x);
+                float pty = 1.0 / (_tileSet->_imageSize.y * tileSize.y);
 
                 quad.bl.texCoords.u = left + ptx;
                 quad.bl.texCoords.v = bottom + pty;
--- a/core/2d/CCFastTMXLayer.cpp
+++ b/core/2d/CCFastTMXLayer.cpp
@@ -595,16 +595,17 @@ void FastTMXLayer::updateTotalQuads()
                 top              = bottom + (tileTexture.size.height / texSize.height);
 
                 // issue#1085 OpenGL sub-pixel horizontal-vertical lines pixel-tolerance fix.
-                float pt = FLT_EPSILON;
+                float ptx = 1.0 / (_layerSize.x * tileSize.x);
+                float pty = 1.0 / (_layerSize.y * tileSize.x);
 
-                quad.bl.texCoords.u = left + pt;
-                quad.bl.texCoords.v = bottom + pt;
-                quad.br.texCoords.u = right - pt;
-                quad.br.texCoords.v = bottom + pt;
-                quad.tl.texCoords.u = left + pt;
-                quad.tl.texCoords.v = top - pt;
-                quad.tr.texCoords.u = right - pt;
-                quad.tr.texCoords.v = top - pt;
+                quad.bl.texCoords.u = left + ptx;
+                quad.bl.texCoords.v = bottom + pty;
+                quad.br.texCoords.u = right - ptx;
+                quad.br.texCoords.v = bottom + pty;
+                quad.tl.texCoords.u = left + ptx;
+                quad.tl.texCoords.v = top - pty;
+                quad.tr.texCoords.u = right - ptx;
+                quad.tr.texCoords.v = top - pty;
 
                 quad.bl.colors = color;
                 quad.br.colors = color;


rh101:
> OpenGL assumes that you'll want to retrieve a pixel on the boundary of your tex-coords. In texture atlases this can be a problem, since that one pixel can mess up your tile map -as seen above.

Can't you simply eliminate this problem by ensuring the individual textures in an atlas have at least 2 pixels padding between them? 

If you use Texture Packer, then it's the "Shape Padding" setting, which if I recall correctly, is set to 2 by default to fix this issue.  The Texture Packer CLI parameter is [--shape-padding](https://www.codeandweb.com/texturepacker/documentation/texture-settings#shape-padding---shape-padding-number), and as you can see at that link, it explicitly states this: "Use a value of at least 2 to avoid dragging in pixels from neighbor sprites when using OpenGL rendering."

With the change you've made in the code, are you absolutely certain there are no side-effects or inaccuracies with the display of the textures?  If you're not sure, then the better way to fix this is to just use the conventional method to prevent this issue, by just making sure you have the required padding when creating your texture atlas.

rh101:
> I have checked and there were no inaccuracies, the `FLT_EPSILON` is such a small value `(0.0000001192092896)` and that wouldn't have any negative side effects besides fixing floating point errors. Also TileMaps do not use texture packer atlas textures, they're created using Tiled where every 32x32 pixels in a texture is a tile.
> 
> > This wouldn't just be an issue with tile maps either, so I'm surprised you haven't experienced this issue previously with non-tile map related sprites.
> 
> I have been through this and I knew about the solution you provided, but that doesn't apply to TileMaps.

In Tiled, the setting is called "spacing", and it's supported in the standard output formats.  I've never worked with Tiled before, but it definitely has that setting somewhere.

Search for "spacing" in these pages under the "tileset" sections:
https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tileset
https://doc.mapeditor.org/en/stable/reference/json-map-format/#tileset

DelinWorks:
Unfortunately, spacing makes working with tiles harder because you can't put tiles next to each other, first tile is at 0 then next one is at 33 then next one is at 66 because of the spacing, don't get me wrong it is possible with some software to create TileMaps with spacing but this solution is way better for workflow and for texture compression so you can fit more tiles in a single some specific resolution, my solution implies in this PR that you can make tiles for each 32 multiplier (i.e. first 0, second 32, third 64) without having these sub-pixel lines.

rh101:
> Unfortunately, spacing makes working with tiles harder because you can't put tiles next to each other, first tile is at 0 then next one is at 33 then next one is at 66 because of the spacing, don't get me wrong it is possible with some software to create TileMaps with spacing but this solution is way better for workflow and for texture compression so you can fit more tiles in a single some specific resolution, my solution implies in this PR that you can make tiles for each 32 multiplier (i.e. first 0, second 32, third 64) without having these sub-pixel lines.

I'm not sure what you mean by "put tiles next to each other".  Is it a manual process to place the tiles in Tiled?

Other than that, the TMX parser in Axmol already supports the "spacing" and "margin" fields, so when it creates the list of tile textures it should have the correct coordinates to each one in the altas, without having to care about the padding/margin.

DelinWorks:
I was talking about assembling tiles in a texture not when you make the map.

rh101:
> I was talking about assembling tiles in a texture not when you make the map.

This is the part that is confusing me.  I don't understand what you mean by assembling the tiles in the texture.  How exactly do you do that? What application do you use for this?

DelinWorks:
When you said spacing and merging you were talking about textures right?

because having this texture:
image

with 0 spacing it's easier to maintain and edit the texture than having to space every tile in that texture.

rh101:
> When you said spacing and merging you were talking about textures right?

I'm trying to understand why it's not possible to use the built-in support in Tiled and texture packers to completely avoid the issue of color bleeding.  Tiled supports it in the tile map output for all formats.  What I don't understand is your statement about it being more difficult to maintain or edit.  What do you mean by that? How are you constructing that texture atlas? 

> with 0 spacing it's easier to maintain and edit the texture than having to space every tile in that texture.

Again, I don't understand that statement, because why would you even care about the spacing given that both texture packers and the TileMap support that setting, so you would never have to deal with it, as it is entirely transparent to you.

DelinWorks:
This is the difference between this solution and without it:

This is where you DON'T shift the tex-coords by `FLT_EPSILON` (Before):
image

and this is where you do (After):
image

This is the tile, I've put an outline around it to see the difference:
image

The one-pixel blue outline is getting rendered on both, just to show that there are no problems.

rh101:
> The one-pixel blue outline is getting rendered on both, just to show that there are no problems.

Sorry, I'm not saying your PR is an issue, I'm just trying to understand what exactly is going on regarding Tiled and the way you're creating the texture atlas.  I would appreciate it if you answered the questions that I put forward in the previous posts.

DelinWorks:
The solutions you mentioned are workarounds, it would be cool if you are new to the engine, and you create a tilemap and that tilemap renders fine and you don't have to go around the trouble of spacing the tiles.

rh101:
It's not about this or other engines, or any opinions on whether the padding is a work-around or not.  

I'm curious to understand why the padding and margin didn't fix the issue, because as you stated earlier, it doesn't apply to TileMaps, but I need to understand why you made that statement, because from all the information I'm reading, it should have worked.

I still would like to know how you are creating your texture atlas, such as the one you posted above?
https://user-images.githubusercontent.com/45469625/221601440-23360fa1-4d9d-4278-bac9-92aa500031b5.png

DelinWorks:
Let's say you've put spacing around the tiles to avoid this problem, It won't be fixed because the spacing you've put there has transparent pixels. in case of texture packer, sure the red blue and green colors are gone but there is still a transparent line between the tiles that goes to the camera's background and leave black lines. unless you put extra pixels around the tile so that if the colors around the padding leak, it won't be notices. See how that's too much work?

rh101:
> Let's say you've put spacing around the tiles to avoid this problem, It won't be fixed because the spacing you've put there has transparent pixels. in case of texture packer, sure the red blue and green colors are gone but there is still a transparent line between the tiles that goes to the camera's background and leave black lines. unless you put extra pixels around the tile so that if the colors around the padding leak, it won't be noticed. See how that's too much work?

Once again, I understand all of that, but the one question I would appreciate an answer for is this: 

How are you creating your texture atlas?

You made statements that I'm trying to fully understand, one being about the padding/margin not applying to TileMaps, and the other being that it's harder to work with.  Those statements do not make sense to me, and I'm trying to understand whether they truly don't work, or if there is something different about the process you're using to create the atlas and tile maps that is causing the padding method not to work.

DelinWorks:
> How are you creating your texture atlas?
I use gimp to edit my texture atlas, I first divide it to lets say 32 and then draw each cell

> You made statements that I'm trying to fully understand, one being about the padding/margin not applying to TileMaps, and the other being that it's harder to work with with the tiles and texture atlas. Those statements do not make sense to me, and I'm trying to understand whether they truly don't work, or if there is something different about the process you're using to create the atlas and tile maps that is causing the padding method not to work.

Padding/margin doesn't apply to tilemaps because like I said above
> Let's say you've put spacing around the tiles to avoid this problem, It won't be fixed because the spacing you've put there has transparent pixels. in case of texture packer, sure the red blue and green colors are gone but there is still a transparent line between the tiles that goes to the camera's background and leave black lines. unless you put extra pixels around the tile so that if the colors around the padding leak, it won't be noticed. See how that's too much work?

I also said that it's harder to work with because with gimp I only got used to tilemaps having no spacing. and adding spacing seems hard and counter-intuitive.

rh101:
> I use gimp to edit my texture atlas, I first divide it to lets say 32 and then draw each cell

Thank you for the answer.  Now it makes sense why you stated that it's harder to work with, since you're creating it manually in a graphic editor.  That's a really tough way to create it.

> Padding/margin doesn't apply to tilemaps because like I said above

Assuming you're referring to the remark regarding gimp, then that is only the case because you're working with it manually. Technically though, the method does work, and does apply to tile maps, since there is support in the tile map format to cater for this specific issue.

> I also said that it's harder to work with because with gimp I only got used to tilemaps having no spacing. and adding spacing seems hard and counter-intuitive when the better solution is this PR.

Now it all makes sense. Context is key, and for anyone reading this, we don't want them to think that there is a technical issue with the padding/margin method, because it should work.  The statements you made were all to do with your specific process of creating the tile maps etc, and that just needed to be clear.

DelinWorks:
> Assuming you're referring to the remark regarding gimp, then that is only the case because you're working with it manually. Technically though, the method does work, and does apply to tile maps, since there is support in the tile map format to cater for this specific issue.

one minor point though is that I was referring to texture packer, this program does add spacing. it's just that it leaves transparent area and doesn't fold the tile itself. that's what's causing the issue above of spacing/padding potentially not working.

rh101:
> it's just that it leaves transparent area and doesn't fold the tile itself.

What do you mean by "fold the tile"?

> that's what's causing the issue above of spacing/padding potentially not working for TileMaps.

That should not be the case.  Have you checked the cpp_tests project to verify this? I'll check the test myself to see what happens, but there should be no reason for it to cause issues in TileMaps, as long as the tile map file (TMX) has the same padding and margin values set to match what is in the sprite atlas.

rh101:
This issue does not exist when padding and margin are used, so it's definitely working. 

For this texture atlas ("tmw_desert_spacing.png"):
tmw_desert_spacing

This TMX file, "orthogonal-test3.tmx" (note "spacing" and "margin" set to match atlas):
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE map SYSTEM "http://mapeditor.org/dtd/1.0/map.dtd">
<map version="1.0" orientation="orthogonal" width="32" height="32" tilewidth="32" tileheight="32">
 <tileset name="Desert" firstgid="1" tilewidth="32" tileheight="32" spacing="1" margin="1">
  <image source="tmw_desert_spacing.png"/>
 </tileset>
 <!-- Layer data is compressed (GZip) binary data, encoded in Base64 -->
 <layer name="Layer 0" width="32" height="32">
  <data encoding="base64" compression="gzip">
   H4sIAAAAAAAAAO3TTQ+CMAwG4KIn8aSelIRE4IIf/P+fZ5dsoTQlqXPlYHp4krFA361sAwA8lJ6oLSzUnAzq5qrQTrA3ymvY8wHVgqNRfp+Zf0W3uP40pvi+SuffURe/T2OK19Xmn9BZcGHvjTDfiTTOuSvadXK0/5r5Nbl3mvZfM1/aWg79L1bZocfvmP+C5Zn/tv+/7F0681v0P2QMIJ/5Lfo/Gtd3zjnnnPtnH/8uAvwAEAAA
  </data>
 </layer>
</map>
```

This is the output, and note no black borders as you would expect if the test was to fail because of the issue you described:
image

DelinWorks:
Now, try to resize the window or use a resolution that's not divisible by 2, you'll see black lines leaking because of the black border spacing around tiles.

rh101:
> Now, try to resize the window or use a resolution that's not divisible by 2, you'll see black lines leaking because of the black border spacing around tiles.

That's a good catch, and can definitely reproduce the issue after resizing the window a few times. I tried a few different resolutions that are odd numbers, but none showed the issue, so it must be something more specific.

DelinWorks:
> What do you mean by "fold the tile"?

Tile Folding is the process of extruding each tile's pixels in one pixel in all directions so that if the error open sub-pixel error was to happen, it will just copy the same texture and no color leaking would be noticeable even though colors did leak.

Many texture packers (not including TexturePacker itself because it doesn't do that me thinks) like free tex packer do this extrusion method which I think I'm calling it wrong idk, but I hope you get the idea.

Like this:
image

DelinWorks:
> That's a good catch, and can definitely reproduce the issue after resizing the window a few times. I tried a few different resolutions that are odd numbers, but none showed the issue, so it must be something more specific.

hmm, that's weird.. I'll see what I can do.

DelinWorks:
I tested the tileset in tiled with a spacing and margin of 0 and this is the result:

image

There are lines as you can see here with a resolution of (1292x726).

image

DelinWorks:
It seems that a constant decimal was not the solution, but rather it should depend on the map size and tile size. I've tested it and all sizes of tile (8, 16, 32, 64) render correctly without any lines.

DelinWorks:
@halx99 done! any thoughts on this?

Issue #1087, opened 2023-03-01
[iOS] App crashes on start-up due to resource path changes


rh101:
- axmol version: latest
- devices test on: iOS simulator and device
- developing environments
   - NDK version: r19c
   - Xcode version: 12.4
   - Visual Studio: 
      - VS version: 2019 (16.11), 2022 (17.4)
      - MSVC version: 1929, 1934
      - Windows SDK version: 10.0.22621.0
   - cmake version: 
Steps to Reproduce:

1. create a new project from the Axmol cpp template 
2. build and run the project

It will crash with the following output:

```
xmol: fullPathForFilename: No file found at CloseNormal.png. Possible missing file.
axmol: fullPathForFilename: No file found at CloseSelected.png. Possible missing file.
Error while loading: 'CloseNormal.png' and 'CloseSelected.png'
Depending on how you compiled you might have to add 'Resources/' in front of filenames in HelloWorldScene.cpp
axmol: fullPathForFilename: No file found at fonts/Marker Felt.ttf. Possible missing file.
Error while loading: 'fonts/Marker Felt.ttf'
Depending on how you compiled you might have to add 'Resources/' in front of filenames in HelloWorldScene.cpp
axmol: fullPathForFilename: No file found at HelloWorld.png. Possible missing file.
Error while loading: 'HelloWorld.png'
Depending on how you compiled you might have to add 'Resources/' in front of filenames in HelloWorldScene.cpp
2023-03-01 17:51:38.692532+1100 HelloCpp[4211:37333] fopen failed for data file: errno = 2 (No such file or directory)
2023-03-01 17:51:38.692783+1100 HelloCpp[4211:37333] Errors found! Invalidating cache...
axmol: QuadCommand: resizing index size from [-1] to [2560]
2023-03-01 17:51:39.171361+1100 HelloCpp[4211:37333] fopen failed for data file: errno = 2 (No such file or directory)
2023-03-01 17:51:39.171471+1100 HelloCpp[4211:37333] Errors found! Invalidating cache...
```

rh101:
Does iOS look into a "Resources" folder by default? I can't find any references in the cmake files or in the source code to "Resources", so it may be something in either XCode or iOS.

rh101:
Original issue fixed in #1092 for all tests projects, but a copy and paste caused a problem with the cpp-template-default, which is now fixed in #1095 

Will close this once that one is done.

PR #1092, opened 2023-03-01
[Apple] Fix for missing resources

rh101:
## Describe your changes
~~It seems that iOS and MacOS require resources to be mapped to the "Resources" folder in the Xcode project, otherwise the content will not be found at runtime.~~

The actual issue is that the resources need to be marked as such in CMake so that they are set to the correct destination in the Xcode project (Build Phases->Copy Files, with "Destination" set to "Resources").

The specific CMake property that needs to be set correctly is `MACOSX_PACKAGE_LOCATION`. More details in the following posts below.

Tested cpp and lua project builds for iOS sim, iOS device, and Mac OS M1, and they're now working correctly.

#1087 


halx99:
Does there has any apple docoment indicate the resouce folder name must be `Resources`?

rh101:
> Does there has any apple docoment indicate the resouce folder name must be `Resources`?

https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW14
https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/20001119-110730
https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW19

and this regarding CMake:
https://cmake.org/cmake/help/latest/prop_tgt/RESOURCE.html

For Mac OS, this is the bundle structure:
image

For iOS, it's a flat structure:
image

The change made in PR #1072 to have them all in "Contents" is fine, but when mapping for Xcode, it seems to require them in a "Resources" folder within the project tree, and on build, it places them in the correct path for the app bundles.  It may be just an Xcode requirement to have them in a "Resources" folder within the project, but I honestly don't know.

rh101:
In Xcode, the build phase copies everything into "Resources" for Mac OS apps, so maybe there is something missing from the cmake configuration that isn't properly setting this up:

image
image

Same for iOS project:

image

rh101:
The only change in this PR is to affect these cmake helpers:

`ax_mark_multi_resources` and `ax_mark_resources`

From my understanding of how it works, it's not actually copying any files, it's just marking them as resources, but is there something that needs to be changed in here (from latest axmol code):
```
# mark `FILES` as resources, files will be put into sub-dir tree depend on its absolute path
function(ax_mark_resources)
    set(oneValueArgs BASEDIR RESOURCEBASE)
    set(multiValueArgs FILES)
    cmake_parse_arguments(opt "" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})

    if(NOT opt_RESOURCEBASE)
        set(opt_RESOURCEBASE Resources)
    endif()

    get_filename_component(BASEDIR_ABS ${opt_BASEDIR} ABSOLUTE)
    foreach(RES_FILE ${opt_FILES} ${opt_UNPARSED_ARGUMENTS})
        get_filename_component(RES_FILE_ABS ${RES_FILE} ABSOLUTE)
        file(RELATIVE_PATH RES ${BASEDIR_ABS} ${RES_FILE_ABS})
        get_filename_component(RES_LOC ${RES} PATH)
        set_source_files_properties(${RES_FILE} PROPERTIES
                                    MACOSX_PACKAGE_LOCATION "${opt_RESOURCEBASE}/${RES_LOC}"
                                    HEADER_FILE_ONLY 1
                                    )

        if(XCODE OR VS)
            string(REPLACE "/" "\\" ide_source_group "${opt_RESOURCEBASE}/${RES_LOC}")
            source_group("${ide_source_group}" FILES ${RES_FILE})
        endif()
    endforeach()
endfunction()
```

rh101:
Just found the problem! 

If "Content" is passed to this:
`ax_mark_multi_resources(common_content_files RES_TO "Content" FOLDERS ${content_folder})`

Look at what happens here in the Xcode project:
image

Notice the issue? It's setting the `Destination` field to `Executables`, rather than `Resources`.
image

rh101:
In `ax_mark_resources`, I just did a quick test by changing:

        set_source_files_properties(${RES_FILE} PROPERTIES
                                    MACOSX_PACKAGE_LOCATION "${opt_RESOURCEBASE}/${RES_LOC}"
                                    HEADER_FILE_ONLY 1
                                    )

to this

        set_source_files_properties(${RES_FILE} PROPERTIES
                                    MACOSX_PACKAGE_LOCATION "Resources/${RES_LOC}"
                                    HEADER_FILE_ONLY 1
                                    )

and even if I "Content" is passed to this in the game `CMakeLists.txt`:
`ax_mark_multi_resources(common_content_files RES_TO "Content" FOLDERS ${content_folder})`

The app worked on iOS.

rh101:
As far as I can tell from all of the above, the changes in this PR are correct.  The other option would be to hardcode "Resources" in the cmake `ax_mark_resources`, changing this:

`MACOSX_PACKAGE_LOCATION "${opt_RESOURCEBASE}/${RES_LOC}"`

to this:
`MACOSX_PACKAGE_LOCATION "Resources/${RES_LOC}"`

Assuming that it is only ever used for resources, and the files **never** require a different "Destination" (`RESOURCEBASE`) to ever be set.

rh101:
I now have a bit more of an understanding regarding how this all works.  The `MACOSX_PACKAGE_LOCATION` property for files that we want as resources do need to be prefixed with "Resources" in their path in order to work for Apple app bundles.

So, technically, setting the `MACOSX_PACKAGE_LOCATION` to this would be fine:
```
set_source_files_properties(${RES_FILE} PROPERTIES
                                    MACOSX_PACKAGE_LOCATION "Resources/${RES_LOC}"
                                    HEADER_FILE_ONLY 1
                                    )
```

The next issue is to do what the display within the IDE via `source_group`. This section is for both XCODE and VS:

        if(XCODE OR VS)
            string(REPLACE "/" "\\" ide_source_group "${opt_RESOURCEBASE}/${RES_LOC}")
            source_group("${ide_source_group}" FILES ${RES_FILE})
        endif()

If we set `MACOSX_PACKAGE_LOCATION "Resources/${RES_LOC}"`, then we can still pass in "Content" to this in the game `CMakeLists.txt`:
`ax_mark_multi_resources(common_content_files RES_TO "Content" FOLDERS ${content_folder})`

and in the IDE, it'll show up in the "Content" folder, rather than "Resources".

On the other hand, if we leave it as:
`MACOSX_PACKAGE_LOCATION "${opt_RESOURCEBASE}/${RES_LOC}"`

and pass in "Resources" in:
`ax_mark_multi_resources(common_content_files RES_TO "Resources" FOLDERS ${content_folder})`

Then both XCODE and VS IDEs will show them in a "Resources" folder within the project tree (that is what happens with the changes in this PR).  As you know, it's just a visual thing, and does not affect the actual folder structure or build.

mariusz102102:
@rh101 thank you for instigation.

@rh101 @halx99 
Is it possible to create automated CI test to detect that engine is in healthy state?
Maybe run cpp-test in some special mode?

As far I remember windows gihub CI hosts allows to run GUI app. Linux not possible. Not sure about MacOS

PR #1098, opened 2023-03-06
[BUG] TileMap flipped/rotated animated tiles incorrect rendering.

DelinWorks:
## Describe your changes
In Tiled map editor, If you animate tiles they look completely fine and render okay in axmol, but when rotated, the flip and rotate flags of each tile are taken into account as the animation index, which is wrong, they should be stored in an animation map and then when it comes time to set the frame, the frame is set with it's corresponding flags.
 
Tiled renders the flipped/rotated animated tiles correctly:
image

but axmol shows the tile itself without its attached animation frames, which is wrong:
image

Go to test commit 2756d96 to see what I mean.

This PR fixes this issue by removing the flipped/rotated flags from gid and stores them in a separate `uint32_t` and then passes this value for each tile animation task. Then the task would set the frame gid of a tile in a layer with the flags its been given before hand: 
image


core/2d/CCFastTMXLayer.cpp
```diff
@@ -356,7 +356,16 @@ void FastTMXLayer::setupTiles()
...
 
                 int pos = static_cast<int>(newX + _layerSize.width * y);
-                int gid = _tiles[pos];
+                uint32_t gid = _tiles[pos];
+                uint32_t flags = 0;
+
+                flags |= gid & kTMXTileHorizontalFlag;
+                flags |= gid & kTMXTileVerticalFlag;
+                flags |= gid & kTMXTileDiagonalFlag;
+
```
rh101, line 365: It may be better to just do this:
`flags |= gid & kTMXFlipedAll;`

since `kTMXFlipedAll          = (kTMXTileHorizontalFlag | kTMXTileVerticalFlag | kTMXTileDiagonalFlag)`

core/2d/CCFastTMXLayer.cpp
```diff
@@ -356,7 +356,16 @@ void FastTMXLayer::setupTiles()
...
+                gid &= ~kTMXTileHorizontalFlag;
+                gid &= ~kTMXTileVerticalFlag;
+                gid &= ~kTMXTileDiagonalFlag;
```
rh101, line 368: This can also be replaced with:
`gid &= ~kTMXFlipedAll;`

core/2d/CCFastTMXLayer.cpp
DelinWorks: Good catch!

core/2d/CCFastTMXLayer.cpp
DelinWorks: > or
> gid &= kTMXFlippedMask;

wait.. don't you need to AND something with the complement to clear bits?

rh101: They are defined as follows:

```
typedef enum TMXTileFlags_
{
    kTMXTileHorizontalFlag = 0x80000000,
    kTMXTileVerticalFlag   = 0x40000000,
    kTMXTileDiagonalFlag   = 0x20000000,
    kTMXFlipedAll          = (kTMXTileHorizontalFlag | kTMXTileVerticalFlag | kTMXTileDiagonalFlag),
    kTMXFlippedMask        = ~(kTMXFlipedAll)
} TMXTileFlags;
```

DelinWorks: Oh sorry I didn't notice the `kTMXFlippedMask` I thought it was `kTMXFlippedAll` that's what confused me

rh101: Either way is fine, just use whatever is makes more sense within the context of the code you're writing (ie. whatever is more readable/requires less effort to figure out what is going on for anyone reading the code).

PR #1113, opened 2023-03-12
fixing some warning messages of code not conform C++20 yet

Joilnen:

## Describe your changes
Only simple C++20 compilation warning fixing


halx99:
UWP require c++17, so please ensure your patch can be compiled in c++17, not only c++20

halx99 (changes requested):
UWP require c++17, so please ensure your patch can be compiled in c++17, not only c++20

rh101:
> Only simple C++20 compilation warning fixing

@Joilnen You can fix the C++17 issue by changing `[=, this]` to `[this]`, since none of them require copying by value, and even if they did, then explicitly capture each variable rather than using `=`.

PR #1122, opened 2023-03-15
Add ClippingNode Performance Test

Yehsam23:
THIS PROJECT IS IN DEVELOPMENT MODE. Any pull requests should merge to branch `dev`, otherwise will be rejected immediately.

## Describe your changes
Add Clipping Node Performance Test to "cpp-test"

## Issue ticket number and link
#1121 

## Checklist before requesting a review
-  [x] I have performed a self-review of my code.
-  [ ] If it is a core feature, I have added thorough tests.
-  [ ] I have checked readme and add important infos to this PR (if it needed).
-  [x] I have added/adapted some tests too.

Issue #1121, opened 2023-03-15
Performance issue vs Cocos2d-x v3.17


Yehsam23:
Download Test Project:
[TestClip.zip](https://github.com/axmolengine/axmol/files/10976641/TestClip.zip)

The following tests were conducted using an Android HTC U11 Plus with the build release APK.

When using 200 ClippingNode and Sprite, there is a significant difference in performance between Axmol and Cocos2d-x v3.17. Axmol's FPS is approximately 9-10, while Cocos2d-x v3.17's FPS is approximately 37-40.

Axmol:
![ax_clip](https://user-images.githubusercontent.com/1954445/225224980-638d9d6e-b144-4410-8d4f-02f94ed31b51.png)

Cocos2d-x v3.17:
![v3_clip](https://user-images.githubusercontent.com/1954445/225225019-c1f2aed5-bdd2-4df4-8c5f-0ae936320fba.png)

When using 4000 Sprites with SpriteA->SpriteB->SpriteA->SpriteB repeating pattern, Axmol's FPS is approximately 28-30, while Cocos2d-x v3.17's FPS is approximately 38-39.

Axmol:
![ax_sprite](https://user-images.githubusercontent.com/1954445/225225266-6feb3152-cb81-4253-8ba9-11250fa8eee4.png)

Cocos2d-x v3.17:
![v3_sprite](https://user-images.githubusercontent.com/1954445/225225083-9e3f5fc2-332f-479b-a3cb-881ec9241ea2.png)

As can be seen from the above, there is a significant performance difference when using ClippingNode.
According to experiment #1094, it is indeed a problem with Cocos2d-x v4.

Cocos2d-x v4 also had a similar issue initially. [issue](https://github.com/cocos2d/cocos2d-x/issues/20668)

I know that Axmol is a fork from Cocos2d-x-4.0. While it may be unfair to compare its performance to Cocos2d-x v3.17, I also hope to find ways to improve its performance.

aismann:
Please make an "official cpp-test" on clipping section too.
-> It can help to detect "issues" with changes in the future too

Yehsam23:
@aismann 
Is to submit a Pull Request directly in "cpp-tests" to add a clipping node performance test?

aismann:
> @aismann Is to submit a Pull Request directly in "cpp-tests" to add a clipping node performance test?

Yes. thas what I mean (Your test scenario will be part of regular tests then)

aismann:
This is on my machine:
[...]

PR #1125, opened 2023-03-20
Improve Android performance

Yehsam23:

## Describe your changes
Decreasing the size of the initialization buffer can prevent FPS drops on Android

#1121

-  [x] I have added/adapted some tests too. (#1122)

### DIFF
--- a/core/renderer/CCRenderer.cpp
+++ b/core/renderer/CCRenderer.cpp
@@ -995,6 +995,10 @@ void Renderer::TriangleCommandBufferManager::createBuffer()
 {
     auto device = backend::Device::getInstance();
 
+    // Not initializing the buffer before passing it to updateData for Android/OpenGL ES.
+    // This change does fix the Android/OpenGL ES performance problem
+    // If for some reason we get reports of performance issues on OpenGL implementations,
+    // then we can just add pre-processor checks for OpenGL and have the updateData() allocate the full size after buffer creation.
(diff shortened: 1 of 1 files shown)


halx99:
maybe the GL buffer don't require init with zero, so the updateData with zero buffer can be removed

rh101:
@Yehsam23 You may not need to pass any data to the updateData method to initialize the buffer, but you can still pass the size, like this:

`vertexBuffer->updateData(nullptr, size);`

since the `const void* data` parameter of the `glBufferData` call can be a null:

`void glBufferData(GLenum target, GLsizei size, const void* data, GLenum usage)`

So there is no need for the memory allocation etc.

Yehsam23:
Can we use the same code for Metal directly, since Metal also does not use updateData?

rh101:
> Can we use the same code for Metal directly, since Metal also does not use updateData?

Yes

rh101:
I do think this is a OpenGL ES v2 implementation issue though, since this issue is not present in OpenGL.  By not calling `updateData` with the full size, the buffer is not pre-allocated, so if the size changes on every call to future `updateData`, then it would mean the buffer would have to be re-allocated with the correct size, which may have an impact on performance.

aismann:
> Regardless, this change does fix the Android/OpenGL ES performance problem, so we should use it. If for some reason we get reports of performance issues on OpenGL implementations, then we can just add pre-processor checks for OpenGL and have the `updateData()` allocate the full size after buffer creation.

@rh101 
Please add a comment for this "reason" on the source

PR #1131, opened 2023-03-24
[BUG] Fix TMX Tiled Map `CSV`, `BASE64`, & `ZLIB` incorrect parsing

DelinWorks:
## Describe your changes
Fix TileMap TMX tiles from appearing janky when the TMX file is saved as `CSV` or `BASE64 ZLIB` or `BASE64 GZIP`, this happens on some latest versions of Tiled editor (possibly 1.9.2 and up) the problem is cause by tiled adding "\n\r" in `CSV` and `BASE64 ZLIB/GZIP` modes in the TMX save file, so these newlines are removed, also base64 has to be trimmed from all spaces to work.

### DIFF
--- a/core/2d/CCTMXXMLParser.cpp
+++ b/core/2d/CCTMXXMLParser.cpp
@@ -688,11 +688,7 @@ void TMXMapInfo::endElement(void* /*ctx*/, const char* name)
             tmxMapInfo->setStoringCharacters(false);
 
             TMXLayerInfo* layer = tmxMapInfo->getLayers().back();
-
-            auto currentStringView = tmxMapInfo->getCurrentString();
-            std::string currentString;
-            std::regex_replace(std::back_inserter(currentString), currentStringView.begin(), currentStringView.end(),
-                               std::regex("[\n\r ]"), "");
+            auto currentString = tmxMapInfo->getCurrentString();
 
             unsigned char* buffer;
             auto len = utils::base64Decode((unsigned char*)currentString.data(), (unsigned int)currentString.length(),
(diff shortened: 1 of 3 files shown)


rh101:
@DelinWorks Regarding the changes, wouldn't something like this be simpler:

`std::regex_replace(currentString, std::regex("[\n\r ]"), "");`

With that you wouldn't need to add any new methods to ccUtils.

rh101:
This may be more efficient, and will operate directly on the `string_view` returned by `getCurrentString()` without all the copying:

```
auto currentStringView = tmxMapInfo->getCurrentString();
std::string currentString;
std::regex_replace(std::back_inserter(currentString), currentStringView.begin(), currentStringView.end(), std::regex("[\n\r ]"), "");
```

DelinWorks:
It is definitely faster!

rh101:
> It is definitely faster!

I initially didn't notice that `getCurrentString()` was returning `std::string_view`, and found it a little strange that the result of that method was explicitly being copied to a new `std::string`.  But yes, avoiding string copies where possible is a good thing.

DelinWorks:
I think that the function `getCurrentString()` itself should return a string that's trimmed and doesn't contain newlines, whatevers storing that string in that format should be looked into, I might want to look deeper into this.

rh101:
> I think that the function `getCurrentString()` itself should return a string that's trimmed and doesn't contain newlines, whatevers storing that string in that format should be looked into, I might want to look deeper into this.

If the raw string isn't required in its original form (with `\r`, `\n` and `" "`), then yes, stripping those characters out while it is initially being read and processed is the most efficient thing to do.

DelinWorks:
That's the best way to solve it! any text that comes to the `textHandler` delegate should be handled appropriately!

halx99 (approved):

PR #617, opened 2022-04-17
Multiple textured layers

balancedwolf:
This enables maps to have more than 1 texture per layer

PR #1135, opened 2023-03-24
[IMPROVMENT] Add support for TMX multiple tilesets layers #617

DelinWorks:
Refer to #617 


rh101:
@DelinWorks Is there any specific reason you created a PR before it's ready?  

Please avoid doing this, and there are several reasons as to why:

1- For anyone who has set a 'watch' on this project to be notified of any changes, it constantly sends emails out for the PR itself, and every change to that PR.
2- It triggers the CI builds for no reason at all when there is no need for it, which is wasteful on resources.

The process should be that you create a branch on your own fork, make changes to that branch, and once you have tested it on your own system and it's ready for review, then you create PR for it on the primary repository.  That way there are no constant notifications, and no wasting of CI resources.

DelinWorks:
This problem is actually harder than I anticipated, I'll try something else

PR #1136, opened 2023-03-25
[IMPROVMENT] Add support for TMX multi-tileset layers

DelinWorks:
Refer to #617

I'm not using the same approach that @balancedwolf used (they used custom shaders and texture indexing), mainly because it's very complicated and uses custom shaders and that approach will not work for all platforms, it will only target DirectX11 and OpenGL, For Metal and older hardware it's another story (metal doesn't support texture arrays), my approach is to use an array that stores each active tileset in a layer and that layer will have many sub-layers, each sub-layer has it's own tileset, draw call, program state, and texture. this approach is simpler and works for all platforms (not yet tested for all platforms but judging it will work).

### DIFF
--- a/core/2d/CCActionManager.cpp
+++ b/core/2d/CCActionManager.cpp
@@ -183,7 +183,7 @@ void ActionManager::addAction(Action* action, Node* target, bool paused)
 
     actionAllocWithHashElement(element);
 
-    //AXASSERT(!ccArrayContainsObject(element->actions, action), "action already be added!");
+    AXASSERT(!ccArrayContainsObject(element->actions, action), "action already be added!");
     ccArrayAppendObject(element->actions, action);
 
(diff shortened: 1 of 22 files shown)


aismann:
@DelinWorks 
An example on cpp_test will be great.

rh101:
Did you accidentally merge an incorrect branch into your changes for this PR in [39d7ffd](https://github.com/axmolengine/axmol/pull/1136/commits/39d7ffde6c90544fd6a6963221b1c16e223e3c98)? There are a lot of changes that are unrelated to the TMX modifications, and they should not be part of it.

DelinWorks:
Noway i did that! sorry my bad.

DelinWorks:
Accidentally nuked this branch my bad 😂

PR #1142, opened 2023-03-27
[IMPROVMENT] Add support for TMX multi-tileset layers 

DelinWorks:
Refer to #1136 

I'll make tests now.


DelinWorks:
@aismann Something happened that made me think this implementation has a bug, this bug made so tiles look wrong and out of order, after hours of trying to find that bug turns out it was non-existent, but rather it was Tiled, when you save tmx files in Tiled it doesn't update the `tsx` files automatically based on the new image, so you need to save them manually anytime you change your texture size, I'm saying this to save you hours of frustration.

aismann:
> @aismann Something happened that made me think this implementation has a bug, this "bug" made tiles look wrong and out of order, after hours of trying to find that bug turns out it was non-existent, but rather it was Tiled, when you save tmx files in Tiled it doesn't update the `tsx` files automatically based on the new image, so you need to save them manually anytime you change your texture size, I'm saying this to save you hours of frustration.

Can you add  a comment plz (on APPENDIX.md ?)

DelinWorks:
@aismann Unfortunately I couldn't fix this bug, spent hours to no avail, I think at that point I should just stop.

expected this:

image

and got:

image

I shuffled the code and rinsed over and over but I couldn't get that bug to be fixed for the life of me.. I think that entire class needs a rewrite or someone with better experience can do it

aismann:
> @aismann Unfortunately I couldn't fix this bug, spent hours to no avail, I think at that point I should just stop.
> 
> expected this:
> 
> image
> 
> and got:
> 
> image
> 
> I shuffled the code and rinsed over and over but I couldn't get that bug to be fixed for the life of me.. I think that entire class needs a rewrite or someone with better experience can do it

Is this the "new example for multi tilesets?
Please add  an issue and let add "pinned".
Add the issue number on the tester subtilte too.
Thanks

aismann:
@halx99
What does you mean?
New branch for this for testing/fixing?

PR #1186, opened 2023-04-25
[IMPROVMENT] Remove unnecessary float pass reference in HSV/HSL

DelinWorks:
## Describe your changes
It's not optimal to pass floats by reference since they're 32bits and doing so copies 64bits for the pointer instead of 32bits on 64bit systems.

### DIFF
--- a/core/base/ccTypes.cpp
+++ b/core/base/ccTypes.cpp
@@ -438,7 +438,7 @@ void HSV::set(float r, float g, float b, float a)
     this->a = a;
 }
 
-void HSV::get(float& r, float& g, float& b) const
+void HSV::get(float r, float g, float b) const
 {
     float hue = remainder(std::fabs(h), 360);
     hue += 360;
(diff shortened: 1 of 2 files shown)


halx99:
Why remove reference from output variable?

halx99:
The hsv get, set should be rename to fromRgba and toRgba

DelinWorks:
Oh I remember... it's to edit floats right away rather than getting returning things! this PR is wrong

halx99:
And some function of the class HSL can be improve, i.g. 

```cpp
    Color3B toColor3B();
    Color4B toColor4B();
    Color4F toColor4F();
```
should changed to:
```cpp
    Color3B toColor3B() const;
    Color4B toColor4B() const;
    Color4F toColor4F() const;
```

halx99:
And
```cpp   
 void set(float r, float g, float b, float a = 1.0F);
    void get(float& r, float& g, float& b) const;
``` changed to
```cpp
 void fromRgba(const Color4F& rgba);
 Color4F toRgba() const;
```

PR #1187, opened 2023-04-25
Refactor HSV/HSL

DelinWorks:
some code in `CCParticleSystemQuad.cpp` depends on changes made in `HSV/HSL` classes.

## Describe your changes
Refactor and improve HSV/HSL const semantics and rename funcs.


core/base/ccTypes.cpp
```diff
@@ -438,112 +427,99 @@ void HSV::set(float r, float g, float b, float a)
     this->a = a;
 }
 
-void HSV::get(float& r, float& g, float& b) const
+const Color4F& HSV::toRgba() const
```
halx99, line 430: Plz don't return reference of temp variable

PR #1242, opened 2023-06-25
[FEATURE] Add mesh renderer instancing support.

DelinWorks:
## Describe your changes
Implemented instanced rendering for `MeshRenderer` class, now foliage and terrain renders much faster! @solan-solan, unfortunately this technique is only supported in OpenGL ES 3.0 so Android has to migrate from OpenGL ES 2.0 to 3.0. 
Angle compatibility DirectX11 is supported and works will.

It works by adding a last attribute after all the vertex attributes with a custom separate buffer, which is just a transformation matrix 4x4 buffer that moves the objects in local space, I only wrote the unlit shader to do that but it's fairly simple, just add an attribute in the vertex shader, i.e.

```glsl
attribute vec4 a_position;
attribute vec2 a_texCoord;
attribute mat4 a_instanceMatrix;
```
`a_instanceMatrix` has to be the last vertex attribute.
The system will automatically set the last item `a_instanceMatrix` to the instance buffer.

This is the API for now.
```cpp
auto m = MeshRenderer::create("rock.obj", "untitled.png");
addChild(m);

m->enableInstancing(MeshMaterial::InstanceMaterialType::UNLIT_INSTANCE, 100);
m->setDynamicInstancing(true); // if you plan to move objects then set this to true.

for (int i = 0; i < 100; i++) {
    auto inst = Node::create();
    m->addInstanceChild(inst);
}
```

Any nodes you alter through actions or anything else (position, rotate, scale) will be applied immediately if instancing is dynamic, otherwise instances will only be rebuilt if a child enters, or when `rebuildInstances()` is called.

https://github.com/axmolengine/axmol/assets/45469625/d9ec8441-cdb3-49ba-850e-2251b5208d8e

The lag was caused by nvidia nsight crashing my pc because of the extremely high poly rocks, it tried to draw an outline for these rocks lol

This will be a draft, any changes or improvements would be appreciated before I write tests!


halx99:
cool!

solan-solan:
@DelinWorks it's amazing that you could forward the feature ahead! I will definitely try it during the week and give you feedback if some ideas appear!

DelinWorks:
https://github.com/axmolengine/axmol/assets/45469625/2b99c43b-3d85-4c74-8759-238227a8da83

I got a pretty good results using a rock mesh with 1000 vertices instanced 30,000 times !!!
Tested on an RTX 3070 and 180~ fps was achieved.

solan-solan:
@DelinWorks I read your code, sorry there were not enough time to launch it yet( I will do it a little later!
There is one thing according to `glVertexAttribDivisor(attribCountPreInst + 0, 1);` It is needed to revert devisor values to 0 in some place after drawing (for example inside `cleanResources()`) Otherwise there will apear issues if some another shader use the same attributes. Check it please, I remember that I have troubles in my tests

DelinWorks:
good catch! also I think `glVertexAttribPointer` also needs it's state reset just like `glVertexAttribDivisor`

solan-solan:
Are you sure about glDisableVertexAttribArray? This function is not called for other attributes. Is it really important and would't it be an overhead?

DelinWorks:
I've read an article and it says it's recommended to reset vertex attrib divisors as well as pointers, since it doesn't cause issues it might actually be a performance overhead, I'll remove it

solan-solan:
@DelinWorks 
I launched your code with cpp-tests. There are some things appear. It is needed to initialize these variables by default
image
The most important is for _instanceMatrixCache, otherwise issue while release pointer operation.

Besides, `Mesh::_instancing` should be initialized to `false` by default (now it is `true`), otherwise general MeshRender objects are not rendered.

Why `MeshRenderer::enableInstancing(MeshMaterial::InstanceMaterialType instanceMat, int count), MeshRenderer::setDynamicInstancing(bool dynamic), MeshRenderer::addInstanceChild(Node* child, bool active), MeshRenderer::rebuildInstances()` affect only first Mesh? Looks like that your solution should to work if MeshRender consists from miltiple meshes, isn't it?

And have you in plane to add the example from screenshort to cpp_test?

DelinWorks:
They're indeed being set to default values, on the constructor of mesh, but `_instancing` was true, my bad!

```
Mesh::Mesh()
    : _skin(nullptr)
    , _visible(true)
    , _instancing(false)
    , _instanceTransformBuffer(nullptr)
    , _instanceCount(0)
    , meshIndexFormat(CustomCommand::IndexFormat::U_SHORT)
    , _meshIndexData(nullptr)
    , _blend(BlendFunc::ALPHA_NON_PREMULTIPLIED)
    , _blendDirty(true)
    , _material(nullptr)
    , _texFile("")
{}
```

and the `AX_SAFE_DELETE_ARRAY(_instanceMatrixCache);` has the word SAFE in it which means it won't delete it unless it's pointing to something.

The rest of the functions should indeed apply to all meshes, but I don't know about `addInstanceChild()`, I think I'll make it so that you need to specify a pointer to a mesh? I really don't know.

DelinWorks:
The hard part about implementing this is to rewrite all the shaders to use instancing. even hardware skinning shaders :<

halx99:
You add two pure virutal function to class `CommandBuffer`: `setInstanceBuffer` and `drawElementsInstanced`. any class inherited from `CommandBuffer` must implment them if it want c++ compiler allow create instance of it. otherwise the child class will still `abstract class`, any c++ compiler will report compiling error when you try to new instance of an abstract class. There are two methods solve the compile error:

1. don't mark `setInstanceBuffer` and `drawElementsInstanced` as pure virutal function, change `virtual void setIndexBuffer(Buffer* buffer) = 0;` to `virtual void setIndexBuffer(Buffer* buffer){}`, and do same change for `drawElementsInstanced`
2. Add empty implementation of `setInstanceBuffer` and `drawElementsInstanced`  to CommandBufferMTL
image

halx99:
> You add two pure virutal function to class `CommandBuffer`: `setInstanceBuffer` and `drawElementsInstanced`. any class inherited from `CommandBuffer` must implment them if it want c++ compiler allow create instance of it. otherwise the child class will still `abstract class`, any c++ compiler will report compiling error when you try to new instance of an abstract class. There are two methods solve the compile error:
> 
> 1. don't mark `setInstanceBuffer` and `drawElementsInstanced` as pure virutal function, change `virtual void setIndexBuffer(Buffer* buffer) = 0;` to `virtual void setIndexBuffer(Buffer* buffer){}`, and do same change for `drawElementsInstanced`
> 2. Add empty implementation of `setInstanceBuffer` and `drawElementsInstanced`  to CommandBufferMTL
>    image

The `CommandBufferMTL` should implement them, the first solution preferred, i.g.

```objc
void CommandBufferMTL::drawElementsInstanced(PrimitiveType primitiveType,
                                             IndexFormat indexType,
                                             std::size_t count,
                                             std::size_t offset,
                                             int instanceCount,
                                             bool /* wireframe */)
{
    prepareDrawing();
    [_mtlRenderEncoder drawIndexedPrimitives:toMTLPrimitive(primitiveType)
                                  indexCount:count
                                   indexType:toMTLIndexType(indexType)
                                 indexBuffer:_mtlIndexBuffer
                           indexBufferOffset:offset
                               instanceCount: instanceCount];
}
```

halx99:
And this PR idea comes from: https://learnopengl.com/Advanced-OpenGL/Instancing , right?

DelinWorks:
> don't mark setInstanceBuffer and drawElementsInstanced as pure virutal function, change virtual void setIndexBuffer(Buffer* buffer) = 0; to virtual void setIndexBuffer(Buffer* buffer){}, and do same change for drawElementsInstanced

What confused me is that it compiled fine the first time I added the virtual functions so I threw that thought out of the window!
and yes this PR is indeed from the article about OpenGL Instancing. I've read that specific article more than once to get instancing to work!

halx99:
> > don't mark setInstanceBuffer and drawElementsInstanced as pure virutal function, change virtual void setIndexBuffer(Buffer* buffer) = 0; to virtual void setIndexBuffer(Buffer* buffer){}, and do same change for drawElementsInstanced
> 
> What confused me is that it compiled fine the first time I added the virtual functions so I threw that thought out of the window! and yes this PR is indeed from the article about OpenGL Instancing. I've read that specific article more than once to get instancing to work!

Nice

solan-solan:
@DelinWorks 

> I think I'll make it so that it applies to all meshes in the mesh renderer

Yes it is the change i talked about. MeshRender should not think about specific instances to my understand.

And I believe that there is no hasty in this feature. The best if you do it with comfortable speed. It works and it is the main=) I will definitely try to integrate it to HeightMap (may be it could be the usage example) and in my axmol projects, but it would not be done right now, even after successfull integration to dev, still the same by the time and priority.

About metal, besides @halx99 mentioned, you could check this example https://metalbyexample.com/instanced-rendering/ , may be it can help. But I could not find _instanced array_ implementation in metal (without this feature there is no correspondence between gles - metal shaders).

And in any way, it would be good to think in future, before integrating to dev, how to close _instance code_ with some macros, otherwise it breaks build on Android by default

DelinWorks:
> About metal, besides @halx99 mentioned, you could check this example https://metalbyexample.com/instanced-rendering/ , may be it can help. But I could not find instanced array implementation in metal (without this feature there is no correspondence between gles - metal shaders).

I can't guarantee that it works, because I don't have a testing environment.

> And in any way, it would be good to think in future, before integrating to dev, how to close instance code with some macros, otherwise it breaks build on Android by default

About that.. how can we switch android to GLES v3.0 @halx99 ?

halx99:
For android, if you add test-case to cpp-tests, we can build apk to test first

DelinWorks:
How TF is android compiling in github CI while I'm getting the error: use of undeclared identifier 'glDrawElementsInstanced'?

halx99:
sorry,  i have do mistake merge when I want check android build at local, will reset later, after reset you can reopen this PR again

DelinWorks:
No worries I'll open another PR

halx99:
@DelinWorks  done, you can recreate another PR from the same branch: `[DelinWorks:instancing-support](https://github.com/DelinWorks/axmol/tree/instancing-support)`

halx99:
> How is android compiling in github CI while I'm getting the error: use of undeclared identifier `glDrawElementsInstanced`?

I known why, because the ci checks for android enable extension `Effekseer` which will add `glad` to include search path. but I think, currently, even though the glDrawElementsInstanced can compiled with `glad` include files, but the app maybe crash when invoke `glDrawElementsInstanced`.

so we still needs find a way to enable GLES3 for android, as I know, the glad may support it but needs regenerated with `loader` enabled from website: https://gen.glad.sh/

[glad-with-loader.zip](https://github.com/axmolengine/axmol/files/11933591/glad-with-loader.zip)

halx99:
I create PR #1253 for you, you can continue working on you branch [DelinWorks:instancing-support](https://github.com/DelinWorks/axmol/tree/instancing-support)

halx99:
> > How is android compiling in github CI while I'm getting the error: use of undeclared identifier `glDrawElementsInstanced`?
> 
> I know why, because the ci checks for android enable extension `Effekseer` which will add `glad` to include search path. but I think, currently, even though the glDrawElementsInstanced can compiled with `glad` include files, but the app maybe crash when invoke `glDrawElementsInstanced`.
> 
> so we still needs find a way to enable GLES3 for android, as I know, the glad may support it but needs regenerated with `loader` enabled from website: https://gen.glad.sh/
> 
> [glad-with-loader.zip](https://github.com/axmolengine/axmol/files/11933591/glad-with-loader.zip)

Don't require, please ignore

DelinWorks:
doesn't android already rely on glad? or does it use it's custom glad generated to support gles2 only?

halx99:
I have update cmake scripts, allow passing `AX_USE_GLAD=TRUE` for android to compile and run

DelinWorks:
so libGLESv2.so already supports gles3 we only need a glad loader?

halx99:
> so libGLESv2.so already supports gles3 we only need a glad loader?

Yes, I can compile the latest axmol with your patch applied and `AX_USE_GLAD=TRUE` to cmake command line, and run on my device:
image

Goto the instancing tests, no crash, but I can't see nothing

halx99:
I have also try apply this patch https://github.com/cocos2d/cocos2d-x/pull/17537, but still doesn't work

DelinWorks:
so it was just a loader that's cool.. good job!

DelinWorks:
odd you said it doesn't show.. did you commit `AX_USE_GLAD` to axmol so I can test?

halx99:
Yes, you should merge latest axmol dev branch to https://github.com/DelinWorks/axmol/tree/instancing-support first, then I will help you to enable AX_USE_GLAD for android

DelinWorks:
I used `set(AX_USE_GLAD ON)` but it didn't work

DelinWorks:
I skipped it for now

image

DelinWorks:
it doesn't show for me too

Screenshot_20230703_161806_CppTests

halx99:
Modify line:  https://github.com/DelinWorks/axmol/blob/instancing-support/core/CMakeLists.txt#L113 from  `option(AX_USE_GLAD "Use glad for platforms which use GL backend" AX_ENABLE_EXT_EFFEKSEER)` to `option(AX_USE_GLAD "Use glad for platforms which use GL backend" ON)`

halx99:
> Modify line: https://github.com/DelinWorks/axmol/blob/instancing-support/core/CMakeLists.txt#L113 from `option(AX_USE_GLAD "Use glad for platforms which use GL backend" AX_ENABLE_EXT_EFFEKSEER)` to `option(AX_USE_GLAD "Use glad for platforms which use GL backend" ON)`

Remember delete cached `AX_USE_GLA` from CMakeCache.txt, compile shoud be passed

DelinWorks:
I think it's now showing because there is an erroneous state we are missing... I'll see what I can do

DelinWorks:
I figured it out, turns out shaders need to follow the modern approach of OpenGL ES v3

the shader predefine `std::string SHADER_PREDEFINE = "#version 100\n precision highp float;\n precision highp int;\n";` in ProgramGL.cpp needs a workaround because I can't specify `#version 300 es` in the shader because of it I had to remove it altogether. 

Also we needed to use layout location instead of attribute so instead of:
```
attribute vec4 a_position;
attribute in vec2 a_texCoord;
attribute in mat4 a_instanceMatrix;
```
we use:
```
layout (location = 0) in vec4 a_position;
layout (location = 1) in vec2 a_texCoord;
layout (location = 2) in mat4 a_instanceMatrix;
```

also there is no varying in modern gl, you need to specify `in` or `out`, there is also no `gl_FragColor` it needs to be a shader variable like `out vec4 FragColor` and writing anything to it will be displayed, I got my phone to render it just fine!

DelinWorks:
I think we're gonna need to rewrite all shaders in modern gl

halx99:
> I think we're gonna need to rewrite all shaders in modern gl to prevent future headaches

Does it break GLESv2 support?

DelinWorks:
of course it does, not if we write two shaders of everything, one in glesv2 and another one for glesv3 for every single shader there is. at this point we'd be fine dropping support for glesv2 since it's pretty much dead.

maybe I'm wrong and we could use a define in shaders but I don't know

DelinWorks:
actually nevermind we don't have to rewrite shaders they work fine as is, it's just features like instancing that require the modern approach so yeah

halx99:
I preferred drop GLESv2 shader

halx99:
with moder shader, it's possible to write one shader with GLSL/HLSL, then use sprivtools to translate metal shader offline, that's would be better than before translated by glsl-optimizer at runtime.

DelinWorks:
ok don't merge yet I'll rewrite shaders in v3

PR #1253, opened 2023-07-03
Add mesh renderer instancing, GLESv3 required

halx99:
Continue #1242 


halx99:
Because this feature require GLESv3, and we decide to drop GLESv2 support, so I mark this PR to 1.1.x
once 1.0.x branch be created, when this PR ready, can be merged into dev

halx99:
I'm NOT sure does glsl-optimizer support translate modern gl shader to metal at runtime

DelinWorks:
hmm that's gonna suck 😕

DelinWorks:
according to this [blog post](https://forum.unity.com/threads/is-opengl-es-2-is-still-needed-in-unity-for-android.976809/) I read it seams that GLES V2 isn't really that popular anymore according to this quote

> The takeaway is that 88% of active devices support GLES 3+. Only 12% are limited to GLES 2. If you have limited time and resources, dropping support for GLES v2 might make sense. You may find older devices driving a disproportionate number of support requests.

We need many opinions from people at this point. @solan-solan do you think dropping GLES v2 is a bad idea? we're dropping it since it's complicated to maintain a version that 12% of devices support, and keep in mind that the 12% will even drop lower in the next years.

halx99:
Try this shader conver tool (compiled from https://github.com/microsoft/ShaderConductor), seems can write shader by `HLSL`, then use it convert to OpenGL, GLES, Metal shader)

[ShaderConductor-f18f7b5_2.zip](https://github.com/axmolengine/axmol/files/11943959/ShaderConductor-f18f7b5_2.zip)

solan-solan:
@DelinWorks 

> do you think dropping GLES v2 is a bad idea?

I really do not know. If you sure that it would be better. But it is needed to fix material also? There is some parsing of shaders now when attributes are saved to internal array. Do you propose to reject gles 2.0 shaders at all? There's no opportunity to leave existing shaders as is and extend material class that it can process gles 3.0 and so on? Version tag could be related from some variable. 
And such question. Is instancing feature really depend on gles v3.0? I launched some example with gles v1.0 and it worked.  Just 'gl_InstanceID' could not be used

halx99:
Can be cloased, it's already merged into latest `dev` branch

PR #1255, opened 2023-07-05
[RENDERER] Translate all axmol shaders to `GLES 310 ES` to support GLESv3

DelinWorks:
## Describe your changes
Currently adding support for GLES 3.0 on android, as such we will need to use a tool like `glslcc` to convert underlying shaders to their platform. These shaders are `SPIRV` compatible!

for android we just load the gles 310 shaders as they are, for windows we use glslcc to convert gles 310 shaders to core/desktop OpenGL shader, for UWP/ANGLE we use glslcc to convert gles 310 shader to compatible gles 300 shaders, and for iOS we use glslcc to convert gles 310 shaders to msl.


rh101:
> Currently adding support for GLES 3.0 on android, as such we will need to use a tool like `glslcc` to convert underlying shaders to their platform.

How will this work exactly? Is it something that occurs at compile-time or at run-time?

Also, applications use their own custom shaders too, along with code so to interact with those shaders.  A guide should be created to detail what needs to change in order to update the existing shaders to gles 310.

DelinWorks:
First of all when you build your game a `shader_cache` folder will be created inside the content folder, if you let's say want a builtin engine shader for windows then a system will run glslcc to compile this shader and put it in `shader_cache`, it will also give you the path to that compiled shader in `shader_cache`, if it's already in `shader_cache` or the hash of the shader didn't change it will instead load the shader in the `shader_cache` instead of running a program again.

this is simple for windows/linux/mac platforms both on build and on runtime, but I don't think we can have `glslcc` in Android/iOS, but I came up with an Idea, shaders inside the content folder under `shaders` folder will get compiled to android on build all of them. and is the case for iOS. and both platforms can look up shaders on `shader_cache` directly, the workflow might seem unintuitive at first but we really need to upgrade from `GLESv2` since features like instancing don't work. and that GLESv2 version is pretty much dead.

Me and halx have been talking about this and we haven't actually dicussed how it can be done specifically, I guess will cross that bridge when we get to it.

rh101:
The reason I asked how it would work is because there are apps they have the shader code embedded as strings inside the C++ code etc.  Since the conversions are being done at compile-time, then shaders embedded within the C++ source can longer be used, and would need to be moved out to individual shader-specific files.  That's just something to be aware of.

DelinWorks:
these apps unfortunately will have to set with axmol `1.0.x` because this feature will be in `1.1.x`, some custom functions can be used to convert embedded C++ shaders to `gles 310` and then `glslcc` but there is also mobile, which kinda makes it difficult really.

rh101:
> anyhow the user has to modify their codebase to get their shaders to work, which no matter how good a migration tool would be (regex match the user's whole codebase to find shaders and change them 💀)

Yes, it wouldn't make sense to keep them embedded in the source, and no point adding anything to cater for that specific situation.  It is best if the shaders are in their own files.

halx99:
about new shader workflow, I think both user and engine shader should trait as shader sources as input of glslcc. we can assume some extension i.e. .frag, .vert to tell glslcc how to compile it. we also can easy tell glslcc build target: gles30, glsl330, msl, and output extension can be fixedb .vert.bin, .frag.bin
user shader source files can be in `Source/shaders`, and the shader folder name not must be shaders, they can put any sub fodder of `Source` because we will choose glslcc by file extension at cmake config time
About shader output directory, just like c++ code, can be ${CMAKE_BINARY_DIR}/shader_bin/
for host platform(osx,win32,linux), we can use soft link to link shader_bin to target directory, and the `shader_bin` directory level shoud be same with `Content`
for mobile platform:
android: can edit gradle to copy folder shader_bin to assets
ios: need check how todo
for uwp: copy to AppX/Content

DelinWorks:
there could be a class that redirects to shader_bin and deals with this stuff automatically, that way when you pass let's say `("shaders/some_shader.vert", "shaders/some_shader.frag")` it would store the complete file path as a hash, and these files will be stored in `shader_bin` as `XXHASH.glsl.vert` or `XXHASH.msl.vert` and same for fragments. if the hash is not found then it would be generated using glslcc, and just like you said a cmake compile step that generates those hashed files and ships them to other platforms!

rh101:
Spent some time thinking more about this.

It would be preferrable to add shader file path to a CMake list, and then pass this list to the relevant section that processes them.  For users who need to include all shaders in a folder, then they can use GLOB, such as:

```
set(SHADER_SRC)
list(APPEND SHADER_SRC
    path/to/shader/MyShader.vert
    path/to/shader/MyShader.frag
     )
```
or 
```
file(GLOB_RECURSE SHADER_SRC path/to/shaders/*.vert)
file(GLOB_RECURSE SHADER_SRC path/to/shaders/*.frag)
```

This means we can have the original shaders stored anywhere we want, but it also means we can use conditional statements to include/exclude specific shaders (such as ones used for testing purposes, or specific platforms etc).  

In addition to that, the CMake script section that copies/soft-links the files that end up in the `shader_bin` directory should be flexible to allow pre-processing of those shader files, or to disallow the copy/soft-linking altogether.  For instance, after the files are generated in `shader_bin`, we may want to run another script on top of this to modify those shader files (such as obfuscation).  We also may want to read the contents of the files in `shader_bin`, and generate cpp source files from them to still allow them to be embedded within the source code during the compilation process, although this would require that the shader processing occurs before the source code build begins (so a pre-build step).

Does this make sense?

DelinWorks:
I agree with the idea, but I'm curious about the specific situations where users would need shaders to be included in source files. In my opinion, it seems unnecessary and impractical, potentially leading to significant technical challenges in the codebase. However, for shader obfuscation purposes, we could consider implementing a packing system that creates a single file, like 'shader_res.bin', containing all the shaders. Then, during runtime, we can load any shader from that packed file. Additionally, we can explore other packing techniques to optimize shader loading performance at runtime.

rh101:
> In my opinion, it seems unnecessary and impractical, potentially leading to significant technical challenges in the codebase.

How so? It has no impact on the shaders provided by Axmol, since they'll still be added to the `shader_bin` folder, and copied to the output, and used in whatever default way the engine handles them.  There is absolutely no impact or technical challenges in the codebase.  

What I proposed in my post above is related to custom shaders, and controlling the final output format, whether it be `.bin`, or embedded directly in the source.  Also, shader processing doesn't have to happen all in one go.  One pre-build step can be used for the shaders provided by Axmol, and another can be used for game-side custom shaders.  They shouldn't be lumped together, and this will allow them to be treat differently.  By default, the custom shaders will also be outputted to the `shader_bin` folder, but this should be modifiable by the developer.

> However, for shader obfuscation purposes, we could consider implementing a packing system that creates a single file, like 'shader_res.bin', containing all the shaders. Then, during runtime, we can load any shader from that packed file. Additionally, we can explore other packing techniques to optimize shader loading performance at runtime. while of course that packed file will have all platform shaders compiled.
 
That isn't entirely necessary, and perhaps something for the future if absolutely required.  It may only benefit really large shader implementations that would be slow to load etc..  Also, having the engine take control of how user shaders are stored is not ideal, and it's something that should be controlled by the developer.

Also, glslcc cmake script supports outputting directly to C source files, and adding them to the target.  While this may not be useful for shaders included with Axmol, it may be something that is used for custom shaders:

Info from [here](https://github.com/septag/glslcc/blob/master/cmake/glslcc.cmake):
```
glslcc_target_compile_shaders_h(target source_files):
#       compile shaders as C header and add them to the target
```

DelinWorks:
oh thanks for clarification! I misunderstood at first but yeah I get what you mean, also I didn't know that glslcc supported that out of the box! I would make the implementation much simpler! and yeah the packingoart isn't important

rh101:
I was just going through glslcc to see what it is capable of, and came across those settings regarding the C file output etc..  Also, I personally use this [tool](https://github.com/laurentlb/Shader_Minifier) to 'minify' the custom shaders, since some are a bit too big, and any reduction in their size helps.  I was just considering how this would all work together with the glslcc, which is why I was looking into it.  So, my pipeline would be minify->glslcc conversion->output (whether it be `.bin` or otherwise).  The minifier may not even be required if the output is optimised in some way by glslcc.

halx99:
> I was just going through glslcc to see what it is capable of, and came across those settings regarding the C file output etc.. Also, I personally use this [tool](https://github.com/laurentlb/Shader_Minifier) to 'minify' the custom shaders, since some are a bit too big, and any reduction in their size helps. I was just considering how this would all work together with the glslcc, which is why I was looking into it. So, my pipeline would be minify->glslcc conversion->output (whether it be `.bin` or otherwise). The minifier may not even be required if the output is optimised in some way by glslcc.

I add a cmake function `ax_add_shader` for compile shader with `glslcc`:

https://github.com/axmolengine/axmol/commit/4cf6060e4288796ba38c096b27d35a525a6620a6

rh101:
> I add a cmake function `ax_add_shader` for compile shader with `glslcc`:

That was super quick! Wasn't expecting this to be done so soon.  I'll test this out when I get a chance to do so.

DelinWorks:
still more to go! this migration is invalid since glslcc compiles shaders from glsl source code not c code so we have to move them to their own shader file and rewrite parts of the internal shader system in axmol, also we need to make a class that appropriately translates shader paths to runtime shader paths on dev and on release

halx99:
> > I add a cmake function `ax_add_shader` for compile shader with `glslcc`:
> 
> That was super quick! Wasn't expecting this to be done so soon. I'll test this out when I get a chance to do so.

You need run setup.ps1 again to install `glslcc` into `tools/external/glslcc/`

rh101:
> still more to go!

There is no doubt that there is more to do, but latest commit with the glslcc can be used to test out the ability to transform the shaders.  This change doesn't actually process the shaders in axmol, as it just seems to be laying the groundwork for the future changes.

> his migration is invalid since glslcc compiles shaders from glsl source code not c code so we have to move them to their own shader file

Sorry, I'm at a complete loss as to what you mean by that.  The only one I can find that is embedded in the cpp source is located in `imgui_impl_ax.cpp`, while the rest are in individual files located in the path `core\renderer\shaders`.  

> also we need to make a class that appropriately translates shader paths to runtime shader paths on dev and on release for each platform

Why would there be any need to translate them, or even a class to be added?  The converted shaders will always be in the same relative location on all platforms, just as they are at the moment. Nothing changes, except perhaps the referenced filenames (such as `.frag.bin` instead of `.frag` if that is how they'll be named), and this can be hard-coded in the engine as it is right now in `core\renderer\Shaders.cpp`.

DelinWorks:
The source files in `core/shaders` are not in pure glsl source code so they have to be converted and the macros too (somehow)

image


also when the shaders are generated: `3>Compiling runtime shader C:/dev/axmol/core/renderer/shaders/3D_color.frag for GLSL330 to C:/dev/axmol/tests/cpp-tests/proj.win32/build/shaders/runtime/3D_color_fs.bin ...`

and your program in loading a shader from the path: `shader/3D_color.frag` how can you load the correct shader that is setting in the build directory under runtime?

rh101:
> The source files in `core/shaders` are not in pure glsl source code so they have to be converted and the macros too (somehow)

Sorry! I understand what you mean now.  I didn't realise that the shaders are set up as C++ literal strings! 

> and your program is loading a shader from the path: `shader/3D_color.frag` how can you load the correct shader that is setting in the build directory under runtime? on release or something it can be moved to `shader_bin` in the content folder and that would also require some path manipulation, let alone different platforms.

From my point earlier, we, as developers, already know that the files will be, for example, "xxxxx_fs.bin".  They aren't randomly named.  So, for your example, we don't reference `shader/3D_color.frag`, we actually reference `runtime/3D_color_fs.bin`, because that filename and path are *exactly* the same on all platforms, and it will never change.  There is absolutely no point referencing `shader/3D_color.frag` and doing anything fancy to map it to `runtime/3D_color_fs.bin`, because 100% of the time it's always going to be `runtime/3D_color_fs.bin`, so just reference that file directly.

DelinWorks:
the shaders reside in `runtime/shaders/[LANG]/xxxxx_fs.bin` how would you know that thr last text after that last slash is deepest of all subpaths? 

ok, if the user references `shader/mycollection/sky.frag` and they're on windows, it would reference `runtime/shader/glsl/sky.frag` so my question is do we replace the first occurrence of word shader with `runtime/shader/glsl/`?

rh101:
> you might be building the project on multiple platforms and compiling shaders between one another might vause conflicts

That's a non-issue.  The cmake output build folder will not be the same for each platform, and given that the shaders will exist within the build folder, they never clash.

DelinWorks:
you're correct

DelinWorks:
that's why discord is better for communication its faster

halx99:
> > you might be building the project on multiple platforms and compiling shaders between one another might vause conflicts
> 
> That's a non-issue. The cmake output build folder will not be the same for each platform, and given that the shaders will exist within the build folder, they never clash.
> 
> I think it needs to be thought out more, because it seems like complexity is being introduced for no apparent reason at all. More specifically, the output path name and output file names.
> 
> So, just a few points:
> 
> 1. Why don't we keep the shader output filename exactly the same as the input filename?
> 2. The language sub-folder is not required, unless we're supporting multiple rendering backends for the same platform, for example, OpenGL and Metal, and which one is used is decided at runtime.  In this case, we do need the shader language subfolder, so `runtime/glsl` and `runtime/metal`.  The only difference is which of these paths is added to the search list (`FileUtils::addSearchPath`), so if it's going to use the Metal renderer, then the code adds the `runtime/metal/` path to the search list, but if OpenGL is being used, then `runtime/glsl` is added instead.  When referencing the shader, it would simply be "Shader.frag", and the correct one loads since it's being read from the only path that exists in the search list.
> 
> So, let's use `shaders/3D_color.frag` as an example for OpenGL:
> 
> 1. It is converted and put into `runtime/glsl/3D_color.frag` (assuming the output filename doesn't need to have an extension of "bin").
> 2. The entire `runtime` folder is copied to the assets folder (is this actually how it was planned? or only the shaders were going to be copied to the assets folder without the `runtime/[LANG]/` path?)
> 3. Since it's OpenGL, add the `runtime/glsl/` to the search list via `FileUtils::addSearchPath`
> 4. The C code references `3D_color.frag`, and it's automatically loaded from the correct folder.
> 
> Is there something not right about this setup?

Latest commits, I remove subfolder, and by default will flat compiled to `${CMAKE_BINARY_DIR}/shaders` and no extension `.bin`, looks like:

image

The suffix `_fs` , `_vs` controled by glslcc, I don't how to control manually yet(maybe don't need care about it)

rh101:
> The suffix `_fs` , `_vs` controled by glslcc, I don't how to control manually yet(maybe don't need care about it)

True, it won't matter what changes glslcc makes to the filenames, as long as they are consistent, because we just reference the `_fs` and `_vs` names directly in the the code.  So, instead of referencing something like `color.frag", we just use "color_fs" directly, since that is always what it will be called.  Does that make sense?

halx99:
for developer targets: win,linux,osx shaders needs link to exe  dir, and load from exe  dir , not use Content folder, for cross platform: ios,android,uwp can just copy to 
Content

halx99:
for developer targets: win,linux,osx shaders needs link to exe  dir, and load from exe  dir , not use Content folder, for cross platform: ios,android,uwp can just copy to 
Content

rh101:
To avoid cluttering the exe dir with individual shader files, should we have them inside a "shaders" or similar sub-dir?

For example, on Windows/Linux/OSX, it would be:
```
App/
├─ app.exe
├─ shaders/
│  ├─ shader1_fs
│  ├─ shader1_vs
```

and on iOS/Android/UWP:
```
App/
├─ app.exe
Content/
├─ shaders/
│  ├─ shader1_fs
│  ├─ shader1_vs
```

The path referenced in the code would be always be `shaders/[ShaderFileName]`. Would this work?

halx99:
> To avoid cluttering the exe dir with individual shader files, should we have them inside a "shaders" or similar sub-dir?
> 
> For example, on Windows/Linux/OSX, it would be:
> 
> ```
> App/
> ├─ app.exe
> ├─ shaders/
> │  ├─ shader1_fs
> │  ├─ shader1_vs
> ```
> 
> and on iOS/Android/UWP:
> 
> ```
> App/
> ├─ app.exe
> Content/
> ├─ shaders/
> │  ├─ shader1_fs
> │  ├─ shader1_vs
> ```
> 
> The path referenced in the code would be always be `shaders/[ShaderFileName]`. Would this work?

Correct, I'm trying make it work, will create branch on `halx99/axmol` once available for test.

halx99:
The branch created: https://github.com/halx99/axmol/tree/1.1-dev, can test on linux, win32, but can't see anything because the compiled shader not compatible with current render flow.

rh101:
> The branch created: https://github.com/halx99/axmol/tree/1.1-dev, can test on linux, win32, but can't see anything because the compiled shader not compatible with current render flow.

All good.  Is there any specific rush on getting this all working? If not, then we can take our time with it.

DelinWorks:
yeah the main goal is to get our feet wet and hopefully the rest can simple to implement.

PR #1275, opened 2023-07-16
Fix for adding axmol bin path to user path environment variable

rh101:
## Describe your changes
Invert if condition checking for existence of the axmol bin path in the user path env variable
Make list mutable to allow for inserting new path

Discussion at #1274 


halx99 (approved):
lgtm

PR #1330, opened 2023-09-14
Fix issue where RichText URL touch listener does not activate on touch

### LINKED ISSUE #1327: RichText URL link will not activate on click in certain conditions
- axmol version: All current
- devices test on:
- developing environments
   - NDK version: r23c
   - Xcode version: 14.2
   - Visual Studio: 
      - VS version: 2019 (16.11), 2022 (17.6+)
      - MSVC version: 1929, 1934, 19.35, 19.36, 19.37
   - cmake version: 
Steps to Reproduce:

```
auto visibleSize = _director->getVisibleSize();

auto* layout = ui::Layout::create();
layout->setContentSize(visibleSize);
layout->setAnchorPoint(Vec2::ANCHOR_MIDDLE);
layout->setPosition(Vec2(visibleSize.width / 2, visibleSize.height));
this->addChild(layout);

auto* listener = ax::EventListenerTouchOneByOne::create();
listener->setSwallowTouches(true);
listener->onTouchBegan = [](Touch*, Event*) -> bool { return true; };
_eventDispatcher->addEventListenerWithSceneGraphPriority(listener, layout);

auto* richText = ui::RichText::createWithXML("<a href='http://www.google.com'>click me</a>");
richText->ignoreContentAdaptWithSize(false);
richText->setContentSize(Size(500, 200));
richText->setPositionNormalized(Vec2(0.5f, 0.5f));

layout->addChild(richText);
```

If a `RichText` with a clickable URL is a in the scene, and there is another node in the scene that has a `EventListenerTouchOneByOne` listener, and the `onTouchBegan` returns true (along with `setSwallowTouches(true)`), then the URL handler will not be triggered.

The URL touch handler is implemented with this code (in `UIRichText.cpp`, class `ListenerComponent`):
```
_touchListener                 = ax::EventListenerTouchAllAtOnce::create();
_touchListener->onTouchesEnded = AX_CALLBACK_2(ListenerComponent::onTouchesEnded, this);
Director::getInstance()->getEventDispatcher()->addEventListenerWithSceneGraphPriority(_touchListener, _parent);
```

All `EventListenerTouchAllAtOnce` listeners will never be triggered if any `EventListenerTouchOneByOne` exist that returns true on `onTouchBegan`.

The reason is in function `EventDispatcher::dispatchTouchEvent(EventTouch* event)`, where the dispatcher first calls all `EventListenerTouchOneByOne` listeners, and then if any return `true` for `onTouchBegan`, it removes the entries from a variable named `mutableTouches` (which holds all current touches).

Once it loops through all `EventListenerTouchOneByOne` listeners, it then gets to the following code which should handle all the `EventListenerTouchAllAtOnce` event listeners:

```
if (allAtOnceListeners && !mutableTouches.empty())
{
... handles all `EventListenerTouchAllAtOnce` here
}
```

The problem is that the `if` condition is false, since `mutableTouches` is empty, so it never enters that block, and no `EventListenerTouchAllAtOnce` listeners ever receive the touch inputs.

The main issue is that `EventListenerTouchOneByOne` and `EventListenerTouchAllAtOnce` are not sorted together (by priority and Z order), so even if a node using `EventListenerTouchAllAtOnce` is on top of every other node in the scene, it would never be called because of other nodes using `EventListenerTouchOneByOne`.  This is what doesn't make sense to me.  

My question is, does anyone know why `EventListenerTouchAllAtOnce` was used for the URL touch handler?  If `EventListenerTouchOneByOne` is used, then this issue does not occur, but I'm not sure if there would be any side-effects.  

All `Widget` sub-classes (like `ui::Button` etc) use `EventListenerTouchOneByOne`, and they all work as intended, even to call `openURL`.

One other thing, `EventListenerTouchAllAtOnce` is *ONLY* used for the `RichText` URL handler; I can't find any usage of it at all in the Axmol source code (not including the `cpp_tests` project of course).

rh101:

- `dev`: Current 2.0 BugFixs or Features

## Describe your changes
Due to the way that touch input is dispatched to touch event listeners, instances of `EventListenerTouchAllAtOnce` will not receive touch inputs until after all instances of `EventListenerTouchOneByOne` are processed.  If any `EventListenerTouchOneByOne` listeners return `true` for the `onTouchBegin` handler, and also are set to swallow touches, then that touch is no longer available for when listeners of type `EventListenerTouchAllAtOnce` are processed. 

The RichText URL touch handler was of type `EventListenerTouchAllAtOnce`, so in a scene where other handlers exist of type `EventListenerTouchOneByOne`, the URL handler may not receive touch inputs.

The fix to this problem is to use `EventListenerTouchOneByOne` for the URL touch handling, and the implementation should behave in exactly the same way as before.

#1327 

### DIFF
--- a/core/ui/UIRichText.cpp
+++ b/core/ui/UIRichText.cpp
@@ -48,63 +48,77 @@
 USING_NS_AX;
 using namespace ax::ui;
 
-class ListenerComponent : public Component
+class UrlTouchListenerComponent : public Component
 {
 public:
     static const std::string COMPONENT_NAME; /*!< component name */
 
-    static ListenerComponent* create(Node* parent,
+    static UrlTouchListenerComponent* create(Node* parent,
                                      std::string_view url,
-                                     const RichText::OpenUrlHandler handleOpenUrl = nullptr)
+                                     RichText::OpenUrlHandler handleOpenUrl = nullptr)
     {
-        auto component = new ListenerComponent(parent, url, std::move(handleOpenUrl));
+        auto* component = new UrlTouchListenerComponent(parent, url, std::move(handleOpenUrl));
+        component->init();
         component->autorelease();
         return component;
     }
 
-    explicit ListenerComponent(Node* parent, std::string_view url, const RichText::OpenUrlHandler handleOpenUrl)
+    explicit UrlTouchListenerComponent(Node* parent, std::string_view url, RichText::OpenUrlHandler handleOpenUrl)
         : _parent(parent), _url(url), _handleOpenUrl(std::move(handleOpenUrl))
     {
-        setName(ListenerComponent::COMPONENT_NAME);
-
-        _touchListener                 = ax::EventListenerTouchAllAtOnce::create();
-        _touchListener->onTouchesEnded = AX_CALLBACK_2(ListenerComponent::onTouchesEnded, this);
-
-        Director::getInstance()->getEventDispatcher()->addEventListenerWithSceneGraphPriority(_touchListener, _parent);
-        _touchListener->retain();
     }
 
-    virtual ~ListenerComponent()
+    bool init() override
+    {
+        if (!Component::init())
+            return false;
+
+        setName(UrlTouchListenerComponent::COMPONENT_NAME);
+
+        _touchListener               = ax::EventListenerTouchOneByOne::create();
+        _touchListener->onTouchBegan = AX_CALLBACK_2(UrlTouchListenerComponent::onTouchBegan, this);
+        _touchListener->onTouchEnded = AX_CALLBACK_2(UrlTouchListenerComponent::onTouchEnded, this);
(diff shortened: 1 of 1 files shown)


rh101:
If this fix is accepted, is there any chance it can also be back-ported to 1.0?

halx99:
> If this fix is accepted, is there any chance it can also be back-ported to 1.0?

sure

rh101:
> > If this fix is accepted, is there any chance it can also be back-ported to 1.0?
> 
> sure

Should I create a new PR with the same changes for 1.0?

halx99:
> > > If this fix is accepted, is there any chance it can also be back-ported to 1.0?
> > 
> > 
> > sure
> 
> Should I create a new PR with the same changes for 1.0?

yes, that could be better

halx99:
you can try use git cherry pick first, if not work then sync code manually

PR #1353, opened 2023-09-24
[IMPROVEMENT] Prevent Label Extra Pixels

DelinWorks:

- `dev`: Current 2.0 BugFixs or Features

## Describe your changes

When creating a label using `createWithBMFont` I noticed that space characters are being drawn even though they don't have any opaque pixels.

You can see in the image below that the character `' '` is being accounted for and rendered which wastes GPU resources. (in this case GPU pixel fill rate is wasted)
image

And this is after making sure that the character `' '` is invalid which will make it not render.
image


halx99:
does it will affect .ttf?

halx99 (approved):

DelinWorks:
sorry I didn't get time to test it but yeah it should work since the code I modified only gets executed for bm fonts afaik

PR #1379, opened 2023-10-07
Use std containers instead uthash

halx99:

- [x] sch update list, use insert_sorted aka SortedList for priority sched


core/2d/ActionManager.h
```diff
@@ -36,9 +36,14 @@ THE SOFTWARE.
 
 NS_AX_BEGIN
 
-class Action;
-
-struct _hashElement;
+struct _ActionHandle
```
j-jorge: Identifiers beginning with an underscore followed by an uppercase letter are always reserved in C++. You should drop the underscore here. If the idea is to have some "internal" type I would suggest to use a `detail` namespace.

core/base/Scheduler.h
```diff
@@ -156,6 +156,34 @@ struct _listEntry;
 struct _hashSelectorEntry;
 struct _hashUpdateEntry;
 
+// data structures
+
+// A list double-linked list used for "updates with priority"
```
j-jorge: There are two many "lists" in this sentence: A ~list~ double-linked list used for "updates with priority"

core/base/Vector.h
```diff
@@ -255,6 +255,12 @@ class Vector
         return _data[index];
     }
 
+    T operator[](size_t index) const
+    {
+        AXASSERT(index >= 0 && index < size(), "index out of range in getObjectAtIndex()");
```
j-jorge: `index >= 0` is always true with size_t.

thirdparty/yasio/pod_vector.hpp
```diff
@@ -162,6 +162,43 @@ class pod_vector {
     }
     return _Where;
   }
+  iterator insert(iterator _Where, const value_type& _Val)
```
j-jorge: Again here, all these _Identifiers cause undefined behavior.

core/base/Scheduler.h
halx99: the comments should be remove

core/base/Scheduler.h
```diff
@@ -32,10 +32,10 @@ THE SOFTWARE.
 #include <functional>
 #include <mutex>
 #include <set>
-
+#include <list>
```
halx99:
```suggestion
```

PR #1409, opened 2023-10-30
[Android] Enable support for restarting application process if OpenGL context is lost

rh101:

## Describe your changes
~~Call native renderer before the sleep in onDrawFrame due to possibility of invalid texture data after delay.~~ This will be in a separate PR

Added `ProcessPhoenix.java`, from this [project](https://github.com/JakeWharton/ProcessPhoenix), which allows the process to be restarted.

Dispatch event if application process will be restarted (`EVENT_APP_RESTARTING`). 

Added `AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST` configuration option, which is disabled if `AX_ENABLE_CACHE_TEXTURE_DATA` is enabled, since there is no point having both enabled.  The default is that `AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST` is disabled.

Question: Do we want the default to be that the application always restarts if the context is lost? That seems to be how other engines handle it (Unity3D, Godot etc).

Related to #1211

### DIFF
--- a/core/platform/PlatformMacros.h
+++ b/core/platform/PlatformMacros.h
@@ -94,6 +94,16 @@ Copyright (c) 2021 Bytedance Inc.
 #    define AX_ENABLE_CACHE_TEXTURE_DATA 0
 #endif
 
+/** @def AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST
+ * Enable this if application should be restarted after the OpenGL context is lost
+ *
+ */
+#if (AX_TARGET_PLATFORM == AX_PLATFORM_ANDROID && !AX_ENABLE_CACHE_TEXTURE_DATA)
+#    define AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST 1
+#else
+#    define AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST 0
+#endif
+
 #if (AX_TARGET_PLATFORM == AX_PLATFORM_ANDROID) || (AX_TARGET_PLATFORM == AX_PLATFORM_WIN32)
 /** Application will crash in glDrawElements function on some win32 computers and some android devices.
  *  Indices should be bound again while drawing to avoid this bug.
--- a/core/platform/android/java/src/org/axmol/lib/AxmolEngine.java
+++ b/core/platform/android/java/src/org/axmol/lib/AxmolEngine.java
@@ -60,6 +60,7 @@ import com.android.vending.expansion.zipfile.APKExpansionSupport;
 import com.android.vending.expansion.zipfile.ZipResourceFile;
 
 import com.enhance.gameservice.IGameTuningService;
+import com.jakewharton.processphoenix.ProcessPhoenix;
 
 import java.io.IOException;
 import java.io.File;
@@ -772,6 +773,10 @@ public class AxmolEngine {
 
     public static int getSDKVersion() {
         return Build.VERSION.SDK_INT;
+    }
+
+    public static void restartProcess() {
+        ProcessPhoenix.triggerRebirth(sActivity);
     }
 
     private static AxmolAccelerometer getAccelerometer() {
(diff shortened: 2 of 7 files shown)


core/platform/PlatformMacros.h
```diff
@@ -94,6 +94,16 @@ Copyright (c) 2021 Bytedance Inc.
...
 #endif
 
+/** @def AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST
+ * Enable this if application should be restarted after the OpenGL context is lost
+ *
+ */
+#if (AX_TARGET_PLATFORM == AX_PLATFORM_ANDROID && !AX_ENABLE_CACHE_TEXTURE_DATA)
+#    define AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST 1
+#else
+#    define AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST 0
```
crazyhappygame, line 104: I can find find place where this define is used

core/platform/android/java/src/org/axmol/lib/AxmolEngine.java
```diff
@@ -774,6 +775,10 @@ public static int getSDKVersion() {
         return Build.VERSION.SDK_INT;
     }
 
+    public static void restartProcess() {
```
crazyhappygame, line 778: I can not find place where this method is called

core/platform/android/java/src/org/axmol/lib/AxmolRenderer.java
```diff
@@ -91,18 +97,18 @@ public void onDrawFrame(final GL10 gl) {
         } else {
             final long now = System.nanoTime();
             final long interval = now - this.mLastTickInNanoSeconds;
-        
+
+            /*
```
crazyhappygame: Can we do this change in separate PR?

core/platform/PlatformMacros.h
```diff
@@ -94,6 +94,16 @@ Copyright (c) 2021 Bytedance Inc.
 #    define AX_ENABLE_CACHE_TEXTURE_DATA 0
 #endif
 
+/** @def AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST
+ * Enable this if application should be restarted after the OpenGL context is lost
+ *
+ */
+#if (AX_TARGET_PLATFORM == AX_PLATFORM_ANDROID && !AX_ENABLE_CACHE_TEXTURE_DATA)
```
crazyhappygame, line 101: Can we make restart on by default?
Can we remove AX_ENABLE_CACHE_TEXTURE_DATA and code related to it in next PR?

core/platform/android/javaactivity-android.cpp
```diff
@@ -101,10 +101,29 @@ JNIEXPORT void JNICALL Java_org_axmol_lib_AxmolRenderer_nativeInit(JNIEnv*, jcla
...
 }
 
+JNIEXPORT void JNICALL Java_org_axmol_lib_AxmolRenderer_nativeOnContextLost(JNIEnv*, jclass)
+{
+#if AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST
+    auto director = ax::Director::getInstance();
+    ax::EventCustom recreatedEvent(EVENT_APP_RESTARTING);
+    director->getEventDispatcher()->dispatchEvent(&recreatedEvent);
+
+    //  Pop to root scene, replace with an empty scene, and clear all cached data before restarting
```
crazyhappygame, line 117: Why this section is needed? Why we have to do any cleanup if app will be restarted?

core/platform/android/javaactivity-android.cpp
```diff
@@ -101,10 +101,29 @@ JNIEXPORT void JNICALL Java_org_axmol_lib_AxmolRenderer_nativeInit(JNIEnv*, jcla
...
+#endif
     }
 }
 
+JNIEXPORT void JNICALL Java_org_axmol_lib_AxmolRenderer_nativeOnContextLost(JNIEnv*, jclass)
+{
+#if AX_ENABLE_RESTART_APPLICATION_ON_CONTEXT_LOST
+    auto director = ax::Director::getInstance();
+    ax::EventCustom recreatedEvent(EVENT_APP_RESTARTING);
+    director->getEventDispatcher()->dispatchEvent(&recreatedEvent);
```
crazyhappygame, line 115: What is the purpose of this event? Is there any chance to process this event before app restart?

crazyhappygame (commented):
Thank you.

PR #1425, opened 2023-11-08
Add virtual file system support to WICImageLoader

rh101:

## Describe your changes
The existing implementation of `WICImageLoader::encodeImageData` uses file streams that bypass FileUtils, so they do not support virtual file systems.  The modifications in this PR will route the file access routines through the `ax::IFileStream` interface, which does support virtual file systems.

Note, I've tested the code, and it works as expected, but I haven't written COM code before, so if possible, please review the implementation of `SimpleFileStream` in `WICImageLoader-winrt.cpp`.  

### DIFF
--- a/core/platform/winrt/WICImageLoader-winrt.cpp
+++ b/core/platform/winrt/WICImageLoader-winrt.cpp
@@ -127,8 +127,7 @@ public:
 
     HRESULT Read(void* pv, ULONG cb, ULONG* pcbRead) override
     {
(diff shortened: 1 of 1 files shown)


halx99:
lgtm

rh101:
> lgtm

Great!

Just FYI, I used these references for info regarding IStream etc:
https://learn.microsoft.com/en-us/windows/win32/api/objidl/nn-objidl-istream
https://learn.microsoft.com/en-us/windows/win32/stg/istream-compound-file-implementation
https://learn.microsoft.com/en-us/windows/win32/api/objidl/nn-objidl-isequentialstream

PR #1472, opened 2023-12-04
Reduce memory usage when using getFileMD5Hash function

rh101:

## Describe your changes
Currently, if `ax::utils::getFileMD5Hash()` is run on a file, then that entire file is loaded into memory.  If the file is very large, it results in sudden high memory usage spikes, which can cause memory issues.  To avoid this problem, I propose that we instead stream the file in smaller chunks.

The chunk size used is configurable, but is set at a default of 16KB.

### DIFF
--- a/core/base/Utils.cpp
+++ b/core/base/Utils.cpp
@@ -468,23 +468,15 @@ std::string computeFileDigest(std::string_view filename, std::string_view algori
     }
 
     int bytesRead;
-    auto totalSize = 0;
     while ((bytesRead = fs->read(buffer.get(), bufferSize)) > 0)
     {
-        totalSize += bytesRead;
         EVP_DigestUpdate(mdctx, buffer.get(), bytesRead);
     }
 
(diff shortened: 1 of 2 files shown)


halx99:
Why not implement a more common API `computeFileDigest` like exist `computeDigest`, then invoke it in getFileMD5Hash

rh101:
> Why not implement a more common API `computeFileDigest` like exist `computeDigest`, then invoke it in getFileMD5Hash

Good idea. I'll update it.

core/base/Utils.cpp
```diff
@@ -441,46 +479,14 @@ std::string getFileMD5Hash(std::string_view filename, uint32_t bufferSize)
     {
```
halx99: don't need check totalSize, empty file also have digest

rh101: I understand. Updating it now.

halx99 (approved):

PR #1489, opened 2023-12-08
fix scene, imports, comments and others

paulocoutinhox:

- `1.x`: 1.x BugFixs

## Describe your changes

- fix scene, imports, comments and others
- remove the use of cocos2d and use axmol
- keep the pattern filename/classname as same
- tested cpp/lua on ios/macos

### DIFF
--- a/templates/cpp-template-default/Source/AppDelegate.cpp
+++ b/templates/cpp-template-default/Source/AppDelegate.cpp
@@ -32,12 +32,10 @@
 #    include "audio/AudioEngine.h"
 #endif
 
-USING_NS_AX;
-
-static ax::Size designResolutionSize = ax::Size(1280, 720);
-static ax::Size smallResolutionSize  = ax::Size(480, 320);
-static ax::Size mediumResolutionSize = ax::Size(1024, 768);
-static ax::Size largeResolutionSize  = ax::Size(2048, 1536);
+static Size designResolutionSize = Size(1280, 720);
+static Size smallResolutionSize  = Size(480, 320);
+static Size mediumResolutionSize = Size(1024, 768);
+static Size largeResolutionSize  = Size(2048, 1536);
 
 AppDelegate::AppDelegate() {}
 
@@ -63,7 +61,7 @@ bool AppDelegate::applicationDidFinishLaunching()
 #if (AX_TARGET_PLATFORM == AX_PLATFORM_WIN32) || (AX_TARGET_PLATFORM == AX_PLATFORM_MAC) || \
     (AX_TARGET_PLATFORM == AX_PLATFORM_LINUX)
         glView = GLViewImpl::createWithRect(
-            "HelloCpp", ax::Rect(0, 0, designResolutionSize.width, designResolutionSize.height));
+            "HelloCpp", Rect(0, 0, designResolutionSize.width, designResolutionSize.height));
 #else
         glView = GLViewImpl::create("HelloCpp");
 #endif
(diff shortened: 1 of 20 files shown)


rh101:
There changes in this PR which are not correct, and others, in my opinion, not necessary. I'll add a comment on each part separately.

@halx99 Please don't merge this in as it is.

aismann:
Replacing ax::"Method()" with "Method()" is  no good idea.
@paulocoutinhox 
What is the reason?

Example:
image

paulocoutinhox:
> Replacing ax::"Method()" with "Method()" is no good idea. @paulocoutinhox What is the reason?

Hi, in my opinion and code style, if we are importing namespace is because we don't want use namespace on every type. Don't make sense to me still prefix classes with namespaces if we already imported it.

aismann:
> > Replacing ax::"Method()" with "Method()" is no good idea. @paulocoutinhox What is the reason?
> 
> Hi, in my opinion and code style, if we are importing namespace is because we don't want use namespace on every type. Don't make sense to me still prefix classes with namespaces if we already imported it.

Using a third party source which using also a `Size()` method like axmol => compiler does not know which one is meaning.

rh101:
> Hi, in my opinion and code style, if we are importing namespace is because we don't want use namespace on every type. Don't make sense to me still prefix classes with namespaces if we already imported it.

That is a personal preference, and it's fine if you want to use that in your own code, but this specific file is part of the template, and making a change like this is forcing your personal preference on every other developer who is creating a project.

paulocoutinhox:
> > > Replacing ax::"Method()" with "Method()" is no good idea. @paulocoutinhox What is the reason?
> > 
> > 
> > Hi, in my opinion and code style, if we are importing namespace is because we don't want use namespace on every type. Don't make sense to me still prefix classes with namespaces if we already imported it.
> 
> Using a third party source which using also a `Size()` method like axmol => compiler does not know which one is meaning.

I personally prefer to leave the project clean and less verbose and adjust external things to suit my project rather than leaving the project with things that are unnecessary for the context. It's better to use the prefix/namespace only where you really need it than having to use it for the majority of our code on a daily basis. If you have something external with the same name, then use it for what is external and not for the day-to-day code, which is most of it.

paulocoutinhox:
> > Hi, in my opinion and code style, if we are importing namespace is because we don't want use namespace on every type. Don't make sense to me still prefix classes with namespaces if we already imported it.
> 
> That is a personal preference, and it's fine if you want to use that in your own code, but this specific file is part of the template, and making a change like this is forcing your personal preference on every other developer who is creating a project.

I understand, but each of you also has your preferences. I can't make a PR with mine? Are they worse? Is there any document here saying which code style should be followed?

rh101:
> I personally prefer to leave the project clean and less verbose and adjust external things to suit my project rather than leaving the project with things that are unnecessary for the context. It's better to use the prefix/namespace only where you really need it than having to use it for the majority of our code on a daily basis. If you have something external with the same name, then use it for what is external and not for the day-to-day code, which is most of it.

> I understand, but each of you also has your preferences. I can't make a PR with mine? Are they worse? Is there any document here saying which code style should be followed?

Once again, personal preference should not be a factor in making changes to code that is used by others.  It's definitely not a good enough reason to make a change like this, because you are forcing your personal preferences on all other developers.

rh101:
It was unnecessary to close this, because some changes made sense, such as changing `HelloWorld` to `HelloWorldScene`.  It's just the other changes needed addressing.

PR #1494, opened 2023-12-10
Improve the default templates (at least the cpp template) 

### LINKED ISSUE #1493: Improve the default templates (at least the cpp template)
For a better (cpp) template let us find some improvements:

1) Changing `class HelloWorld `to `class HelloWorldScene`. 
2) Maybe a redesign of the folder structure like on [axmol-game-demo](https://github.com/paulocoutinhox/axmol-game-demo)

Anything missed there? Everybody proposal is welcome.

aismann:

## Describe your changes
see #1493 and diff

#1493


aismann:
@halx99 ,
Merging this PR can be the startpoint for the single template with Lua/CPP

templates/cpp-template-default/Source/MainScene.cpp
```diff
@@ -80,26 +80,26 @@ bool HelloWorld::init()
...
-    //keyboardListener->onKeyPressed  = AX_CALLBACK_2(HelloWorld::onKeyPressed, this);
-    //keyboardListener->onKeyReleased = AX_CALLBACK_2(HelloWorld::onKeyReleased, this);
+    //keyboardListener->onKeyPressed  = AX_CALLBACK_2(MainScene::onKeyPressed, this);
+    //keyboardListener->onKeyReleased = AX_CALLBACK_2(MainScene::onKeyReleased, this);
     //_eventDispatcher->addEventListenerWithFixedPriority(keyboardListener, 11);
 
 
 
-    // add a label shows "Hello World"
+    // add a label shows "HelLo World"
```
paulocoutinhox: typo

templates/cpp-template-default/Source/MainScene.h
```diff
@@ -23,15 +23,15 @@
  THE SOFTWARE.
  ****************************************************************************/
 
-#ifndef __HELLOWORLD_SCENE_H__
-#define __HELLOWORLD_SCENE_H__
+#ifndef __MAINSCENE_H__
```
paulocoutinhox: __MAIN_SCENE_H_

templates/cpp-template-default/Source/MainScene.h
```diff
@@ -64,7 +64,7 @@ class HelloWorld : public ax::Scene
     void menuCloseCallback(Ref* sender);
 
 private:
-    ExampleGameState _gameState = ExampleGameState::init;
+    GameState _gameState = GameState::init;
 };
 
-#endif  // __HELLOWORLD_SCENE_H__
+#endif  // __MAINSCENE_H__
```
paulocoutinhox: _MAIN_SCENE_H

paulocoutinhox:
To me this PR make sense, but the name `CPPTemplate` is horrible to me. You put `CPPTemplate` everywhere. 

Other problem is that changing only this package/bundle dont work, because `org.axmol.hellocpp` is replaced in other places in `axproj-template.json`.

Between `CPPTemplate` and `hellocpp` i prefer `hellocpp`.

Other thing is that you need follow the cases where it is lowecase and uppercase.

aismann:
> To me this PR make sense, but the name `CPPTemplate` is horrible to me. You put `CPPTemplate` everywhere.
> 
> Other problem is that changing only this package/bundle dont work, because `org.axmol.hellocpp` is replaced in other places in `axproj-template.json`.
> 
> Between `CPPTemplate` and `hellocpp` i prefer `hellocpp`.
> 
> Other thing is that you need follow the cases where it is lowecase and uppercase.

Why is 'hellocpp' better?

rh101:
@aismann As @paulocoutinhox mentioned in his post, the "HelloCpp" and "hellocpp" are placeholders that are replaced when a new Axmol project is created.  If you're going to update the name of the project to "CPPTemplate", then you must also change this file to make sure it can look up the correct strings for replacement:

https://github.com/axmolengine/axmol/blob/dev/templates/cpp-template-default/axproj-template.json

paulocoutinhox:
And if you want to change HelloCpp, I suggest you use a better and more definitive name then, like "AxmolCpp". Something that doesn't look like a hello world or template. I think it's strange to have something written "template" in my project right away.

paulocoutinhox:
Also, you need do a simple search in axmol source code, there is other places with bundle 

image

rh101:
> And if you want to change HelloCpp, I suggest you use a better and more definitive name then, like "AxmolCpp". Something that doesn't look like a hello world or template. I think it's strange to have something written "template" in my project right away.

Please avoid using the word "Axmol" in the project name, or anything that needs to be replaced, as it makes it more difficult to search for that string, since so much has "Axmol" in it already.

In fact, there is nothing wrong with HelloCpp at all.  It's a placeholder, and as such, it is always going to be replaced by the developer, so all it needs to be is unique enough to notice, and easy enough to search for.

Also, has anyone ever actually logged an issue related to "HelloCpp", and I don't mean an issue related to personal preference, but a specific concern or technical issue about its usage?  If not, then perhaps putting much focus on changing it is of no real benefit to developers using Axmol.

aismann:
> And if you want to change HelloCpp, I suggest you use a better and more definitive name then, like "AxmolCpp". Something that doesn't look like a hello world or template. I think it's strange to have something written "template" in my project right away.

1) 'CPPTemplate/HelloCpp' will be replaced on` axmol new...`
2) 'CPPTemplate/HelloCpp' are only placeholders and will add cpp-template-default into project(axmol) for tmp test on e.g.: VSStudio.
image

> Also, has anyone ever actually logged an issue related to "HelloCpp", and I don't mean an issue related to personal preference, but a specific concern or technical issue about its usage? If not, then perhaps putting much focus on changing it is of no real benefit to developers using Axmol.

3) My reason to rename 'HelloCpp' is:
image

paulocoutinhox:
Cannot be "CppProject", "MyProject", "MyApp", "CppApp", something more beautiful?

aismann:
> Cannot be "CppProject", "MyProject", "MyApp", "CppApp", something more beautiful?

No. Its only for:
add cpp-template-default into project(axmol) **for tmp test** 
=> NOT for your project (which is mostly wrong used of course)
Hint: Changing anything on 'HellpCpp/HelloLua' had changed also the template itself on your axmol folder

paulocoutinhox:
The problem is that you are changing everything for CPP, you need do like me on latests PR, that is change for BOTH templates.

You changing all for this, but lua template follow the same pattern, ex: HelloLua, org.axmol.hellolua. 

The more you keep messing with it, the harder it is to maintain two templates. I've already spent a lot of time adjusting both projects and now you're going to make them different again.

aismann:
> The more you keep messing with it, the harder it is to maintain two templates. I've already spent a lot of time adjusting both projects and now you're going to make them different again.

Sorry. you should not using axmol which is continues improving ;)
@paulocoutinhox if you familiar with Lua maybe you can write a PR which makes one template for cpp/lua (using this PR as start point)?

rh101:
> Sorry. you should not using axmol which is contininus improving ;)

Improvements to Axmol are great, but, in order for developers to use Axmol, it needs to be relatively stable, in terms of frequency of changes, and the impact that the changes have.  Stating that a developer should not be using Axmol because it's continuously improving is actually a negative, not a positive, and perhaps a little inappropriate too, don't you think?

aismann:
> > Sorry. you should not using axmol which is contininus improving ;)
> 
> Improvements to Axmol are great, but, in order for developers to use Axmol, it needs to be relatively stable, in terms of frequency of changes, and the impact that the changes have. Stating that a developer should not be using Axmol, because it's continuously improving is actually a negative, not a positive, and perhaps a little inappropriate too, don't you think?

No I think not. Everybody can use own projects which was created before whithout an effect there.
A new template is part of the improvement of axmol. A sentence like` I've already spent a lot of time adjusting both projects and now you're going to make them different again` is funny.


> Sorry. you should not using axmol which is contininus improving ;)

Was also mean as a funny sentence of course (look on the smiley)

templates/cpp-template-default/Source/MainScene.h
aismann: done

aismann:
@halx99 
Im finished, Can be merged.

aismann:
Ok. I see neverone understand what 'HelloCpp' realy is:
Its not an one project!
Its always using the sources of cpp-default-template.
So CPPTemplate is more correct.
Btw  changing anything on 'HelloCpp' was always changing the cpp-default-template content not a 'HelloCpp' stuff.
An old Cocos2d-x mistake which will be correct now.
Thats my point of view

halx99:
I perfer HelloCpp, a real project name can be build & test
________________________________

aismann:
HelloCpp lets developer meaning thats an own project  and not the template itself.

HelloCpp should never used as a full project. 
See the comment on CMakeList too:
`add cpp-template-default into project(axmol) for tmp test`

`axmol new...` wil be affected

rh101:
@aismann I'm sure everyone understands that it is just a temporary project name for the template and nothing more.  There is no need to force the issue. 

There are a number of items in this PR that are actual good to change, such as the `HelloWorldScene` change to `MainScene` etc., so best to enable them to be merged in sooner rather than later by limiting this PR to only those useful changes for now.  

The project name change can be given more consideration, and can always be changed in a future PR if the need arises.

templates/cpp-template-default/Source/AppDelegate.cpp
```diff
@@ -34,10 +34,10 @@
...
 USING_NS_AX;
 
-static ax::Size designResolutionSize = ax::Size(1280, 720);
-static ax::Size smallResolutionSize  = ax::Size(480, 320);
-static ax::Size mediumResolutionSize = ax::Size(1024, 768);
-static ax::Size largeResolutionSize  = ax::Size(2048, 1536);
+static Size designResolutionSize = Size(1280, 720);
+static Size smallResolutionSize  = Size(480, 320);
+static Size mediumResolutionSize = Size(1024, 768);
+static Size largeResolutionSize  = Size(2048, 1536);
```
rh101: Please leave these as they were, with the `ax::` prefix.  `Size` is too generic, so that type name may exist in other libraries outside of Axmol.

aismann:
Should be ready for merge now

halx99 (approved):

PR #1501, opened 2023-12-11
[BUG] Fix TMX Camera Culling.

DelinWorks:

## Describe your changes
Fixed incorrect TMX culling calculations for camera zooming causing tiles on top and right of camera to flicker.

### DIFF
--- a/core/2d/FastTMXLayer.cpp
+++ b/core/2d/FastTMXLayer.cpp
@@ -146,8 +146,8 @@ void FastTMXLayer::draw(Renderer* renderer, const Mat4& transform, uint32_t flag
 
         rect.origin.x -= _tileSet->_tileSize.x;
         rect.origin.y -= _tileSet->_tileSize.y;
-        rect.size.x += s.x * (zoom / 2) / 2 + _tileSet->_tileSize.x * zoom;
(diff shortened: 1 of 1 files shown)


paulocoutinhox:
Thanks man :)

aismann:
How can the fix checked?
Which cpp-test can be used for check the fix is working well?

paulocoutinhox:
This solve the effect of don't draw top/right tiles on tmx map with zoom, as you can see here:

https://github.com/axmolengine/axmol/assets/395096/055524ca-d71f-4306-ae92-bdc10e450d45

aismann:
Maybe a cpp-test is usefull too?

DelinWorks:
the new culling function is written by me and I have done a mistake in the equation, it doesn't need a cpp-test since it's a minor bug that can easily be fixed

aismann:
> the new culling function is written by me and I have done a mistake in the equation, it doesn't need a cpp-test since it's a minor bug that can easily be fixed

For the new culling function was never written a test which had found this small bug?
=> So maybe a small test is usefull anyhow. Maybe for future changes.

DelinWorks:
odd thing is I didn't face a problem with it so I disregarded it as working

paulocoutinhox:
I don't think that this kind of thing need a test, since it is only a fix in the latest implementation

PR #1523, opened 2023-12-15
GLES3: Fix scalar array layout

### LINKED ISSUE #1510: Gl es 300 layout std140
- axmol version: dev
- devices test on: windows
- developing environments
   - NDK version: r23c
   - Xcode version: 14.2
   - Visual Studio: 
      - VS version: 2019 (16.11), 2022 (17.6+)
      - MSVC version: 1929, 1934, 19.35, 19.36, 19.37
   - cmake version: 
Steps to Reproduce:

I started to port shaders in my project from gles 1 to
gles 3 and stuck with the similar issue 
https://stackoverflow.com/questions/15616558/regarding-arrays-in-layout-std140-uniform-block-for-opengl
For example _stride_ between elements inside float array will equal sizeof(vec4) if this array declared with std140. 
Sorry, I can't check testcpp right now, but there are some arrays like _pointLightUniformRangeInverseValues which simply pass to the uniform block. Can someone confirm that they are handled properly? 

Ps. I chose to do all arrays with vec4 base type if they intended for shader to minimize calculation on the core side

halx99:

### DIFF
--- a/core/renderer/shaders/base.glsl
+++ b/core/renderer/shaders/base.glsl
@@ -19,3 +19,27 @@
 #define BINORMAL TEXCOORD7
 #define BLENDINDICES COLOR1
 #define BLENDWEIGHT COLOR2
+
+#if !defined(GLES2)
+#  define vfloat_def(x, y) vec4 x[(y + 3) / 4]
+#  define vfloat_at(x, y) x[y / 4][y % 4]
+
+#  define vvec2_def(x, y) vec4 x[(y * 2 + 3) / 4]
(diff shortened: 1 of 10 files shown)


core/renderer/shaders/base.glsl
solan-solan, line 1: I still pay attention to vvec3_at1 implementation. Just if you precise know that it would work on each gpu, since there could happen acces to vec4 unit with indecies more then 3.
Ps. This fix can be checked with test where orc and for point lights. I will do it a little later

halx99: updated

solan-solan (approved):

PR #1565, opened 2024-01-01
Add sub-protocol and optional header support to WebSocket connection

### LINKED ISSUE #1564: WebSocket implementation issues
While testing the experimental WebSocket implementation, I ran into a few issues:

1 - No support for custom header fields. Custom/optional headers are allowed by the RFC; check last few paragraphs of [RFC Section 1.3](https://www.rfc-editor.org/rfc/rfc6455.html#section-1.3)
2 - No support for sub-protocols (via header `Sec-WebSocket-Protocol`), [RFC6455 Section 1.9](https://www.rfc-editor.org/rfc/rfc6455.html#section-1.9)
3 - The current implementation outputs `"WS: PONG frame from server mismatch!"` to the console on unexpected WebSocket PONG messages, but [RFC6455 Section 5.5.3](https://www.rfc-editor.org/rfc/rfc6455.html#section-5.5.3) states: "A Pong frame MAY be sent unsolicited.  This serves as a unidirectional heartbeat.  A response to an unsolicited Pong frame is not expected.".  This PONG should just be ignored by the client.  As an example, the WebSocket implementation in Microsoft .NET sends the PONG for the purpose of keeping the connection alive: https://github.com/dotnet/runtime/blob/25efee0de1c72e40222bbfc7e864fc3f9aa3aa8b/src/libraries/System.Net.WebSockets/src/System/Net/WebSockets/ManagedWebSocket.cs#L586C47-L586C47 

**Describe the solution you'd like**
To support custom headers and sub-protocol, perhaps change the `WebSocket::open` method to:
```
bool WebSocket::open(Delegate* delegate,
                     std::string_view url,
                     std::string_view caFilePath,
                     std::map<std::string, std::string> extraHeaders,
                     std::string protocols)
{
    _delegate   = delegate;
    _url        = url;
    _caFilePath = FileUtils::getInstance()->fullPathForFilename(caFilePath);
    _requestUri = Uri::parse(url);
    _protocols  = std::move(protocols);
    _headers    = std::move(extraHeaders);
// ...
}
```

Then perhaps this change in `void WebSocket::handleNetworkEvent`:
```
void WebSocket::handleNetworkEvent(yasio::io_event* event)
{
// ...
    case YEK_ON_OPEN:
//...
            // response: base64(sha1(key + 258EAFA5-E914-47DA-95CA-C5AB0DC85B11))
            obs.write_bytes("Sec-WebSocket-Key: ");
            obs.write_bytes(_handshakeSecKey);
            obs.write_bytes("\r\n");

            if (!_protocols.empty())
            {
                obs.write_bytes("Sec-WebSocket-Protocol: ");
                obs.write_bytes(_protocols);
                obs.write_bytes("\r\n");
            }

            for (auto&& header : _headers)
            {
                obs.write_bytes(fmt::format("{0}: {1}\r\n", header.first, header.second));
            }

            obs.write_bytes("User-Agent: yasio-ws\r\n");
// ...
```

Also, the line `AXLOG("WS: PONG frame from server mismatch!\n");` should either be removed, or the text changed:
```
int WebSocket::on_frame_end(websocket_parser* parser)
{
// ...
        case WS_OP_PING:
            AXLOG("WS: control frame: PING");
            WebSocketProtocol::sendFrame(*ws, ws->_receivedData.data(), ws->_receivedData.size(),
                                         ws::detail::opcode::pong);
            break;
        case WS_OP_PONG:
            AXLOG("WS: control frame: PONG");
            if (ws->_receivedData.size() != 4 || 0 != memcmp(ws->_receivedData.data(), "WSWS", 4))
                AXLOG("WS: PONG frame from server mismatch!\n");
            break;
        }
    }

    return 0;
}
```

Perhaps changing the text to `AXLOG("WS: Unsolicited PONG frame from server (possible keep-alive)\n");` would be more appropriate, to avoid confusing anyone seeing the "mismatch" and assuming something is wrong.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

@halx99 Do you want me to create a PR with the changes based on the suggestions above, or will you be taking care of this?

Other than the items listed above, the WebSocket implementation seems to be working well with the current server my app connects to (based on MS ,NET 8.0).

rh101:

## Describe your changes
Add sub-protocol and optional header support to WebSocket connection.
Unsolicited WebSocket PONG messages received should be ignored.

#1564 

### For each PR
       
   Optional:
   -  [ ] I have checked readme and add important infos to this PR.
          
### For core/new feature
-  [ ] I have checked readme and add important infos to this PR.
-  [ ] I have added thorough tests.

### DIFF
--- a/core/network/WebSocket-wasm.cpp
+++ b/core/network/WebSocket-wasm.cpp
@@ -54,7 +54,7 @@ WebSocket::WebSocket() {}
 
 WebSocket::~WebSocket() {}
 
-bool WebSocket::open(Delegate* delegate, std::string_view url, std::string_view caFilePath, const char* protocols)
+bool WebSocket::open(Delegate* delegate, std::string_view url, std::string_view caFilePath, std::string_view protocols)
 {
     if (url.empty())
     {
@@ -64,7 +64,9 @@ bool WebSocket::open(Delegate* delegate, std::string_view url, std::string_view 
 
     _delegate = delegate;
 
-    EmscriptenWebSocketCreateAttributes ws_attrs = {url.data(), protocols, EM_TRUE};
+    EmscriptenWebSocketCreateAttributes ws_attrs = {
+            url.data(), protocols.empty() ? nullptr : std::string(protocols).c_str(),
+                                                        EM_TRUE};
 
     AXLOG("ws open url: %s, protocols: %s", ws_attrs.url, ws_attrs.protocols);
 
--- a/core/network/WebSocket-wasm.cpp
+++ b/core/network/WebSocket-wasm.cpp
@@ -64,9 +64,11 @@ bool WebSocket::open(Delegate* delegate, std::string_view url, std::string_view 
 
     _delegate = delegate;
 
-    EmscriptenWebSocketCreateAttributes ws_attrs = {
-            url.data(), protocols.empty() ? nullptr : std::string(protocols).c_str(),
-                                                        EM_TRUE};
+    _url          = url;
+    _subProtocols = protocols;
+
+    EmscriptenWebSocketCreateAttributes ws_attrs = {_url.c_str(),
+                                                    _subProtocols.empty() ? nullptr : _subProtocols.c_str(), EM_TRUE};
 
     AXLOG("ws open url: %s, protocols: %s", ws_attrs.url, ws_attrs.protocols);
 
--- a/core/network/WebSocket-wasm.h
+++ b/core/network/WebSocket-wasm.h
@@ -12,8 +12,8 @@ NS_AX_BEGIN
 namespace network
 {
 /**
- * WebSocket is wrapper of the libwebsockets-protocol, let the develop could call the websocket easily.
- * Please note that all public methods of WebSocket have to be invoked on Cocos Thread.
+ * WebSocket implementation using yasio.
+ * Please note that all public methods of WebSocket have to be invoked on Axmol Thread.
  */
 class AX_DLL WebSocket
 {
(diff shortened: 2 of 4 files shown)


halx99:
1. wasm websocket impl constructor prototype should same with native impl
2. move extra headers set to member function, just like HttpRequest setHeaders
3. use `std::vector<std::string>` store custom headers, same with HttpRequest

rh101:
> 1. wasm websocket impl constructor prototype should same with native impl
> 
>     2. move extra headers set to member function, just like HttpRequest setHeaders
> 
>     3. use `std::vector<std::string>` store custom headers, same with HttpRequest

Will do.

rh101:
Emscripten WebSocket API has no support for custom headers, so the `setHeaders`/`getHeaders` doesn't apply to the WASM WebSocket implementation.

NOTE: For WASM WebSocket, the work-around to the lack of support of custom headers in the initial HTTP connection upgrade request is to pass the data as HTTP parameters instead.

core/network/WebSocket-wasm.cpp
```diff
@@ -64,7 +64,9 @@ bool WebSocket::open(Delegate* delegate, std::string_view url, std::string_view
 
     _delegate = delegate;
 
-    EmscriptenWebSocketCreateAttributes ws_attrs = {url.data(), protocols, EM_TRUE};
+    EmscriptenWebSocketCreateAttributes ws_attrs = {
+            url.data(), protocols.empty() ? nullptr : std::string(protocols).c_str(),
+                                                        EM_TRUE};
```
halx99: temp variable will be invalid when it used by new wasm websocket

rh101: That should work given the documentation states the following (checked it last night before implementing it):
` // This string can point to a stack local variable, the string is read immediately at a call to emscripten_websocket_new().`

https://github.com/emscripten-core/emscripten/blob/85c099ab519d574c3ce1f6d3603eb74aabbdf604/system/include/emscripten/websocket.h#L83

Isn't that correct? If it's not, then I'll change it.

halx99:
yes, it's incorrect

rh101:
> yes, it's incorrect

Silly me, of course it's incorrect. I'll fix it now.

Creating a local `std::string subProtocols = protocols;` then passing `subProtocols.c_str()` should work, but it may as well be stored as a member of that object, since the url seems to be too (although `_url` wasn't being used for some reason).

halx99 (approved):

PR #1580, opened 2024-01-04
Fix shader cache leak

j-jorge:
The ShaderCache class had two static members: a map for the cache, and
a pointer to the single instance of ShaderCache. The map was
supposedly cleared in the destructor but there was actually no
instance of ShaderCache (getInstance() was never called and the other
functions accessed the map directly). This commit removes the
singleton "interface" such that only the map is used, then it cleared
the map upon request.

### DIFF
--- a/core/renderer/Texture2D.cpp
+++ b/core/renderer/Texture2D.cpp
@@ -82,7 +82,6 @@ Texture2D::~Texture2D()
 #if AX_ENABLE_CACHE_TEXTURE_DATA
     VolatileTextureMgr::removeTexture(this);
 #endif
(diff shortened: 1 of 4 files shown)


halx99:
what about convert `_cachedShaders` to non-static member?

j-jorge:
@halx99 A non-static member and using the class via the getInstance() function? It would have been more consistent with the other classes indeed :/

halx99:
> @halx99 A non-static member and using the class via the getInstance() function? It would have been more consistent with the other classes indeed :/

Yes

PR #1584, opened 2024-01-05
templates: cpp: AppDelegate: Show stats only on debug mode

Roker2:
## Changes
I added code which exclude stats from screen on Release mode

## Checklist

### DIFF
--- a/templates/cpp-template-default/Source/AppDelegate.cpp
+++ b/templates/cpp-template-default/Source/AppDelegate.cpp
@@ -70,8 +70,10 @@ bool AppDelegate::applicationDidFinishLaunching()
         director->setGLView(glView);
     }
 
+#if _AX_DEBUG > 0
     // turn on display FPS
     director->setStatsDisplay(true);
+#endif
 
     // set FPS. the default value is 1.0/60 if you don't call this
     director->setAnimationInterval(1.0f / 60);


rh101:
We have to be careful when making changes like this to the template. The template should remain as simple as possible, and adding pre-processor checks like this in the template just adds "noise" to the code.

The template is meant as a starting point, and that file will be edited during the development process.  The developer of a project can easily add or delete any code they want to it, such as adding their own pre-processor checks if required.

Roker2:
> We have to be careful when making changes like this to the template. The template should remain as simple as possible, and adding pre-processor checks like this in the template just adds "noise" to the code.
> 
> The template is meant as a starting point, and that file will be edited during the development process. The developer of a project can easily add or delete any code they want to it, such as adding their own pre-processor checks if required.

But this code can show that "Hey, you know, we have this thing and you can use it!". If it is not okay for you, you can close this PR.

rh101:
> But this code can show that "Hey, you know, we have this thing and you can use it!". If it is not okay for you, you can close this PR.

Oh sorry, it's not about it being okay or not; it was just my opinion regarding changes like that.  

Also, the stats display can be used in release builds to monitor performance, which is why it was made to receive a boolean to turn it on or off. so it's not explicitly limited to debug mode.

Roker2:
> > But this code can show that "Hey, you know, we have this thing and you can use it!". If it is not okay for you, you can close this PR.
> 
> Oh sorry, it's not about it being okay or not; it was just my opinion regarding changes like that.
> 
> Also, the stats display can be used in release builds to monitor performance, which is why it was made to receive a boolean to turn it on or off. so it's not explicitly limited to debug mode.

Hm. It have a sense. But I thought adding this `#if` will be useful... Okay, I think we can close this PR.

halx99:
Thanks, close as @rh101  said

PR #1610, opened 2024-01-12
Allow to override AX_USE_COMPAT_GL from user's CMakeLists.txt

smilediver:

## Describe your changes

This change enables users to override `AX_USE_COMPAT_GL` option from user's `CMakeLists.txt`. We need to be able to force the use of OpenGL on macOS, because we have shader graph editor, that generates and previews shaders in runtime.

Current fix is probably not optimal, so I'm willing to change it if you can suggest a better way to implement it.

N/A

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/cmake/Modules/AXPlatform.cmake
+++ b/cmake/Modules/AXPlatform.cmake
@@ -56,6 +56,10 @@ else()
     return()
 endif()
 
+if (NOT DEFINED WIN32)
+    set(WIN32 FALSE)
+endif()
+
 # generators that are capable of organizing into a hierarchy of folders
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 # simplify generator condition, please use them everywhere
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -51,9 +51,7 @@ endif()
 include(CMakeDependentOption)
 
 # by default: use angle on win32
-if(NOT DEFINED AX_USE_COMPAT_GL)
-    cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" ON "WIN32" OFF)
-endif()
+cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" ${WIN32} "WIN32 OR APPLE" OFF)
 
 # choosing render backend for all target platforms: win32, winuwp, ios, tvos, android, linux, osx
 if(APPLE AND (NOT AX_USE_COMPAT_GL))
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -51,7 +51,9 @@ endif()
 include(CMakeDependentOption)
 
 # by default: use angle on win32
-cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" ON "WIN32" OFF)
+if(NOT DEFINED AX_USE_COMPAT_GL)
+    cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" ON "WIN32" OFF)
+endif()
 
 # choosing render backend for all target platforms: win32, winuwp, ios, tvos, android, linux, osx
 if(APPLE AND (NOT AX_USE_COMPAT_GL))


halx99:
the original code is optional variable, so you can set it to TRUE as a cache variable at your cmake file

smilediver:
I've tried overriding the value in project's CMakeLists.txt file with: `set(AX_USE_COMPAT_GL ON)` and `set(AX_USE_COMPAT_GL ON CACHE BOOL "enable AX_USE_COMPAT_GL" FORCE)`, but `cmake_dependent_option()` call always overrides the value (I've printed it before and after the call).

P.S. I'm testing on macOS.

rh101:
> I've tried overriding the value in project's CMakeLists.txt file with: `set(AX_USE_COMPAT_GL ON)` and `set(AX_USE_COMPAT_GL ON CACHE BOOL "enable AX_USE_COMPAT_GL" FORCE)`, but `cmake_dependent_option()` call always overrides the value (I've printed it before and after the call).

`cmake_dependent_option` won't allow the option to be set by the user unless all the items in the dependency list are true, which in this case is the requirement to have "WIN32" set.  No matter what you do, it won't work in it's current form.

Perhaps change that entire block to something like this may work:
```
if(WIN32)
    option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" ON)
elseif(MACOSX)
    option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" OFF)
else()
    set(AX_USE_COMPAT_GL OFF CACHE BOOL "Whether use compatibility GL as render backend" FORCE)
endif()
```

smilediver:
Yeah, it's probably a better solution. If everyone agrees, I can change it to that.

halx99:
does double OFF works?

rh101:
> does double OFF works?

Sorry, I'm not sure what you mean.

halx99:
> > does double OFF works?
> 
> Sorry, I'm not sure what you mean.

discard me

halx99:
I have test on windows, it's works from command line `cmake -B buildtest -DAX_USE_COMPAT_GL=ON` and cmake:
```cmake
cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" ON "MACOS" OFF)
```

rh101:
> I have test on windows, it's works from command line `cmake -B buildtest -DAX_USE_COMPAT_GL=ON` and cmake:
> 
> ```cmake
> cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" ON "MACOS" OFF)
> ```

Shouldn't it be "OFF" by default for MACOSX?

`cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" OFF "MACOS" OFF)`

Also, do you mean add both lines in `core/CMakeLists.txt`, one for WIN32 and another for MACOSX?

```
cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" ON "WIN32" OFF)
cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" OFF "MACOS" OFF)`
```
Or did you mean something else?

halx99:
sorry my mistake: the value in CMakeCache.txt is correct, and type is `:INTERNAL`, but in cmake config step incorrect

halx99:
> WIN32

maybe a mistake use of `cmake_dependent_option`: https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

halx99:
Better solution is:

1. AXPlatform.cmake insert follow code at line:59
```cmake
if (NOT DEFINED WIN32)
    set(WIN32 FALSE)
endif()
```
3. `cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" ${WIN32} "WIN32 OR APPLE" OFF)`

rh101:
> 3\. cmake_dependent_option(AX_USE_COMPAT_GL "Whether use compatibility GL as render backend" ${WIN32} "WIN32 OR APPLE" OFF)

Oh! Yes that's a good solution.  I didn't notice that you could use conditional operators (the `OR`) inside the dependency section.

smilediver:
Pushed the latest solution that was discussed.

halx99 (approved):

PR #1612, opened 2024-01-13
Remove unused var io.BackendRendererUserData

Samoongeer:


extensions/ImGui/imgui_impl_ax.cpp
```diff
@@ -1546,7 +1546,6 @@ void ImGui_ImplAx_Init()
 {
     auto bd                    = ImGui_ImplGlfw_GetBackendData();
     auto& io                   = ImGui::GetIO();
-    io.BackendRendererUserData = (void*)bd;
```
rh101, line 1549: That is not an unused variable.  Can you please explain the reason for wanting to remove this line?

halx99: axmol implementation share GLFW backend platform user data, so yes, it's unused, and not like https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_opengl3.cpp#L242

rh101: I understand, and thanks for clarifying that.

PR #1620, opened 2024-01-19
Enable default aidl file to project especially for In-app purchases

armanhossiny:

- `1.x`: 1.x BugFixs

### For each PR
       
   Optional:
          
### For core/new feature PR


halx99 (approved):

rh101:
Isn't this functionality deprecated in Android?

halx99:
Not sure, there is a stackoverflow thread: https://stackoverflow.com/questions/76232982/how-to-suppress-warning-it-will-be-removed-in-version-9-0-of-the-android-gradle

rh101:
image

From here: https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes#default-changes

This is really a project-specific option, and I personally feel that it is not a good idea to add it to the template.  If a developer requires it, then that developer can add it, but it's not something that should be forced on all developers creating new projects.

halx99:
WARNING: The option setting 'android.defaults.buildfeatures.buildconfig=true' is deprecated.

Got this warning message: https://github.com/simdsoft/axmol/actions/runs/7584089520/job/20657048877

halx99:
The docs may means the props in gradle.properties is deprecated, instead define per build.gradle DSL namespace like:
```txt
android {
   buildFeatures {
       buildconfog = true
   }
}
```

rh101:
> The docs may means the props in gradle.properties is deprecated, instead define per build.gradle DSL namespace like:

I didn't realise it's still supported via build.gradle. All good!

PR #1624, opened 2024-01-20
Fix use UserDefault

git-nujen:

- `1.x`: 1.x BugFixs

### For each PR
       
   Optional:
          
### For core/new feature PR


halx99:
The setFileName of UserDefault is a static function, so don't need invoke getInstance, and create instance of UserDefault depends a valid FileName, so don't do that

PR #1690, opened 2024-02-16
Remove the Enhance API.

j-jorge:
The Samsung Enhance API has been added in Cocos2D-x in 2015 to allow Samsung to tweak the games for their devices. It's a set of functions for which we do not have the implementation, consequently we can't maintain it.

This PR removes this API.


halx99:
I'm not sure do we need remove Enhance API. some questions:
- it's unused?
- it's break changes?

j-jorge:
I can't tell if it is used or not because it was a closed-source module. For what it's worth it seems to have been removed from Cocos Engine. It adds nothing to Axmol except code to maintain, and we can't even tell if it works, so IMHO it should go :)

halx99 (approved):

PR #1699, opened 2024-02-18
 Sequence::create(FiniteTimeAction* action1, ...)   => Add informal text : ATTENTION! Last entry have to be the 'nullptr'. 

aismann:
## Describe your changes
 Sequence::create(FiniteTimeAction* action1, ...)   => Add informal text : ATTENTION! Last entry have to be the 'nullptr'. 

see also Discord:
https://discordapp.com/channels/1099599084895088670/1183301042511806476/1208418963718152232

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/cmake/Modules/AXLinkHelpers.cmake
+++ b/cmake/Modules/AXLinkHelpers.cmake
@@ -282,7 +282,7 @@ function(ax_link_cxx_prebuilt APP_NAME AX_ROOT_DIR AX_PREBUILT_DIR)
             $<TARGET_FILE_DIR:${APP_NAME}>)
 
         if (BUILD_SHARED_LIBS)
-            add_custom_command(TARGET ${ax_target} POST_BUILD
+            add_custom_command(TARGET ${APP_NAME} POST_BUILD
                 COMMAND ${CMAKE_COMMAND} -E copy_if_different
                 "${AX_ROOT_DIR}/${AX_PREBUILT_DIR}/bin/${BUILD_CONFIG_DIR}glad.dll"
                 "${AX_ROOT_DIR}/${AX_PREBUILT_DIR}/bin/${BUILD_CONFIG_DIR}glfw.dll"


rh101:
@aismann That is a little misleading, since the `nullptr` requirement is only for this method:
`static Sequence* createWithVariableList(FiniteTimeAction* action1, va_list args);`

That comment should be added to that method, not to the whole class. Perhaps like this:

```
  /** Helper constructor to create an array of sequence-able actions.
   *
   * @param action1 The first sequenceable action.
   * @param args The va_list variable.
   * @return An autoreleased Sequence object.
   * @js NA
   * @note ATTENTION! Last entry must be a 'nullptr' to correctly terminate the list.
   */
  static Sequence* createWithVariableList(FiniteTimeAction* action1, va_list args);
```

aismann:
> @aismann That is a little misleading, since the `nullptr` requirement is only for this method: `static Sequence* createWithVariableList(FiniteTimeAction* action1, va_list args);`
> 
> That comment should be added to that method, not to the whole class. Perhaps like this:
> 
> ```
>   /** Helper constructor to create an array of sequence-able actions.
>    *
>    * @param action1 The first sequenceable action.
>    * @param args The va_list variable.
>    * @return An autoreleased Sequence object.
>    * @js NA
>    * @note ATTENTION! Last entry must be a 'nullptr' to correctly terminate the list.
>    */
>   static Sequence* createWithVariableList(FiniteTimeAction* action1, va_list args);
> ```

@rh101 

`Sequence* Sequence::create(FiniteTimeAction* action1, ...)`  calls the `Sequence::createWithVariableList(action1, params);`
So I think at least this both methods need the info text. Right?

rh101:
> `Sequence* Sequence::create(FiniteTimeAction* action1, ...)` calls the `Sequence::createWithVariableList(action1, params);`
> So I think at least this both methods need the info text. Right?

Sorry, yes, you are correct, both of those methods are the ones that needs the comment.

PR #1728, opened 2024-03-02
set(CMAKE_UNITY_BUILD ON)

crazyhappygame:
### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,8 @@
 cmake_minimum_required(VERSION 3.22)
 
 project(axmol)
+# Enable Unity Build globally
+set(CMAKE_UNITY_BUILD ON)
 
 if(XCODE)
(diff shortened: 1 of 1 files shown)


rh101:
What is the reason for this PR?

The unity build setting should not be set in the engine cmake files, since developers using Axmol would no longer have control over it.

EDIT: Even the official documentation states the same https://cmake.org/cmake/help/latest/variable/CMAKE_UNITY_BUILD.html

"Projects should not set this variable, it is intended as a developer control to be set on the [cmake(1)](https://cmake.org/cmake/help/latest/manual/cmake.1.html#manual:cmake(1)) command line or other equivalent methods. The developer must have the ability to enable or disable unity builds according to the capabilities of their own machine and compiler."

rh101:
Given that the suggested change should not be merged in, is there any reason to keep this PR open?

crazyhappygame:
thanks for closing 

I would like to check if unity build will work out of the box. Unfortunately many changes are required…
Something to consider in the future.

PR #1764, opened 2024-03-25
Add support for verbose trace level logging

rh101:
## Describe your changes

The most verbose log level at the moment is `LogLevel::Debug`, and while this serves its purpose, it is not suitable for trace-level logging, as it would end up with too much log spam, along with the possibility of revealing information that otherwise should not be revealed in a release/production version of an application.

This PR adds `LogLevel::Trace`, which is specifically for the purpose of verbose logging.

@halx99 If you accept this PR, do you want a color associated with the trace messages? If so, which color should be used for this?

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/base/Console.cpp
+++ b/core/base/Console.cpp
@@ -118,6 +118,9 @@ AX_API LogItem& preprocessLog(LogItem&& item)
             std::string_view levelName;
             switch (item.level_)
             {
+            case LogLevel::Trace:
+                levelName = "T/"sv;
+                break;
             case LogLevel::Debug:
                 levelName = "D/"sv;
                 break;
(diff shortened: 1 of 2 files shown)


halx99 (approved):

halx99:
about color, I think we can refer to spdlog.

rh101:
> about color, I think we can refer to spdlog.

They seem to use white foreground color for the trace level `\x1b[37m`, which seems to be common for trace messages.  I'll set it to that unless there are any objections.

Spdlog also uses cyan `\x1b[36m` for debug level. Should that be used too?

rh101:
`Debug` level output is set to cyan, and `Trace` level output is set to white, both of which seem to be used in other apps (spdlog and Android etc.)

PR #1799, opened 2024-04-08
Fix crash on iOS

smilediver:
## Describe your changes

This fixes the following crash on iOS: https://stackoverflow.com/questions/55361057/this-copy-of-libswiftcore-dylib-requires-an-os-version-prior-to-12-2-0

### For each PR
       
   Optional:
          
### For core/new feature PR


halx99:
why need /usr/lib/swift, axmol not use swift

smilediver:
We use CocoaPods and some pods use Swift, for example Firebase. "/usr/lib/swift" is added by the pods xcconfig file to fix this issue, but `CMAKE_INSTALL_RPATH` set by Axmol sets this on the app target and inserts the "@executable_path/Frameworks" before the pod xcconfig overriding the value.

Screenshot 2024-04-08 at 17 25 34

PR #1808, opened 2024-04-09
Fix same uniforms in fragment and vertex shaders not working in Metal

smilediver:
## Describe your changes

Context: in OpenGL API setting shader uniforms works at the program level, which means that same named uniforms in vertex and fragment shaders are treated as the same variable, so when you set a uniform it is set in both shaders at once. In Metal API uniforms are set for vertex and fragment shaders separately.

In current Axmol's Metal implementation if you have a same named uniform in vertex and fragment shaders, then only the vertex shader's variable is set, which is different from OpenGL. This PR fixes Metal part to match OpenGL semantics.

This PR may have breaking changes as it modifies `UniformLocation` and `ShaderModuleMTL::getUniformLocation()`. But probably it's not a common thing to access `UniformLocation` internals or calling `ShaderModuleMTL::getUniformLocation()`.

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/renderer/backend/ProgramState.cpp
+++ b/core/renderer/backend/ProgramState.cpp
@@ -230,16 +230,10 @@ void ProgramState::setCallbackUniform(const backend::UniformLocation& uniformLoc
 
 void ProgramState::setUniform(const backend::UniformLocation& uniformLocation, const void* data, std::size_t size)
 {
-    switch (uniformLocation.shaderStage)
-    {
-    case backend::ShaderStage::VERTEX:
-        setVertexUniform(uniformLocation.location[0], data, size, uniformLocation.location[1]);
-        break;
-    case backend::ShaderStage::FRAGMENT:
-        setFragmentUniform(uniformLocation.location[0], data, size, uniformLocation.location[1]);
-        break;
-    default:;
-    }
+    if (uniformLocation.vertStage)
+        setVertexUniform(uniformLocation.vertStage.location, data, size, uniformLocation.vertStage.offset);
+    if (uniformLocation.fragStage)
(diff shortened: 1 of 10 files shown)


halx99:
Please fix the ci check: https://github.com/axmolengine/axmol/actions/runs/8620021129/job/23625989923?pr=1808#step:4:3879

smilediver:
Pushed fix. Also noticed a uniform type mismatch in positionNormalTexture.vert and colorNormalTexture.frag, so fixed that too.

core/renderer/backend/ProgramState.cpp
```diff
@@ -230,16 +230,10 @@ void ProgramState::setCallbackUniform(const backend::UniformLocation& uniformLoc
...
-        setVertexUniform(uniformLocation.location[0], data, size, uniformLocation.location[1]);
-        break;
-    case backend::ShaderStage::FRAGMENT:
-        setFragmentUniform(uniformLocation.location[0], data, size, uniformLocation.location[1]);
-        break;
-    default:;
-    }
+    if (uniformLocation.vertStage)
+        setVertexUniform(uniformLocation.vertStage.location, data, size, uniformLocation.vertStage.offset);
+    if (uniformLocation.fragStage)
```
halx99, line 236: Add preprocessor:
```cmake
#ifdef AX_USE_METAL
if (uniformLocation.fragStage)
        setVertexUniform(uniformLocation.vertStage.location, data, size, uniformLocation.vertStage.offset);
#endif
```

And the function: `setFragmentUniform`  can be also in preprocessor: #ifdef AX_USE_METAL

halx99 (approved):

halx99:
cheers

PR #1830, opened 2024-04-17
Fixed an issue where ProgressTimer didn't properly set or cascade its opacity

TyelorD:
## Describe your changes
This fixes a long standing bug from the Cocos2d-x  3.x days that I fixed in our source code for our games with it. I never made a pull request to Cocos2d-x since they stopped development and maintenance, but thought it might be useful for Axmol!

N/A

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/2d/ProgressTimer.cpp
+++ b/core/2d/ProgressTimer.cpp
@@ -231,8 +231,21 @@ void ProgressTimer::updateDisplayedOpacity(uint8_t parentOpacity)
 {
     Node::updateDisplayedOpacity(parentOpacity);
 
-    _sprite->updateDisplayedOpacity(_displayedOpacity);
+    _displayedOpacity = _realOpacity * parentOpacity / 0xFF;
+    
+    _sprite->setOpacity(_displayedOpacity);
+    updateColor();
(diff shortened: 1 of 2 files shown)


core/2d/ProgressTimer.cpp
```diff
@@ -231,8 +231,21 @@ void ProgressTimer::updateDisplayedOpacity(uint8_t parentOpacity)
...
-    _sprite->updateDisplayedOpacity(_displayedOpacity);
+    _displayedOpacity = _realOpacity * parentOpacity / 0xFF;
+    
+    _sprite->setOpacity(_displayedOpacity);
+    updateColor();
     updateProgress();
+    
+    if (_cascadeOpacityEnabled)
+    {
+        _sprite->updateDisplayedOpacity(_displayedOpacity);
```
smilediver, line 242: I may be wrong, but since `_sprite` is considered part of `ProgressTimer` widget, then shouldn't `_sprite->updateDisplayedOpacity(_displayedOpacity);` be always called? As I understand `_cascadeOpacityEnabled` part should only apply for the children.

PR #1844, opened 2024-04-22
When loading csb files, prevent repeated loading of plist files

tkzcfc:
### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/2d/SpriteFrameCache.cpp
+++ b/core/2d/SpriteFrameCache.cpp
@@ -172,7 +172,7 @@ void SpriteFrameCache::removeUnusedSpriteFrames()
 
 void SpriteFrameCache::removeUnusedSpriteSheets()
 {
-    std::vector<string> willRemoveSpriteSheetFileNames;
+    std::vector<std::string_view> willRemoveSpriteSheetFileNames;
     for (auto&& it : _spriteSheets)
     {
         bool isUsed = false;
--- a/core/2d/SpriteFrameCache.cpp
+++ b/core/2d/SpriteFrameCache.cpp
@@ -170,6 +170,33 @@ void SpriteFrameCache::removeUnusedSpriteFrames()
     }
 }
 
+void SpriteFrameCache::removeUnusedSpriteSheetFile()
+{
+    std::vector<string> willRemoveSpriteSheetFileNames;
+    for (auto&& it : _spriteSheets)
+    {
+        bool isUsed = false;
+        for (auto&& frame : it.second->frames)
+        {
+            auto spriteFrame = getSpriteFrameByName(frame);
+            if (spriteFrame && spriteFrame->getReferenceCount() > 1)
+            {
+                isUsed = true;
+                break;
+            }
+        }
+
+        if (!isUsed)
+            willRemoveSpriteSheetFileNames.push_back(it.first);
+    }
+
+    for (auto& spriteSheetFileName : willRemoveSpriteSheetFileNames)
+    {
+        AXLOG("axmol: SpriteFrameCache: removing unused sprite sheet file : %s", spriteSheetFileName.c_str());
+        removeSpriteSheet(spriteSheetFileName);
+    }
+}
+
 void SpriteFrameCache::removeSpriteFrameByName(std::string_view name)
 {
     // explicit nil handling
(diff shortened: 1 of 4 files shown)


core/2d/SpriteFrameCache.cpp
```diff
@@ -170,6 +170,33 @@ void SpriteFrameCache::removeUnusedSpriteFrames()
     }
 }
 
+void SpriteFrameCache::removeUnusedSpriteSheetFile()
```
rh101: This method name, `removeUnusedSpriteSheetFile`, is a little misleading, and perhaps it's better to call it `removeUnusedSpriteSheets`, since that is exactly what it does.

tkzcfc: modified

rh101:
Is there any reason that you can't use `SpriteFrameCache::isSpriteFramesWithFileLoaded` to check if a sprite sheet is already loaded?

For instance, you've added `_loadedPlists`, and use it like this:

```
if (!_loadedPlists.contains(plist + png))
{
    // ...
}
```

Does the following not work?

```
if (!SpriteFrameCache::getInstance()->isSpriteFramesWithFileLoaded(plist + png))
{
    // ...
}
```

tkzcfc:
removeUnusedSpriteFrames may release unused sprite frames

tkzcfc:
We need to reload

rh101:
> removeUnusedSpriteFrames may release unused sprite frames

Sorry, I'm not quite sure that I understand what you mean.

`isSpriteFramesWithFileLoaded` is implemented as follows:
```
bool SpriteFrameCache::isSpriteFramesWithFileLoaded(std::string_view plist) const
{
    return isSpriteSheetInUse(plist) && isPlistFull(plist);
}
```

It checks if the sprite sheet entry is full via `isPlistFull`, so if any frames were removed through `removeUnusedSpriteFrames`, then `isPlistFull` would return `false`, which means `isSpriteFramesWithFileLoaded` returns false.  Isn't that the case?

tkzcfc:
ok, so we don’t even need the _loadedPlists variable?

rh101:
> ok, so we don’t even need the _loadedPlists variable?

Sorry, yes, that is what I meant.  Have you checked if `SpriteFrameCache::getInstance()->isSpriteFramesWithFileLoaded` works for you? If it does work, then you do not need `_loadedPlists`.

tkzcfc:
Let me try

rh101 (approved):

rh101:
@halx99 I can't quite figure out how to set the unresolved conversation as resolved. I thought it was the "review changes", but that wasn't it, so please ignore the review approval listed above.

extensions/cocostudio/src/cocostudio/ActionTimeline/CSLoader.cpp
```diff
@@ -1438,7 +1447,11 @@ Node* CSLoader::createNodeWithFlatBuffersForSimulator(std::string_view filename)
     //    AXLOG("textureSize = %d", textureSize);
     for (int i = 0; i < textureSize; ++i)
     {
-        SpriteFrameCache::getInstance()->addSpriteFramesWithFile(textures->Get(i)->c_str());
+        std::string plist = textures->Get(i)->c_str();
```
halx99:
```suggestion
        std::string_view plist = textures->Get(i)->c_str();
```

extensions/cocostudio/src/cocostudio/ActionTimeline/CSLoader.cpp
```diff
@@ -1074,7 +1080,10 @@ Node* CSLoader::nodeWithFlatBuffersFile(std::string_view fileName, const ccNodeL
     for (int i = 0; i < textureSize; ++i)
     {
         std::string plist = textures->Get(i)->c_str();
```
halx99:
```suggestion
        std::string_view plist = textures->Get(i)->c_str();
```

extensions/cocostudio/src/cocostudio/ActionTimeline/CSLoader.cpp
```diff
@@ -950,7 +953,10 @@ Node* CSLoader::createNode(const Data& data, const ccNodeLoadCallback& callback)
         for (int i = 0; i < textureSize; ++i)
         {
             std::string plist = textures->Get(i)->c_str();
```
halx99:
```suggestion
            std::string_view plist = textures->Get(i)->c_str();
```

core/2d/SpriteFrameCache.h
```diff
@@ -246,6 +248,8 @@ class AX_DLL SpriteFrameCache
 
     SpriteFrame* findFrame(std::string_view frame);
 
+    std::string getSpriteFrameName(SpriteFrame* frame);
```
halx99:
```suggestion
    std::string_view getSpriteFrameName(SpriteFrame* frame);
```

core/2d/SpriteFrameCache.cpp
```diff
@@ -376,6 +403,18 @@ SpriteFrame* SpriteFrameCache::findFrame(std::string_view frame)
     return _spriteFrames.at(frame);
 }
 
+std::string SpriteFrameCache::getSpriteFrameName(SpriteFrame* frame)
```
halx99:
```suggestion
std::string_view SpriteFrameCache::getSpriteFrameName(SpriteFrame* frame)
```

halx99 (changes requested):

halx99 (approved):

PR #1853, opened 2024-04-26
Cleanup cc stubs: ccMenuCallback => menuCallback

aismann:
## Describe your changes

Rename ax::ccMenuCallback to ax::menuCallback

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/2d/MenuItem.cpp
+++ b/core/2d/MenuItem.cpp
@@ -4,6 +4,7 @@ Copyright (c) 2010-2012 cocos2d-x.org
 Copyright (c) 2011      Zynga Inc.
 Copyright (c) 2013-2016 Chukong Technologies Inc.
 Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
+Copyright (c) 2019-present Axmol Engine contributors (see AUTHORS.md).
 
 https://axmolengine.github.io/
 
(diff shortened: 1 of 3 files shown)


core/2d/MenuItem.cpp
```diff
@@ -148,7 +149,7 @@ void MenuItemLabel::setLabel(Node* var)
     _label = var;
 }
 
-MenuItemLabel* MenuItemLabel::create(Node* label, const ccMenuCallback& callback)
+MenuItemLabel* MenuItemLabel::create(Node* label, const menuCallback& callback)
```
halx99, line 152: the type shoud starts with uppercase

TyelorD: Even though menuCallback is just a typedef alias for `std::function<void(Ref*)>`?

halx99: maybe use common callback is better

halx99: maybe: InputCallback

aismann: There be 3 same typedef
image
What is your proposal?

smilediver: I vote for just getting rid of `cc` prefix, ie leave `MenuCallback`, `WidgetClickCallback`, etc.

aismann:
I closed it. need a deeper look

PR #1859, opened 2024-04-27
Add ability to create console apps

smilediver:
This PR has two changes:

* It adds `CONSOLE` option to the `ax_setup_app_config()` function to allow creating console apps that use Axmol.
* And a small change to `1k/build.ps1` to print out what CMake command is run when building.


halx99:
entrypoint will missing when build as console app

smilediver:
I'm aware of this, but I think if someone will use this option, then he knows what he is doing and it will be an easy fix. And I don't know of a simple way to solve this.

rh101:
> I'm aware of this, but I think if someone will use this option, then he knows what he is doing and it will be an easy fix. And I don't know of a simple way to solve this.

What about adding a `AX_CONSOLE_APP` definition or similar in cmake, via [add_compile_definition](https://cmake.org/cmake/help/latest/command/add_compile_definitions.html#command:add_compile_definitions), that you can check for in the code, and have the correct main function used in `proj.win32/main.cpp` and `main.h` as a result?

```
#if defined(AX_CONSOLE_APP)
int main()
#else
int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmdLine, int nCmdShow)
#endif
{
...
}
```

halx99:
> > I'm aware of this, but I think if someone will use this option, then he knows what he is doing and it will be an easy fix. And I don't know of a simple way to solve this.
> 
> What about adding a `AX_CONSOLE_APP` definition or similar in cmake, via [add_compile_definition](https://cmake.org/cmake/help/latest/command/add_compile_definitions.html#command:add_compile_definitions), that you can check for in the code, and have the correct main function used in `proj.win32/main.cpp` and `main.h` as a result?
> 
> ```
> #if defined(AX_CONSOLE_APP)
> int main()
> #else
> int WINAPI _tWinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPTSTR lpCmdLine, int nCmdShow)
> #endif
> {
> ...
> }
> ```

I fixed it

PR #1863, opened 2024-04-29
Fix FileUtils on macOS

smilediver:
This PR fixes the following issues that `unit-tests` have exposed on macOS.

* Fix `FileUtils::getFileSize()` returning 0 instead of -1 when file is not found
* Fix `FileUtils::fullPathForDirectory()` not returning path with a trailing slash
* Fix `FileUtils::getDefaultResourceRootPath()` returning empty string on macOS
* Fix `FileUtils::isFileExist()` returning true for directories on macOS

### DIFF
--- a/core/platform/FileUtils.cpp
+++ b/core/platform/FileUtils.cpp
@@ -754,14 +754,14 @@ std::string FileUtils::fullPathForDirectory(std::string_view dir) const
                 }
             }
 
-            if (result.empty() and isPopupNotify())
+            if (result.empty() && isPopupNotify())
             {
                 AXLOG("axmol: fullPathForDirectory: No directory found at %s. Possible missing directory.", dir.data());
             }
         }
     }
 
-    if (not result.empty() and result.back() != '/')
+    if (!result.empty() && result.back() != '/')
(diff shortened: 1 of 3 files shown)


core/platform/FileUtils.cpp
```diff
@@ -715,48 +715,58 @@ std::string FileUtils::fullPathForDirectory(std::string_view dir) const
...
+
+            if (result.empty() and isPopupNotify())
+            {
+                AXLOG("axmol: fullPathForDirectory: No directory found at %s. Possible missing directory.", dir.data());
+            }
         }
     }
 
-    if (isPopupNotify())
+    if (not result.empty() and result.back() != '/')
```
rh101: Please avoid using `not` and `and`, as they are not the convention in this project.

smilediver: Fixed

smilediver:
Added two more fixes:

* Fix `FileUtils::getDefaultResourceRootPath()` returning empty string on macOS
* Fix `FileUtils::isFileExist()` returning true for directories on macOS

PR #1871, opened 2024-04-30
Enable running `unit-tests` in GitHub workflow

smilediver:
Trying to enable running `unit-tests` in GitHub workflow.


smilediver:
Windows is hitting #1872 issue.

@halx99 Can you help me in setting up running on macOS? I can't figure out how to insert arch for the command line here: https://github.com/smilediver/axmol/blob/ade1eba39c25f3d1a3909019db4fb9cdaf3067ba/.github/workflows/build.yml#L124

And I also have no idea why Linux fails. I don't know if the path to exe is wrong or something else...

halx99:
sure

halx99:
The linux will crash due to initialize FileUtilsLinux too early before entering `main` entrypoint

halx99:
does the doctest support run tests in `main` function?

halx99:
image

smilediver:
Thanks for getting this to work!

PR #1875, opened 2024-05-02
[Spine] Fix two color tint data not being copied across to backend buffer on first frame

### LINKED ISSUE #1824: Spine Two-Color Tint Bug
- axmol version: 2.1.2
- devices test on: M1 Macbook Pro
- developing environments
   - Xcode version: 15.2
   - cmake version: 3.29.0

Steps to Reproduce:
1. After getting the Axmol project building and running on M1 Macs, update the Spine runtimes to the latest version off GitHub.
2. Add the Spine Cocos2d-x runtimes examples into the new Axmol project, and use the Spine Cocos2d-x runtimes AppDelegate.h instead of the one packaged with Axmol.
3. Change line # 60 from `auto glview = director->getOpenGLView();` to `auto glview = director->getGLView();` and line # 65 from `director->setOpenGLView(glview);` to `director->setGLView(glview);`
4. To better see this issue change line # 107 of AppDelegate.cpp from director->setAnimationInterval(1.0f / 60); to director->setAnimationInterval(1.0f); (This gives us longer to see the first frame failing to render properly.)
5. In AppDelegate.cpp include "RaptorExample.h", un-comment line # 113 and then comment out line # 114, this way the AppDelegate loads up directly into the example that uses two-color tinting.
6. Add the following lines to the RaptorExample.cpp before it's added as a child (e.g. between lines # 53 and # 54): skeletonNode->setDebugBonesEnabled(true); skeletonNode->setDebugMeshesEnabled(true); (This shows that the skeleton is loaded and should be rendered on the first frame, but isn't while two-color tint is enabled.)
7. Run the project on your arm64 Mac, and observe that the Spine-boy Raptor skeleton doesn't render anything except the debug bones and meshes for the first frame, and then renders correctly after that.
8. For posterity, comment out line # 52 of the RaptorExample.cpp class to disable two-color tinting here, and then re-build and re-run the project, observe that this time the Spine-boy Raptor skeleton properly renders on the first frame, as well as subsequent frames.

This bug also exists in Cocos2d-x v4.0, but not Cocos2d-x v3.17.2, likely due to all the Renderer, GL, and Shader changes made to support Metal on Apple devices.

I've submitted a [bug report](https://github.com/EsotericSoftware/spine-runtimes/issues/2501) to the Esoteric Software team as well, but it's very possible that their fix will need to be back ported to Axmol from Cocos2d-x (it'd be nice if they started supporting Axmol officially since Cocos2d-x is dead, but I don't see that happening anytime soon).

I will mention that in my use case it sometimes takes 24 frames or more before the skeleton renders for the first time, and not a single frame like the repro steps I provided.

I've played around with trying to update the skeleton, and the SkeletonTwoColorBatch with `deltaTime = 0.0f`, I've tried to force flush for the lastTwoColorTrianglesCommand on each draw, I've tried disabling culling, setting the skeleton to its setup pose, and updating its world transforms, but nothing solves the issue.

This does bring up two additional Spine related bugs:
1. ~~The version of Spine packaged with Axmol contains the [following bug](https://github.com/EsotericSoftware/spine-runtimes/issues/2401), and should be updated the latest source code from GitHub to fix this bug. This is a critical issue that causes the game to crash when attempting to load a skeleton that contains mesh deformations (which was a headache I had to deal with last week).~~
(That one was my fault due to getting the Axmol 2.1.2 release and not the latest code from GitHub.)
2. ~~CC_USE_CULLING is used in the SkeletonRenderer.cpp code for the Spine Cocos2d-x runtimes, and thus the Axmol runtimes without modifications, but CC_USE_CULLING is never defined to be an alias for AX_USE_CULLING. This causes the Spine runtimes to ignore the Axmol AX_USE_CULLING config option in the Configs.h file.~~
(With the merge of #1825, this issue is now fixed.)

rh101:
## Describe your changes

The two color tint vertex data was being updated after a new render command (`TwoColorTrianglesCommand`) is added, but the issue is that the `addCommand` method would have already updated the back-end buffer with the passed in data.  This meant that the vertex data in the back-end buffer did not contain the updated values.

It's primarily an issue with `TwoColorTrianglesCommand`, because of how it works.  The non-two color tint code uses `TrianglesCommand`, and does not seem to copy the data on `addCommand`, but rather hold a reference to the vertex data to be used later in the rendering process (if I understand it correctly).  Regardless of that, the section for non-two color tint has also been changed to update the data before the `Renderer::addCommand` is called.

#1824 

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/extensions/spine/src/spine/SkeletonRenderer.cpp
+++ b/extensions/spine/src/spine/SkeletonRenderer.cpp
@@ -406,25 +406,28 @@ namespace spine {
 					triangles.indices = batch->allocateIndices(triangles.indexCount);
 					memcpy(triangles.indices, _clipper->getClippedTriangles().buffer(), sizeof(unsigned short) * _clipper->getClippedTriangles().size());
 
-					axmol::TrianglesCommand *batchedTriangles = batch->addCommand(renderer, _globalZOrder, texture, _programState, blendFunc, triangles, transform, transformFlags);
-
-					const float *verts = _clipper->getClippedVertices().buffer();
-					const float *uvs = _clipper->getClippedUVs().buffer();
-                    V3F_C4B_T2F *vertex = batchedTriangles->getTriangles().verts;
-                    for (int v = 0, vn = batchedTriangles->getTriangles().vertCount, vv = 0; v < vn; ++v, vv += 2, ++vertex) {
-                        vertex->vertices.x = verts[vv];
-                        vertex->vertices.y = verts[vv + 1];
+                    const float* verts  = _clipper->getClippedVertices().buffer();
+                    const float* uvs    = _clipper->getClippedUVs().buffer();
+                    V3F_C4B_T2F* vertex = triangles.verts;
+                    for (int v = 0, vn = triangles.vertCount, vv = 0; v < vn;
+                         ++v, vv += 2, ++vertex)
+                    {
+                        vertex->vertices.x  = verts[vv];
+                        vertex->vertices.y  = verts[vv + 1];
                         vertex->texCoords.u = uvs[vv];
                         vertex->texCoords.v = uvs[vv + 1];
-                        vertex->colors = color4B;
+                        vertex->colors      = color4B;
                     }
+					batch->addCommand(renderer, _globalZOrder, texture, _programState, blendFunc, triangles, transform, transformFlags);
 				} else {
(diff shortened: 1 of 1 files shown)


halx99:
cheers, thanks @rh101

PR #1631, opened 2024-01-24
Fix shaders not copying to final build on macOS for non Xcode builds

smilediver:
Which branch your pull-request should merge into?

- `dev`: New features or remove deprecated features

## Describe your changes

Fixes shaders not copying to final build on macOS for non Xcode builds. For example, when using Ninja to build.

### For each PR
- [ ] Add Copyright if it missed:   
      - `"Copyright (c) 2019-present Axmol Engine contributors (see AUTHORS.md)."`
- [ ] I have performed a self-review of my code.
       
   Optional:
   - [ ] I have checked readme and add important infos to this PR.
   - [ ] I have added/adapted some tests too.
          
### For core/new feature PR
- [ ] I have checked readme and add important infos to this PR.
- [ ] I have added thorough tests.

PR #1908, opened 2024-05-21
Fix shaders not copying to final build on macOS for non Xcode builds (2)

smilediver:
## Describe your changes

I attempted to fix this in #1631, but it works only if the shaders were already built (since it scans the `runtime/axslc` dir for files), so you have to rerun CMake again to pick them up. This PR fixes that.

How it works: when `ax_target_compile_shaders()` is called on an app or library target to compile the shaders, it sets `AX_COMPILED_SHADERS` variable on the target. The variable contains a list of compiled shaders. When `ax_setup_app_config()` is called on the app target, it recursively iterates all targets being linked, extracts the list of all shaders from them, and adds them as resources.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,12 +29,6 @@
 
 cmake_minimum_required(VERSION 3.22)
 
-if(APPLE)
-    enable_language(C CXX OBJC OBJCXX)
-else()
-    enable_language(C CXX)
-endif()
-
 project(axmol)
 
 if(XCODE)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,9 @@
 cmake_minimum_required(VERSION 3.22)
 
 if(APPLE)
-    enable_language(OBJC OBJCXX)
+    enable_language(C CXX OBJC OBJCXX)
+else()
+    enable_language(C CXX)
 endif()
 
 project(axmol)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,10 @@
 
 cmake_minimum_required(VERSION 3.22)
 
+if(APPLE)
+    enable_language(OBJC OBJCXX)
+endif()
+
 project(axmol)
 
 if(XCODE)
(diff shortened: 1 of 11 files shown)


halx99:
Note: cmake 3.28+ is strong recommanded in mac/ios/tvos, the new cmake feature EMBED_RESOURCES will auto sync all compiled shaders in to deploy target

smilediver:
Hmm.. CI osx builds show up as green, but they have actually failed.

smilediver:
> Note: cmake 3.28+ is strong recommanded in mac/ios/tvos, the new cmake feature EMBED_RESOURCES will auto sync all compiled shaders in to deploy target

Do you mean `XCODE_EMBED_RESOURCES`? We do use 3.29, but this fix is for non Xcode builds, ie for Ninja.

halx99:
> > Note: cmake 3.28+ is strong recommanded in mac/ios/tvos, the new cmake feature EMBED_RESOURCES will auto sync all compiled shaders in to deploy target
> 
> Do you mean `XCODE_EMBED_RESOURCES`? We do use 3.29, but this fix is for non Xcode builds, ie for Ninja.

yes, I have never use non-xcode for apple platforms to build axmol yet.

halx99:
strange, why the patch cause glfw build error

smilediver:
> strange, why the patch cause glfw build error

I think this is due to CMake 3.20 requirement and adding `enable_language(OBJC OBJCXX)`. After CMake 3.16 I think it needs full language specification like:

```cmake
if(APPLE)
    enable_language(C CXX OBJC OBJCXX)
else()
    enable_language(C CXX)
endif()
```

Pushed fix, lets see if this will fix it.

smilediver:
Looks like everything is working now.

halx99:
> > strange, why the patch cause glfw build error
> 
> I think this is due to CMake 3.20 requirement and adding `enable_language(OBJC OBJCXX)`. After CMake 3.16 I think it needs full language specification like:
> 
> ```cmake
> if(APPLE)
>     enable_language(C CXX OBJC OBJCXX)
> else()
>     enable_language(C CXX)
> endif()
> ```
> 
> Pushed fix, lets see if this will fix it.

What about add to https://github.com/axmolengine/axmol/blob/dev/cmake/Modules/AXConfigDefine.cmake#L214 instead adding to project cmake one by one?

PR #1910, opened 2024-05-21
Allow video media controls to be rotated left or right by 90 degrees

rh101:
## Describe your changes

If a video is played that has content which is rotated by 90 degrees, then it would be nice to be able to rotate the media controls to match that orientation.  There is no way for the code to detect rotated content, since the video is rendered in that way, and not actually rotated, so this PR introduced the ability for the developer to set the rotation.  The 3 options would be Default, RotatedLeft (90°) and RotatedRight (90°).

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/ui/UIMediaPlayer.cpp
+++ b/core/ui/UIMediaPlayer.cpp
@@ -98,13 +98,28 @@ struct PrivateVideoDescriptor
 
                 if (!videoView->isKeepAspectRatioEnabled())
                 {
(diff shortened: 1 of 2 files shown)


rh101:
Example video that is playing content rendered at 90° (same behaviour using any media player):
image

The feature in this PR allows the rotation of the media controls, so a call to:
```
auto* mediaController = videoPlayer->getMediaController();
mediaController->setOrientation(ui::MediaController::Orientation::RotatedRight);
```
would result in this, which looks much better:
image

PR #1913, opened 2024-05-22
Remove box2d-optimized

aismann:
Removed Box2D-optimized
This project is now obsolete; 
Erin Catto has developed Box2D v3 with many performance improvements, so please use that instead

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -1,7 +1,6 @@
 
 project(3rdparty)
 
-option(OPT_BOX2D_OPTIMIZED "" OFF)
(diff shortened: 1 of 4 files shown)


halx99:
shoud we also need remove box2d-optimized from 3rdparty?

halx99:
Also CMakeLists: https://github.com/axmolengine/axmol/blob/dev/3rdparty/CMakeLists.txt#L4 opiton and relatived code

And here: https://github.com/axmolengine/axmol/blob/dev/README_CN.md?plain=1#L85

aismann:
@halx99 
Thanks for removing on  CMakeLists.txt.

PR #1924, opened 2024-05-24
Fix passing non null terminated strings for null terminated params

smilediver:
Fixes some cases where `std::string_view` is passed to `[NSString stringWithUTF8String:]`. `stringWithUTF8String` expects a zero terminated string.

### DIFF
--- a/core/media/AvfMediaEngine.mm
+++ b/core/media/AvfMediaEngine.mm
@@ -194,14 +194,14 @@ bool AvfMediaEngine::open(std::string_view sourceUri)
         // Media Framework doesn't percent encode the URL, so the path portion is just a native file path.
         // Extract it and then use it create a proper URL.
         Path             = sourceUri.substr(7);
-        NSString* nsPath = [NSString stringWithUTF8String:std::string(Path).c_str()];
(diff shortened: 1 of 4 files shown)


halx99:
caller should ensure string_view is null-terminated

core/media/AvfMediaEngine.mm
```diff
@@ -194,14 +194,14 @@ - (void)observeValueForKeyPath:(NSString*)keyPath
         // Media Framework doesn't percent encode the URL, so the path portion is just a native file path.
         // Extract it and then use it create a proper URL.
         Path             = sourceUri.substr(7);
-        NSString* nsPath = [NSString stringWithUTF8String:Path.data()];
+        NSString* nsPath = [NSString stringWithUTF8String:std::string(Path).c_str()];
```
halx99: performance will slow due to copy a std::string

halx99: could you try: https://developer.apple.com/documentation/foundation/nsstring/1407339-initwithbytes?language=objc

smilediver: Changed

smilediver:
> caller should ensure string_view is null-terminated

No, that's a bad idea. The whole purpose of `std::string_view` is to represent a non null-terminated string. So if API takes std::string_view, then it must treat it as not null-terminated.

Issue #1956, opened 2024-06-03
Clipping issue with different child stencils


rh101:
- axmol version:
- devices test on: Windows 10
- developing environments
   - NDK version: r23c
   - Xcode version: 14.2+
   - Visual Studio: 
      - VS version: 2022 (17.9+)
      - MSVC version: 19.39+
      - Windows SDK version: 10.0.22621.0+
   - cmake version: 
Steps to Reproduce:

If different stencils are added as children of a parent stencil, then the last child stencil added is applied to all other stencil positions.

For example, in the cpp-tests `void HoleDemo::pokeHoleAtPoint(Vec2 point)`, change this line:

`auto holeStencil = Sprite::create("Images/hole_stencil.png"); `

to this:
`auto holeStencil = RandomHelper::random_int(0, 1) == 0 ? Sprite::create("Images/hole_stencil.png") : Sprite::create("Images/hole_stencil2.png");`

Add this `hole_stencil2.png` image to the to the `axmol\tests\cpp-tests\Content\Images` folder:
![hole_stencil2](https://github.com/axmolengine/axmol/assets/8603230/4d64d8b3-283a-4e90-bfe7-59b9b1418bff)

Run cpp-tests, then click on "29:Node: Clipping" test.  Then navigate to "2:Hole Demo".  Click on the spinning square image, and you will see the following:
![image](https://github.com/axmolengine/axmol/assets/8603230/15c2ffb9-95b8-47e1-920b-dea5910d3176)
![image](https://github.com/axmolengine/axmol/assets/8603230/39fedd8c-c944-445e-8871-0f77b224f64a)

MP4 video of issue:
https://github.com/axmolengine/axmol/assets/8603230/5c82dd9a-be6d-4246-8193-9b0217f3329e

The problem is that the last image used as a stencil is applied to all other existing stencils.

The expected output is that each new stencil would not affect the previous stencils applied, so there should be a mix of round and square holes in that image.

Is this a limitation of the `ClippingNode`, or is this a bug?

rh101:
I think this has something to do with the `ProgramState` being set on all child stencil nodes. They are all being set to the same `ProgramState` instance, but if this is changed to be a unique instance per child, then the problem is fixed, but memory usage goes up because of some other issue that I'm trying to track down.

rh101:
Issue should now be fixed in PR #1957 

The resulting output is now this:
[...]

PR #1957, opened 2024-06-03
Apply child stencils correctly to the clipping node

rh101:
Fix incorrect reference count for original program states stored in collection

## Describe your changes
When using different child stencils in a `ClippingNode`, the `ProgramState` applied to each child was exactly the same, resulting in incorrect output.

With this change, it now allows the developer to set whether the `ClippingNode` uses the same unique `ProgramState` per child stencil.

An existing issue is also fixed, due to the `_originalStencilProgramState` collection not retaining the `ProgramState` instances saved in it.  These instances were being released by the parent node after being added to the `_originalStencilProgramState`, which meant that they may have been freed if their reference count became 0.  The change in this PR ensures that they are correctly retained prior to being added to the collection.

This change will not affect existing usage, as the new functionality is optional.

#1956 

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/2d/ClippingNode.cpp
+++ b/core/2d/ClippingNode.cpp
@@ -35,7 +35,7 @@
 
 NS_AX_BEGIN
 
-ClippingNode::ClippingNode() : _stencil(nullptr), _stencilStateManager(new StencilStateManager()) {}
+ClippingNode::ClippingNode() : _stencilStateManager(new StencilStateManager()) {}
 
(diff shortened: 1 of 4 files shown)


aismann:
Please add a new test to cpp_test too.

rh101:
> Please add a new test to cpp_test too.

Test added. Output should be the following:
image

PR #1987, opened 2024-06-11
Change cmake version from 3.23 to 3.22

crazyhappygame:
## Describe your changes
Change cmake version from 3.23 to 3.22
android sdk has only 3.22. Using anythin higher then 3.22 enforces to use non android sdk cmake
```
 % ./sdkmanager --list | grep cmake
  cmake;3.22.1         | 3.22.1        | CMake 3.22.1                            | cmake/3.22.1        
  cmake;3.10.2.4988404                                                                     | 3.10.2            | CMake 3.10.2.4988404                                                 
  cmake;3.18.1                                                                             | 3.18.1            | CMake 3.18.1                                                         
  cmake;3.22.1                                                                             | 3.22.1            | CMake 3.22.1                                                         
  cmake;3.6.4111459                                                                        | 3.6.4111459       | CMake 3.6.4111459       
```
### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/1k/dm/CMakeLists.txt
+++ b/1k/dm/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-cmake_minimum_required(VERSION 3.23)
+cmake_minimum_required(VERSION 3.22)
 
 project(dm)
 
--- a/1k/fetch.cmake
+++ b/1k/fetch.cmake
@@ -4,7 +4,7 @@
 #   _1kfetch_cache_dir
 #   _1kfetch_manifest
 # 
-cmake_minimum_required(VERSION 3.23)
+cmake_minimum_required(VERSION 3.22)
 
 ### 1kdist url
 find_program(PWSH_PROG NAMES pwsh powershell NO_PACKAGE_ROOT_PATH NO_CMAKE_PATH NO_CMAKE_ENVIRONMENT_PATH NO_CMAKE_SYSTEM_PATH NO_CMAKE_FIND_ROOT_PATH)


rh101:
The project dictates which version of CMake is required, and should not be restricted to what version is bundled with a specific IDE.  I am not certain if the current 3.23 version set in Axmol is actually necessary for Android, but that is not relevant, because in future Axmol may require a higher version of CMake for functionality or bug fixes, and there is absolutely no reason to be restricting it because the required version is not bundled with an IDE.

crazyhappygame:
Version 3.23 is not required by axmol.
I agree that if axmol will need higher version we should use it.

AlexandreK38:
I may be wrong but in the android [app build.gradle](https://github.com/axmolengine/axmol/blob/dev/templates/common/proj.android/app/build.gradle) template we have
```
// Setup native build tools: ndk & cmake
    def nbtInfo = axutils.findNativeBuildTools(project, '3.28.1+')
    ndkVersion = nbtInfo[0]
    if(nbtInfo[1]) {
        ndkPath = nbtInfo[1]
    }
    def cmakeVer = nbtInfo[2]
    def cmakeOptions = Eval.me(nbtInfo[3])
```
so to me it seems to require cmake 3.28.1 at least, as it is given as version in cmake
```
externalNativeBuild {
        cmake {
            version "$cmakeVer"
            path "../../CMakeLists.txt"
        }
    }
```

halx99:
> I may be wrong but in the android [app build.gradle](https://github.com/axmolengine/axmol/blob/dev/templates/common/proj.android/app/build.gradle) template we have
> 
> ```
> // Setup native build tools: ndk & cmake
>     def nbtInfo = axutils.findNativeBuildTools(project, '3.28.1+')
>     ndkVersion = nbtInfo[0]
>     if(nbtInfo[1]) {
>         ndkPath = nbtInfo[1]
>     }
>     def cmakeVer = nbtInfo[2]
>     def cmakeOptions = Eval.me(nbtInfo[3])
> ```
> 
> so to me it seems to require cmake 3.28.1 at least, as it is given as version in cmake
> 
> ```
> externalNativeBuild {
>         cmake {
>             version "$cmakeVer"
>             path "../../CMakeLists.txt"
>         }
>     }
> ```

Yes, axmol gradle utils will auto find properly cmake in system path

PR #1992, opened 2024-06-14
Don't output log when custom ILogOutput was set

### LINKED ISSUE #1991: Axmol keeps printing logs if custom logger is set with `setLogOutput()`
We have a custom log system for capturing and printing logs, so we reroute Axmol's logs using `setLogOutput()` to our system. But the problem is that Axmol still keeps printing them, so we get duplicated output.

I think if a custom logger is set through `setLogOutput()`, then Axmol should stop printing them. Or at least there should be an option to turn off the output.

halx99:
### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/base/Logging.cpp
+++ b/core/base/Logging.cpp
@@ -161,6 +161,12 @@ AX_API LogItem& preprocessLog(LogItem&& item)
 
 AX_DLL void outputLog(LogItem& item, const char* tag)
 {
+    if (!s_logOutput) writeLog(item, tag);
+    else s_logOutput->write(item, tag);
+}
+
+AX_DLL void writeLog(LogItem& item, const char* tag)
+{
 #if defined(__ANDROID__)
     int prio;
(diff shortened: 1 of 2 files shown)


rh101:
Would it be too complex to allow for having multiple outputs for the logger?  

For instance, having a simple API to clear all registered loggers, add, and output:

```
static std::vector<std::shared_ptr<ILogOutput>> s_loggers;

void clearLoggers()
{
    s_loggers.clear();
}

void addLogger(std::shared_ptr<ILogOutput> logger)
{
    s_loggers.emplace_back(std::move(logger));
}

void outputLog(LogItem& item, const char* tag)
{
    for (auto&& logger : s_loggers)
    {
        logger->write(item, tag);
    }
}
```

This would allow logging to both the console and any custom output sink.  The default console logger should also inherit from `ILogOutput`, so the developer can choose to register it as a logger etc..

PR #2006, opened 2024-06-27
Make debug and verbose level logging in release builds optional

### LINKED ISSUE #2005: Support removal of debug logging calls from builds
At the moment there is no way to strip out all `AXLOGD` and `AXLOGV` calls from the builds, which is something that is supported by the `AXLOG` call (yet it doesn't use `fmt` for formatting).  These calls may contain output that should not be part of a release, whether it's sensitive information or things that would never make sense to have in release builds.

Would it make sense to support something such as the following?

```
#if AX_STRIP_DEBUG_LOGGING
#    define AXLOGV(...) \
        do              \
        {               \
        } while (0)
#    define AXLOGD(...) \
        do              \
        {               \
        } while (0)
#else
#    define AXLOGV(fmtOrMsg, ...) AXLOG_WITH_LEVEL(ax::LogLevel::Verbose, fmtOrMsg, ##__VA_ARGS__)
#    define AXLOGD(fmtOrMsg, ...) AXLOG_WITH_LEVEL(ax::LogLevel::Debug, fmtOrMsg, ##__VA_ARGS__)
#endif
```

By default, `AX_STRIP_DEBUG_LOGGING` would be false, so debug logging would be compiled in, but at least it would give the developer the option to completely strip out any code related to debug and verbose logging calls from a build.

rh101:
## Describe your changes
By default, `AXLOGD` and `AXLOGV` will be disabled in release builds.

#2005

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/base/Logging.h
+++ b/core/base/Logging.h
@@ -139,8 +139,20 @@ inline void printLogT(_FmtType&& fmt, LogItem& item, _Types&&... args)
 #define AXLOG_WITH_LEVEL(level, fmtOrMsg, ...) \
     ax::printLogT(FMT_COMPILE("{}" fmtOrMsg "\n"), ax::preprocessLog(ax::LogItem{level}), ##__VA_ARGS__)
 
-#define AXLOGV(fmtOrMsg, ...) AXLOG_WITH_LEVEL(ax::LogLevel::Verbose, fmtOrMsg, ##__VA_ARGS__)
-#define AXLOGD(fmtOrMsg, ...) AXLOG_WITH_LEVEL(ax::LogLevel::Debug, fmtOrMsg, ##__VA_ARGS__)
+#if defined(_AX_DEBUG) || _AX_DEBUG > 0
+#    define AXLOGV(fmtOrMsg, ...) AXLOG_WITH_LEVEL(ax::LogLevel::Verbose, fmtOrMsg, ##__VA_ARGS__)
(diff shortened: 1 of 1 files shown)


core/base/Logging.h
```diff
@@ -139,8 +139,20 @@ inline void printLogT(_FmtType&& fmt, LogItem& item, _Types&&... args)
 #define AXLOG_WITH_LEVEL(level, fmtOrMsg, ...) \
     ax::printLogT(FMT_COMPILE("{}" fmtOrMsg "\n"), ax::preprocessLog(ax::LogItem{level}), ##__VA_ARGS__)
 
-#define AXLOGV(fmtOrMsg, ...) AXLOG_WITH_LEVEL(ax::LogLevel::Verbose, fmtOrMsg, ##__VA_ARGS__)
-#define AXLOGD(fmtOrMsg, ...) AXLOG_WITH_LEVEL(ax::LogLevel::Debug, fmtOrMsg, ##__VA_ARGS__)
+#if defined(_AX_DEBUG) && _AX_DEBUG > 0
```
smilediver, line 142: The `defined(_AX_DEBUG)` part is not necessary, it could be written as just `#if _AX_DEBUG > 0`. In this case if `_AX_DEBUG` is not defined it's converted to 0.

rh101: Of course! Since it's still correct as-is, we can leave it in there until some other changes need to be done, and address it then.

PR #2007, opened 2024-06-27
Add interfaces for easy expansion

binxiaojiao:


aismann:
@binxiaojiao 
plz add on each file which miss this  `Copyright (c) 2019-present Axmol Engine contributors (see AUTHORS.md).`

core/3d/Mesh.h
```diff
@@ -322,6 +322,10 @@ class AX_DLL Mesh : public Object
     std::vector<float> _spotLightUniformRangeInverseValues;
 
     std::string _texFile;
+
+    // add by binxiaojiao
```
rh101, line 326: Comments like this should not be added, since it doesn't add extra useful information to the code

core/3d/Mesh.h
```diff
@@ -322,6 +322,10 @@ class AX_DLL Mesh : public Object
...
+public:
+    std::function<void()>& GetVisibleChanged() { return _visibleChanged; }
```
rh101, line 328: This should follow the convention used in Axmol, which is to use a lowercase first character, so it should be `getVisibleChanged()`

halx99: I prefer public callback like engine callback renderer comnand

PR #2010, opened 2024-06-30
update package android

armanhossiny:
## Describe your changes

update package :

- androidx.media3:media3-exoplayer-ima
- androidx.annotation:annotation
- com.google.guava:guava


halx99:
Not sure does axmol https://github.com/axmolengine/axmol/blob/dev/core/platform/android/java/src/org/axmol/lib/MediaCodecVideoRenderer.java#L16 works on latest android media 1.3.1 because we modify based on 1.2.1

PR #2012, opened 2024-07-01
Fix error when compile on LINUX ubuntu2204

bwaiox:
## Describe your changes

fix error when compile on LINUX ubuntu2204

None


aismann:
@bwaiox 
maybe this is better?
`  int sph_cmb = (int) sph_yy[0];`
...
` sph_cmb = (int) sph_yy[sph_la];`

PR #2056, opened 2024-07-26
Improve FastRNG implementation

### LINKED ISSUE #2039: FastRNG seems to have problematic implementation
FastRNG seems to have problematic implementation. For example, take the `range()` function:

https://github.com/axmolengine/axmol/blob/602a2cab8fd45144f85b329a6f54429e3aa98a2e/core/math/FastRNG.h#L71-L75

It's unclear what the range is supposed to be. Is it [min, max] or [min, max)? I would expect it to be [min, max), but it seems it's [min, max]. If you call `range(0, 1)`, then you get `rng() / RNG_RAND_MAX`, so almost all random values will be 0, but if you're (un)lucky and it generates `RNG_RAND_MAX`, then it will be 1. If I understand correctly, even if interval is supposed to be [min, max], then the max value has a very low chance to be generated. Also, if the interval is [INT_MIN, INT_MAX], then it lacks the precision.

I think this class needs to be redesigned. The interval should be specified as `[min, max)`, so you could safely do stuff like `array[FastRNG::range(0, array.size)]` or `array[FastRNG::max(array.size)]` . And the math should be carefully rewritten to use doubles. But this needs to be discussed to understand what the original intervals were supposed to be, and if there would be potential breaking changes.

Edit: also the rng() implementation itself is very sketchy, I can't find any docs or other implementations about this particular generator.

DelinWorks:
## Describe your changes
improved FastRNG struct to have better performance, clearer inclusive and exclusiveness, and more uniform distribution of random numbers generated.
also improved particle system and cpp-tests

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/math/FastRNG.h
+++ b/core/math/FastRNG.h
@@ -28,9 +28,6 @@
 #ifndef __FAST_RNG_H__
 #define __FAST_RNG_H__
 
-#include <cmath>
-#include <limits.h>
-
 /** A fast more effective seeded random number generator struct, uses xoshiro128**.
  * It uses a simple algorithm to improve the speed of generating random numbers with a decent quality,
  * Use this if you're planning to generate large amounts of random numbers in a single frame.
--- a/core/math/FastRNG.h
+++ b/core/math/FastRNG.h
@@ -12,87 +12,117 @@
  */
 struct FastRNG
 {
-    const uint32_t RNG_RAND_MAX = std::numeric_limits<uint32_t>::max();
-    uint32_t _x     = 1;
-    uint32_t _y     = 2;
-    uint32_t _z     = 4;
-    uint32_t _w     = 8;
-    uint32_t _carry = 0;
-    uint32_t _k     = 0;
-    uint32_t _m     = 0;
-    uint64_t _seed  = 0;
+    uint32_t s[4];
 
-    FastRNG() { seed_rng((uint32_t)time(NULL)); }
-    FastRNG(uint64_t seed) { seed_rng_64(seed); }
-
-    // initialize this object with seed
-    void seed_rng(uint64_t seed)
+    uint64_t nextSeed(uint64_t& state)
     {
-        auto seed1 = static_cast<uint32_t>(seed);
-        auto seed2 = static_cast<uint32_t>(seed >> 32);
-
-        _seed  = seed;
-        _x     = static_cast<int32_t>(seed1) | 1;
-        _y     = static_cast<int32_t>(seed2) | 2;
-        _z     = static_cast<int32_t>(seed1) | 4;
-        _w     = static_cast<int32_t>(seed2) | 8;
-        _carry = 0;
+        uint64_t z = (state += 0x9e3779b97f4a7c15);
+        z          = (z ^ (z >> 30)) * 0xbf58476d1ce4e5b9;
+        z          = (z ^ (z >> 27)) * 0x94d049bb133111eb;
+        return z ^ (z >> 31);
     }
 
-    // initialize this object with a uint64_t seed
-    // DEPRECATED: use seed_rng instead
-    void seed_rng_64(uint64_t seed) { seed_rng(seed); }
+    FastRNG() { seed(static_cast<uint64_t>(rand()) << 32 | rand()); }
 
-    // returns a random uint32_t value
-    uint32_t rng()
+    void seed(uint64_t seed)
     {
-        _x = _x * 69069 + 12345;
-        _y ^= _y << 13;
-        _y ^= _y >> 17;
-        _y ^= _y << 5;
-        _k     = (_z >> 2) + (_w >> 3) + (_carry >> 2);
-        _m     = _w + _w + _z + _carry;
-        _z     = _w;
-        _w     = _m;
-        _carry = _k >> 30;
-        return _x + _y + _w;
+        uint64_t state = seed;
+        uint64_t states[2];
+        memset(states, 0, 16);
+        states[0] = nextSeed(state);
+        states[1] = nextSeed(state);
+        memcpy(s, states, 16);
     }
 
-    // returns a random integer from min to max
-    int32_t range(int32_t min, int32_t max)
+    static inline uint32_t rotL(const uint32_t x, int k) { return (x << k) | (x >> (32 - k)); }
+
+    uint32_t next()
     {
-        return static_cast<int32_t>(min + rng() / (RNG_RAND_MAX / (max - min)));
+        const uint32_t result = rotL(s[1] * 5, 7) * 9;
+
+        const uint32_t t = s[1] << 9;
+
+        s[2] ^= s[0];
+        s[3] ^= s[1];
+        s[1] ^= s[2];
+        s[0] ^= s[3];
+
+        s[2] ^= t;
+
+        s[3] = rotL(s[3], 11);
+
+        return result;
     }
 
-    // returns a random unsigned integer from min to max
-    uint32_t rangeu(uint32_t min, uint32_t max)
+    // generates a random real that ranges from 0.0 to 1.0
+    template <typename T>
+    T nextReal()
     {
-        return static_cast<uint32_t>(min + rng() / (RNG_RAND_MAX / (max - min)));
+        if (std::is_same<T, float>::value)
+            return static_cast<T>(next() >> 8) * 0x1.0p-24f;
+        else if (std::is_same<T, double>::value)
+            return static_cast<T>((static_cast<uint64_t>(next()) << 32 | next()) >> 11) * 0x1.0p-53;
+        return 0;  // possibly assert?
     }
 
-    // returns a random integer from 0 to max
-    int32_t max(int32_t max = INT_MAX)
+    // generates a random real that ranges from min to max and is uniformly distributed
+    template <typename T>
+    T nextReal(T min, T max)
     {
-        return static_cast<int32_t>(static_cast<float>(rng()) / static_cast<float>(RNG_RAND_MAX / (max - 0)));
+        return static_cast<T>(min + nextReal<T>() * (max - min));
     }
 
-    // returns a random unsigned integer from 0 to max
-    uint32_t maxu(uint32_t max = UINT_MAX) { return static_cast<uint32_t>(0 + rng() / (RNG_RAND_MAX / (max - 0))); }
-
-    // returns a random float from min to max
-    float rangef(float min = -1.0F, float max = 1.0F)
+    // generates a random integer that ranges from min inclusive to max exclusive [min, max) and is uniformly
+    // distributed
+    template <typename T>
+    T nextInt(T min, T max)
     {
-        return min + static_cast<float>(rng()) / (static_cast<float>(RNG_RAND_MAX) / (max - min));
+        return min + static_cast<T>(nextMax(static_cast<uint32_t>(max - min)));
     }
 
-    // returns a random float from 0.0 to max
-    float maxf(float max) { return static_cast<float>(rng()) / (static_cast<float>(RNG_RAND_MAX) / (max - 0.0f)); }
+    // generates a random integer from 0 to max that is uniformly distributed
+    uint32_t nextMax(uint32_t max)
+    {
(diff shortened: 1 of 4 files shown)


halx99:
image
image

```cpp
int main()
{
#if defined(_DEBUG)
    std::print(std::cout, "test on debug build\n");
#else
    std::print(std::cout, "test on release build\n");
#endif
    constexpr int test_rounds = 10;
    constexpr int test_count = 10000000;
    FastRNG rgn;
    std::minstd_rand stdrgn;

    double total1 = 0;
    double total2 = 0;

    rgn.seed(time(NULL));
    stdrgn.seed(time(NULL));

    for (int k = 0; k < test_rounds; ++k) {

        auto start = axmol::highp_clock();
        for (int i = 0; i < test_count; ++i) {
            total1 += rgn.rangef(0, 1);
        }
        auto diff = axmol::highp_clock() - start;

        std::print(std::cout, "FastRGN cost: {} ms\n", diff / 1000.0);

        start = axmol::highp_clock();
        for (int i = 0; i < test_count; ++i) {
            total2 += stdrgn() / (double)stdrgn.max();
        }
        diff = axmol::highp_clock() - start;

        std::print(std::cout, "minstd_rand cost: {} ms\n", diff / 1000.0);
    }

    std::print(std::cout, "total1={}, total2={}", total1, total2);
    return 0;
}
```

release build, fast rng is faster than std::minstd_rand, good job, merged

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
-#define _FAST_RNG_H__
-
-#include <cmath>
-#include <limits.h>
-
-/** A fast more effective seeded random number generator struct, made by kiss rng.
+/****************************************************************************
+ Copyright (c) 2012 cocos2d-x.org
+ Copyright (c) 2013-2016 Chukong Technologies Inc.
+ Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
```
smilediver, line 4: You can drop these three if you was original author of FastRNG.

```
 Copyright (c) 2012 cocos2d-x.org
 Copyright (c) 2013-2016 Chukong Technologies Inc.
 Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
```

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
-    uint32_t _k     = 0;
-    uint32_t _m     = 0;
-    uint64_t _seed  = 0;
-
-    FastRNG() { seed_rng((uint32_t)time(NULL)); }
-    FastRNG(uint64_t seed) { seed_rng_64(seed); }
-
-    // initialize this object with seed
-    void seed_rng(uint64_t seed)
+    uint32_t s[4];
```
smilediver, line 39: I think state should be made private.

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
+
+    // SplitMix64 implementation, doesn't modify any state for this instance
+    // but it is used to seed xoshiro128** state
+    uint64_t nextSeed(uint64_t& state)
```
smilediver, line 43: This should also be private and static.

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
+        uint64_t z = (state += 0x9e3779b97f4a7c15);
+        z          = (z ^ (z >> 30)) * 0xbf58476d1ce4e5b9;
+        z          = (z ^ (z >> 27)) * 0x94d049bb133111eb;
+        return z ^ (z >> 31);
     }
 
-    // initialize this object with a uint64_t seed
-    // DEPRECATED: use seed_rng instead
-    void seed_rng_64(uint64_t seed) { seed_rng(seed); }
+    FastRNG() { seed(static_cast<uint64_t>(rand()) << 32 | rand()); }
```
smilediver, line 51: Please add `FastRNG(uint64_t seed)` back, and maybe it should be extended to `FastRNG(uint64_t seed1, uint64_t seed2 = 0)` since it has 128 bit state.

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
 
-    // returns a random uint32_t value
-    uint32_t rng()
+    // there is no need to seed this instance of FastRNG
+    // because it has already been seeded with rand() by constructor
+    // you can override the seed by giving your own 64-bit seed
+    void seed(uint64_t seed)
```
smilediver, line 56: I'm thinking maybe it would be nice to extend this to `seed(uint64_t seed1, uint64_t seed2 = 0)`. If `seed2` is 0 it could use split max to init the state, if it's not 0, then the seed could be set directly. This way you could have access to full 128 bit seed range.

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
+        memset(states, 0, 16);
+        states[0] = nextSeed(state);
+        states[1] = nextSeed(state);
+        memcpy(s, states, 16);
     }
 
-    // returns a random integer from min to max
-    int32_t range(int32_t min, int32_t max)
+    // returns a copy of x rotated k bits to the left
+    static inline uint32_t rotL(const uint32_t x, int k) { return (x << k) | (x >> (32 - k)); }
```
smilediver, line 67: Should be private.

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
-    uint32_t rangeu(uint32_t min, uint32_t max)
+    // generates a random real that ranges from 0.0 to 1.0
+    template <typename T>
+    T nextReal()
     {
-        return static_cast<uint32_t>(min + rng() / (RNG_RAND_MAX / (max - min)));
+        if (std::is_same<T, float>::value)
+            return static_cast<T>(next() >> 8) * 0x1.0p-24f;
+        else if (std::is_same<T, double>::value)
+            return static_cast<T>((static_cast<uint64_t>(next()) << 32 | next()) >> 11) * 0x1.0p-53;
```
smilediver, line 95: Both ifs above should be constexpr.

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
+        return min + static_cast<T>(nextMax(static_cast<uint32_t>(max - min)));
     }
 
-    // returns a random float from 0.0 to max
-    float maxf(float max) { return static_cast<float>(rng()) / (static_cast<float>(RNG_RAND_MAX) / (max - 0.0f)); }
-
-    // returns a random float from 0.0 to 1.0
-    float float01() { return static_cast<float>(rng()) / (static_cast<float>(RNG_RAND_MAX) / (1.0f - 0.0f)); }
+    // generates a random integer from 0 to max exclusive that is uniformly distributed using fastrange algorithm
+    uint32_t nextMax(uint32_t max)
```
smilediver, line 114: This should be private.

tests/unit-tests/Source/core/math/FastRNGTests.cpp
```diff
@@ -29,100 +29,196 @@
...
-        CHECK_EQ(doctest::Approx(-0.318979), rng.rangef(-1.0f, 1.0f));
-        CHECK_EQ(doctest::Approx(19.1121), rng.rangef(10.0f, 20.0f));
+        CHECK_EQ(doctest::Approx(-0.210655), rng.rangef(-1.0f, 1.0f));
+        CHECK_EQ(doctest::Approx(-0.33731), rng.rangef(-1.0f, 1.0f));
+        CHECK_EQ(doctest::Approx(11.4775), rng.rangef(10.0f, 20.0f));
         // These overflow
-        //CHECK_EQ(doctest::Approx(1.0), rng.rangef(-FLT_MAX, FLT_MAX));
-        //CHECK_EQ(doctest::Approx(1.0), rng.rangef(-FLT_MAX, FLT_MAX));
+        // CHECK_EQ(doctest::Approx(1.0), rng.rangef(-FLT_MAX, FLT_MAX));
+        // CHECK_EQ(doctest::Approx(1.0), rng.rangef(-FLT_MAX, FLT_MAX));
```
smilediver, line 161: Are these still overflowing?

halx99:
BTW: do we need move class `FastRNG` into namespace `ax`?

DelinWorks:
thank @smilediver he's the one that wrote and suggested these awesome optimizations

DelinWorks:
> BTW: do we need move class `FastRNG` into namespace `ax`?

come to think of it yeah it should be in the ax namespace

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+ ****************************************************************************/
+
+#ifndef __FAST_RNG_H__
+#define __FAST_RNG_H__
+
```
smilediver, line 30: This is missing a bunch of headers, people will get errors trying to compile.

DelinWorks:
also some functions need to be static too, I'll do PR

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
+    double ranged(double min = -1.0F, double max = 1.0F) { return nextReal<double>(min, max); }
+    // wrapper for nextReal<double>(0, max)
+    double maxd(double max) { return nextReal<double>(0, max); }
+
+    // wrapper for nextReal<float>()
+    float float01() { return nextReal<float>(); }
+    // wrapper for nextReal<double>()
+    float double01() { return nextReal<double>(); }
+    // wrapper for next() & 1, true or false based on LSB
+    float bool01() { return next() & 1; }
```
smilediver, line 155: Return type should be bool. Also add static cast to bool.

core/math/FastRNG.h
```diff
@@ -1,98 +1,158 @@
...
+
+    // wrapper for nextReal<double>(min, max)
+    double ranged(double min = -1.0F, double max = 1.0F) { return nextReal<double>(min, max); }
+    // wrapper for nextReal<double>(0, max)
+    double maxd(double max) { return nextReal<double>(0, max); }
+
+    // wrapper for nextReal<float>()
+    float float01() { return nextReal<float>(); }
+    // wrapper for nextReal<double>()
+    float double01() { return nextReal<double>(); }
```
smilediver, line 153: Return type should be double

smilediver (commented):
I've also noticed that this never was in `ax` namespace. Maybe it should be added?

core/math/FastRNG.h
DelinWorks: initializing the state with your own two uint64_t is a bad idea, they recommend that splitmix64 is used for seeding xoshiro128**

core/math/FastRNG.h
DelinWorks: you already do have the 128 bit seed range that's derived from the splitmix64 function

tests/unit-tests/Source/core/math/FastRNGTests.cpp
DelinWorks: yes

core/math/FastRNG.h
DelinWorks: all github CIs pass so no problem, the includes are brought from other places in the engine so it would compile fine, unless you extract the FastRNG class

halx99:
miss include `math/MathBase.h`

core/math/FastRNG.h
smilediver: You can't have 128 bits by supplying just 64 bits. But maybe it's ok, since authors do recommend it this way. Maybe there's something to it.

core/math/FastRNG.h
smilediver: If you include only `FastRNG.h` it will fail to compile. This is not good. It may not compile with all Axmol headers either.

PR #2057, opened 2024-07-27
add FastRNG to ax ns and more improvements 

DelinWorks:
## Describe your changes
#2056

### DIFF
--- a/core/math/FastRNG.h
+++ b/core/math/FastRNG.h
@@ -28,6 +28,8 @@
 #ifndef __FAST_RNG_H__
 #define __FAST_RNG_H__
 
+NS_AX_MATH_BEGIN
+
 /** A fast more effective seeded random number generator struct, uses xoshiro128**.
  * It uses a simple algorithm to improve the speed of generating random numbers with a decent quality,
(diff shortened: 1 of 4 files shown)


core/math/FastRNG.h
```diff
@@ -28,6 +25,10 @@
 #ifndef __FAST_RNG_H__
 #define __FAST_RNG_H__
 
+#include<math/MathBase.h>
```
halx99:
```suggestion
#include "math/MathBase.h"
```

halx99:
also should add `include <stdint.h>`

core/math/FastRNG.h
```diff
@@ -26,6 +26,8 @@
 #define __FAST_RNG_H__
 
 #include "math/MathBase.h"
+#include <iostream>
```
halx99: well, iostream not required

DelinWorks: for std::is_same

DelinWorks: I guess iostream replaced with type_traits

PR #2061, opened 2024-07-30
Optimize ProgramState::setTexture() to avoid vector (de)allocations

smilediver:
`ProgramState::setTexture()` on each call creates new `TextureInfo`, which allocates new vectors, and then assigns it to another `TextureInfo` freeing old vectors and moving the new ones. This optimization removes these redundant (de)allocations by setting the values directly and reusing the old vectors if possible.

### DIFF
--- a/core/renderer/backend/ProgramState.cpp
+++ b/core/renderer/backend/ProgramState.cpp
@@ -127,6 +127,25 @@ void TextureInfo::assign(TextureInfo&& other)
     }
 }
 
+void TextureInfo::assign(int slot, int index, backend::TextureBackend* texture)
+{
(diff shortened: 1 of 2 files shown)


core/renderer/backend/ProgramState.cpp
```diff
@@ -127,6 +127,22 @@ void TextureInfo::assign(TextureInfo&& other)
     }
 }
 
+void TextureInfo::assign(int slot, int index, backend::TextureBackend* texture)
+{
+    if (textures.size() != 1 or textures[0] != texture or slots[0] != slot or indexs[0] != index)
+    {
+        releaseTextures();
```
halx99, line 134: should invoke indexs.clear(); slots.clear() in releaseTextures if you use xxx.resize(1, val), otherwise the val can't be update

PR #2070, opened 2024-08-03
Refactor math simd

halx99:
## Describe your changes

- Rewrite all neon inline asm code with NEON intrinsics, this will allow `WASM` and `winuwp-arm64` can compile with NEON intrinsic
- Implement SSE transformVertices & transformIndices
- Add cmake option `AX_WASM_ISA_SIMD` to support build wasm with SIMD support, supports values `sse`, `neon`

Relative PR: https://github.com/axmolengine/axmol/pull/2065

### For each PR
       
   Optional:
          
### For core/new feature PR


core/platform/PlatformMacros.h
```diff
@@ -444,15 +465,25 @@ public:                                                             \
...
 #else
-#    define UTILS_LIKELY(exp) (!!(exp))
+#    define UTILS_LIKELY(exp)   (!!(exp))
 #    define UTILS_UNLIKELY(exp) (!!(exp))
 #endif
 
+#if defined(_MSC_VER)
+// MSVC does not support loop unrolling hints
+#    define UTILS_UNROLL
+#    define UTILS_NOUNROLL
```
smilediver, line 482: Shouldn't these be with `AX_` prefix? And probably could be shortened to just `AX_UNROLL` and `AX_NOUNROLL`.

halx99: > Shouldn't these be with `AX_` prefix? And probably could be shortened to just `AX_UNROLL` and `AX_NOUNROLL`.

yes, should be better

PR #2072, opened 2024-08-04
Improve Vec4Base sub-class functionality and Vec4 directional vector method name change

rh101:
## Describe your changes

Move methods from `Vec4` to `Vec4Base`, specifically `set(x,y,z,w)` and `set(float*)`.  This allows their usage in Color4F and other sub-classes of `Vec4Base`.

Remove trivial copy constructor method, since it does not need to be explicitly implemented.

Rename `Vec4::set(const Vec4& p1, const Vec4& p2)` method to `Vec4::setDirection` to reflect its actual purpose.

If anyone currently uses the `Vec4::set(const Vec4& p1, const Vec4& p2)`, then it is a simple rename to `Vec4::setDirection(const Vec4& p1, const Vec4& p2)`.  The `set` name was a bit misleading, and this change renames it to what the method actually does, which is to calculate the directional vector from point `p1` to point `p2`.  The compiler will throw an error regarding this, so there is no danger that this goes unnoticed.

Note that I did try to add the `AX_DEPRECATED_ATTRIBUTE` specifier to the the `Vec4::set(const Vec4& p1, const Vec4& p2)` method, but the compiler would throw errors when attempting to use the other `set` methods that were moved to `Vec4Base`, since they were being hidden by the `Vec4::set` method.

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/math/Vec4.cpp
+++ b/core/math/Vec4.cpp
@@ -1,6 +1,7 @@
 /**
  Copyright 2013 BlackBerry Inc.
  Copyright (c) 2017-2018 Xiamen Yaji Software Co., Ltd.
+ Copyright (c) 2019-present Axmol Engine contributors (see AUTHORS.md).
 
  Licensed under the Apache License, Version 2.0 (the "License");
  you may not use this file except in compliance with the License.
--- a/core/math/Vec4.cpp
+++ b/core/math/Vec4.cpp
@@ -27,23 +27,15 @@
 
 NS_AX_MATH_BEGIN
 
-Vec4::Vec4() : Vec4Base() {}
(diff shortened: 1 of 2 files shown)


aismann:
@rh101 
Is the title correct? 
...Enhanced Color4F ...

rh101:
> Is the title correct?
> ...Enhanced Color4F ...

Fair point, as initially there was changes to the Color4F class, but they I reverted them before creating this PR, since they were not required. I'll fix the title to be more appropriate, since the enhancement to Color4F was more a side-effect of the change.

PR #2075, opened 2024-08-05
Add missing include

smilediver:
Adds missing include

### DIFF
--- a/core/math/Vec4.h
+++ b/core/math/Vec4.h
@@ -24,6 +24,7 @@
 #ifndef MATH_VEC4_H
 #define MATH_VEC4_H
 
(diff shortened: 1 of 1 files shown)


halx99:
for which macro ?

rh101:
> for which macro ?

My mistake, since I moved `void set(const float* array)` from `Vec4.cpp` to `Vec4.h`, and it's using the `GP_ASSERT` macro in the body of that function.  Strange that it didn't throw a compilation error in the build.

halx99:
pch, I guess

smilediver:
Yup, it's for `GP_ASSERT`. We include individual headers, so we will catch issues like these.

PR #2084, opened 2024-08-12
[ImGui] Allow setting of custom font config to enable merging of multiple fonts and other features

rh101:
## Describe your changes

When adding multiple fonts to ImGui, the only way to merge fonts is to set the `ImFontConfig::MergeMode = true`, otherwise glyphs from additional fonts do not render correctly.

This change allows custom `ImFontConfig` to be set on a font that is added via `ImGuiPresenter::addFont`.  This change does not affect existing usage of the `ImGuiPresenter::addFont` method, because the `fontConfig` parameter has a default value of `ImFontConfig()`.  The default value is valid, since it is what ImGui does internally if a `nullptr` is passed to one of the `ImGui::AddFontXYZ` methods.

For example:
```
auto* presenter = extension::ImGuiPresenter::getInstance();
presenter->addFont("fonts/arial.ttf", 14.f);

ImFontConfig fontConfig;
fontConfig.MergeMode = true; // This is set to true for any additional fonts. This must be false for the first font passed in
presenter->addFont("fonts/fa-regular-400.ttf", 14.f, { ICON_MIN_FA, ICON_MAX_FA, 0 } , fontConfig); // example of adding Font Awesome glyphs
```

### For each PR
       
   Optional:
          
### For core/new feature PR


rh101:
Recreating this pull request since there seems to be some error with the branch

PR #2096, opened 2024-08-20
Replace DrawNode (v1) with DrawNode (v2)

aismann:
## Describe your changes

Replace DrawNode (v1) with DrawNode (v2)
->Remove extension DrawNodeEx

### For each PR
       
   Optional:
          
### For core/new feature PR


aismann:
@hal, @rh101 
How can I fix this issue:
/home/runner/work/axmol/axmol/extensions/scripting/lua-bindings/auto/axlua_base_auto.cpp:52459:75: error: too few arguments to function call, expected at least 10, have 9
 52459 |         cobj->drawPie(arg0, arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8);
       |         ~~~~~~~~~~~~~                                                     ^
/home/runner/work/axmol/axmol/core/2d/DrawNode.h:689:10: note: 'drawPie' declared here
  689 |     void drawPie(const ax::Vec2& center,
      |          ^       ~~~~~~~~~~~~~~~~~~~~~~~

rh101:
> @hal, @rh101 How can I fix this issue: /home/runner/work/axmol/axmol/extensions/scripting/lua-bindings/auto/axlua_base_auto.cpp:52459:75: error: too few arguments to function call, expected at least 10, have 9 52459 | cobj->drawPie(arg0, arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8); | ~~~~~~~~~~~~~ ^ /home/runner/work/axmol/axmol/core/2d/DrawNode.h:689:10: note: 'drawPie' declared here 689 | void drawPie(const ax::Vec2& center, | ^ ~~~~~~~~~~~~~~~~~~~~~~~

Don't you just need to regenerate the LUA bindings? Also, there should be a lot more errors than just `drawPie`, since the fields have changed in other methods too, but since the parameter count is the same, it may not be throwing errors for them.

aismann:
> Don't you just need to regenerate the LUA bindings? Also, there should be a lot more errors than just `drawPie`, since the fields have changed in other methods too, but since the parameter count is the same, it may not be throwing errors for them.

How can I do it?
I m not familiar with "axmol lua".

rh101:
> How can I do it?

Run `genbindings.ps1`, which is in `tools/ci/`

aismann:
DrawNodeV2 v0.97 is ready for review now.

core/2d/DrawNode.cpp
```diff
@@ -169,39 +234,39 @@ bool DrawNode::init()
...
-
+    ensureCapacityTriangle(512);
     _dirtyTriangle = true;
-    _dirtyLine     = true;
-    _dirtyPoint    = true;
+
+    ensureCapacityPoint(64);
+    ensureCapacityLine(256);
+    _dirtyLine = true;
+    _dirtyPoint = true;
```
smilediver: I think starting sizes are way too big, and for simple cases (like rectangle masks) wastes a lot of memory. I think it would be better to change buffers to use `std::vector` and start empty.

core/2d/DrawNode.cpp
```diff
@@ -281,219 +346,176 @@ void DrawNode::draw(Renderer* renderer, const Mat4& transform, uint32_t flags)
...
+
     Vec2* vertices = _abuf.get<Vec2>(segments + 1);
 
     float t = 0.0f;
     for (unsigned int i = 0; i < segments; i++)
     {
-        vertices[i].x = powf(1 - t, 2) * origin.x + 2.0f * (1 - t) * t * control.x + t * t * destination.x;
-        vertices[i].y = powf(1 - t, 2) * origin.y + 2.0f * (1 - t) * t * control.y + t * t * destination.y;
+        vertices[i].x = powf(1.0f - t, 2.0f) * origin.x + 2.0f * (1.0f - t) * t * control.x + t * t * destination.x;
+        vertices[i].y = powf(1.0f - t, 2.0f) * origin.y + 2.0f * (1.0f - t) * t * control.y + t * t * destination.y;
```
smilediver: I don't know if compiler can optimize the `powf()` call... it would be better to square by hands. Note, that there are more `powf()` calls in other places.

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
+
+    // See also example on https://www.angusj.com/clipper2/Docs/Units/Clipper/Types/EndType.htm
+    enum EndType
+    {
+        Square,
+        Round,
+        Butt,
+    };
+
+    enum class Version
```
smilediver: Why is this needed?

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
+        v0,  // DrawNode cocos2dx/axmol 1.0
+        v1,  // DrawNode 1.0
+        v2,  // DrawNode 2.0 (aka DrawNode)
+        v3,
+        v4,
+    };
+
+
+    // Internal function _drawPoint
+    void _drawPoint(const Vec2& position, const float pointSize, const Color4B& color, const DrawNode::PointType pointType);
```
smilediver: All these internal methods should be `private`.

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
+        float thickness);
+
+    /** Set backwards compatible with cocos2dx/axmol 2.0
+    *
+    * @param isConvex swap the '_isConvex' flag.
+    * @js NA
+    */
+    bool swapIsConvex(bool isConvex) {
+        _isConvexTmp = _isConvex; _isConvex = isConvex; return _isConvexTmp;
+    };
```
smilediver: This is so confusing... what is going on here and why this is needed? :)

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
+
+    /** Set the DrawNode draw order
+    *
+    * @param drawOrder. true/false = On/Off
+
+    * @js NA
+    */
+    void setDNDrawOrder(bool drawOrder) {
+        _drawOrder = drawOrder;
```
smilediver: What does this do? Needs a better comment.

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
+    };
+
+    /** Set the DrawNode transformation
+    *
+    * @param transform. true/false = On/Off
+    *
+    * @js NA
+    */
+    void setDNTransform(bool transform) {
+        _dnTransform = transform;
```
smilediver: Needs better comment.

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
+    */
+    void setDNScaleX(float scaleX) {
+        _dnScale.x = scaleX;
+    };
+
+    /** Set the DrawNode scaleY for each drawing primitive after this.
+
+    * @js NA
+    */
+    void setDNScaleY(float scaleY) {
```
smilediver: There's already `setDNScale`, maybe remove `setDNScaleX` and `setDNScaleY`?

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
+
+
+    /** DrawNode transform method.
+    *
+    * @param vertices A Vec2 vertices list.
+    * @param count The number of vertices.
+    * @param closedPolygon The closedPolygon flag.
+    * @js NA
+    */
+    ax::Vec2* transform(const ax::Vec2* vertices, unsigned int& count, bool closedPolygon = false);
```
smilediver: Does this really need to be public? Looks like internal method to me.

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
+        _dnRotation = 0.0f;
+        _dnPosition = Vec2(0.f, 0.f);
+        _dnLineWidth = 1.0f;
+    };
+
+    /** Ensure the DrawNode properties.
+
+    * @js NA
+    */
+    void ensureDNValues()
```
smilediver: Probably should be `saveDNValues()`.

But I think `ensureDNValues()` and `restoreDNValues()` could be replaced by something like `setDNSettings()` and `getDNSettings()` that would get and set a struct with these values. This way you don't have a bunch of useless tmp members inside the class, and user can better control what he saves, restores and how many copies he has. With this design you could even get rid of all the `getDNXXX()` methods and use `getDNSettings().position`, etc. The `Settings` name is an example, maybe there's a better one. Then `resetDNValues()` could also be removed and `setDNValues()` could have a default parameter for resetting.

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
-    /** Draw a point.
+     /** Draw a point.
      *
      * @param point A Vec2 used to point.
      * @param pointSize The point size.
      * @param color The point color.
      * @js NA
      */
-    void drawPoint(const Vec2& point, const float pointSize, const Color4B& color);
+    void drawPoint(const ax::Vec2& point, const float pointSize, const ax::Color4B& color, DrawNode::PointType pointType = DrawNode::PointType::Rect);
```
smilediver: I would not expect a point to be a rectangle by default... I think it should be a circle. Or this was an old behavior?

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
+    * @param destination The line destination.
+    * @param color The line color.
+    * @js NA
+    */
+    void drawLine(const ax::Vec2& origin,
+        const ax::Vec2& destination,
+        const ax::Color4B& color,
+        float thickness = 1.0f,
+        DrawNode::EndType etStart = DrawNode::EndType::Round,
+        DrawNode::EndType etEnd = DrawNode::EndType::Round);
```
smilediver: Maybe default line endings should be square? Not sure...

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
+    void setDNDrawOrder(bool drawOrder) {
+        _drawOrder = drawOrder;
+    };
+
+    /** Swap the _drawOrder flag.
+    *
+    * @js NA
+    */
+    bool swapDNDrawOrder() {
+        return _drawOrder = (_drawOrder) ? false : true;
```
smilediver: These all swap methods are very strange. I would get rid of them...

core/2d/DrawNode.h
```diff
@@ -70,399 +75,826 @@ class AX_DLL DrawNode : public Node
...
 
-    bool _dirtyTriangle     = false;
-    bool _dirtyPoint        = false;
-    bool _dirtyLine         = false;
-    bool _isolated          = false;
-    float _lineWidth        = 0.0f;
-    float _defaultLineWidth = 0.0f;
+    // transforming stuff
+    ax::Vec2 _dnScale;
+    ax::Vec2 _dnScaleTmp = _dnScale;
```
smilediver: Are you sure that initialization order for things like these are correct combined with the constructor? I think it's very confusing that part is being initialized here, and another part is in constructor. Either everything should be initialized here, or everything should be initialized in the constructor.

smilediver (commented):
These `swapIsConvex(true);` calls are really suspicious and should be more thoroughly reviewed.

core/2d/DrawNode.cpp
aismann: axmol has 79 places => maybe its better writing a issue for replacing "most of all powf" with a 'hand optimized' method.   
image

core/2d/DrawNode.h
aismann: Its the possible to detect the 'DrawNode version' on customer side.
But can be to much...axmol version should be enough.
I remove it.

core/2d/DrawNode.h
aismann: done.

core/2d/DrawNode.cpp
aismann: Thats the 'original'  values from cocos. At the moment I dont want change it.
image

core/2d/DrawNode.h
aismann: Correct. Its the behavior of Cocos2dx

core/2d/DrawNode.h
aismann: Thats a internal method (I will make it private and rename too.)
This method is very important for 'performance'.
Its used on _drawPolygon() and switch the polytri fill with the 'old cocos2dx fill method (like a fan).
Some primitives can use "allways' the fill method 'fan' which is faster.
swapIsConvex method set its and 'save the current value' and restore the value again after the drawing.

core/2d/DrawNode.h
aismann: You are right. I will use only one initialized way.

core/2d/DrawNode.h
aismann: Using the default from drawSegment (Round is needed if you draw connected lines too)

core/2d/DrawNode.h
aismann: Removed it with setter/getter

core/2d/DrawNode.h
aismann: I will do.
Its a fast switch on/off for all DN properties

core/2d/DrawNode.h
aismann: You are right.

core/2d/DrawNode.h
aismann: Ok. I will remove.

core/2d/DrawNode.h
aismann: Its for performance there 
off = cocos2dx behaviour => faster but works only on 1.0f thickness

core/2d/DrawNode.h
aismann: I renamed it:
-     void setDefaultDNValues()
-     void saveDNValues()
-     void restoreDNValues()

aismann:
DrawNodeV2 is ready for review again

aismann:
> These `swapIsConvex(true);` calls are really suspicious and should be more thoroughly reviewed.

Is an internal method now
And has a better comment:
    // Internal function _swapIsConvex
    // Using on _drawPolygon() and set the fill method "fan" 'hardcoded' which is faster as fill method "PoliTri"

aismann:
@halx99 
For me it can merge.

core/2d/DrawNode.h
```diff
@@ -79,7 +74,7 @@ class AX_DLL DrawNode : public ax::Node
     {
         Circle,
         Rect,
-        Triangle,
+        Triangle, // not implemented at the moment
```
smilediver: If it's not implemented, then it should be removed.

smilediver:
> Is an internal method now And has a better comment: // Internal function _swapIsConvex // Using on _drawPolygon() and set the fill method "fan" 'hardcoded' which is faster as fill method "PoliTri"

It looks like it should be a parameter for `_drawPolygon()` and `_drawPoly()` methods.

Also `_swapIsConvex(true)` is called for Bezier and Spline curves... is this right?

core/2d/DrawNode.h
aismann: good idea. I comment it (so it will not lost)

aismann:
> > Is an internal method now And has a better comment: // Internal function _swapIsConvex // Using on _drawPolygon() and set the fill method "fan" 'hardcoded' which is faster as fill method "PoliTri"
> 
> It looks like it should be a parameter for `_drawPolygon()` and `_drawPoly()` methods.
As long the backwards behavior of the cocos2dx fill method should supported as long it have to be a seperate method.
=> It can set to "Mark as deprecated".  => The "Backwards compatibility"
See: 
image


> Also `_swapIsConvex(true)` is called for Bezier and Spline curves... is this right?
Yes its right! All 'curves' be (small) lines. And lines be like a rect.

smilediver (commented):
My biggest gripe is `saveDNValues()` and `restoreDNValues()`. They force each DrawNode to carry tmp fields, when most people will never use them. Usage is also limited, since you can save only once, and you have no idea if something is already saved there or not without checking all your code.

And I think `_isConvex` field could be removed completely.

Also maybe all these `getDN` and `setDN` could have a better name.

---

This is just a side note, and general advice for the future, but in cases like these, a better overall design is to make `DrawNode` a simple container of primitives that are drawn. It would have methods that can only clear or take additional primitives. So you end up with a simple and clean class that looks something like this:

```cpp
class DrawNode: Node {
    std::vector<Vertex> _triangles;
    TrianglesCommand _cmd;
public:
    void clear();
    void addTriangles(...);
    void draw();
};
```

Then, you can have another class like `Drawer` (probably bad name, just for example) that can create different geometry objects to fill the data for `DrawNode`. So you end up with something like this:

```cpp
class Drawer {
    std::vector<Vertex> _triangles;
    ... // Other data actual for drawing, like scale, line width, etc...
public:
    void setPosition(...);
    void drawLine(...);
    void drawStar(...);
    std::vector<Vertex> getTriangles() { return _triangles; }
};
```

And then you can have something like:

```cpp
void drawSomething() {
    auto d = Drawer();
    d.setPosition(...);
    d.drawLine(...);
    d.drawStar(...);
    getSomeDrawNode().addTriangles(d.getTriangles());
}
```

This way `Drawer` can keep any data it needs while creating geometry, but `DrawNode` needs to keep only a triangle list (or other draw primitives). This way you also get a modular design, where both classes are fully independent, so you can use `Drawer` for other things, and you can fill triangles for `DrawNode` from other places.

aismann:
> My biggest gripe is `saveDNValues()` and `restoreDNValues()`. They force each DrawNode to carry tmp fields, when most people > will never use them. 

I remove it. I understand what you mean of course.

aismann:
> This is just a side note, and general advice for the future, but in cases like these, a better overall design is to make `DrawNode` a simple container of primitives that are drawn. It would have methods that can only clear or take additional primitives. So you end up with a simple and clean class that looks something like this:
> 
> ```c++
> class DrawNode: Node {
>     std::vector<Vertex> _triangles;
>     TrianglesCommand _cmd;
> public:
>     void clear();
>     void addTriangles(...);
>     void draw();
> };
> ```
> 
> Then, you can have another class like `Drawer` (probably bad name, just for example) that can create different geometry objects to fill the data for `DrawNode`. So you end up with something like this:
> 
> ```c++
> class Drawer {
>     std::vector<Vertex> _triangles;
>     ... // Other data actual for drawing, like scale, line width, etc...
> public:
>     void setPosition(...);
>     void drawLine(...);
>     void drawStar(...);
>     std::vector<Vertex> getTriangles() { return _triangles; }
> };
> ```
> 
> And then you can have something like:
> 
> ```c++
> void drawSomething() {
>     auto d = Drawer();
>     d.setPosition(...);
>     d.drawLine(...);
>     d.drawStar(...);
>     getSomeDrawNode().addTriangles(d.getTriangles());
> }
> ```
Sorry no time for redesign of DrawNode again.
The goal of the current DrawNode (v2) is/was be compatible as much as possible with the Cocos2dx v4.0. DrawNode API

core/2d/DrawNode.cpp
```diff
@@ -38,71 +38,76 @@
...
- * @param count The number of verts measured in points.
- */
+* @param verts A pointer to point coordinates.
+* @param count The number of verts measured in points.
+*/
 inline bool isConvex(const Vec2* verts, int count)
 {
     bool isPositive = false, isNegative = false;
-    for (int i = 0; i < count; i++)
+    for (unsigned int i = 0; i < count; i++)
```
halx99:
```suggestion
    for (int i = 0; i < count; i++)
```

core/2d/DrawNode.cpp
```diff
@@ -38,71 +38,76 @@
...
     {
         Vec2 A = verts[i];
```
halx99:
```suggestion
        auto& A = verts[i];
```

core/2d/DrawNode.cpp
```diff
@@ -38,71 +38,76 @@
...
         Vec2 B = verts[(i + 1) % count];
```
halx99:
```suggestion
        auto& B = verts[(i + 1) % count];
```

core/2d/DrawNode.cpp
```diff
@@ -38,71 +38,76 @@
...
         Vec2 C = verts[(i + 2) % count];
```
halx99:
```suggestion
        auto& C = verts[(i + 2) % count];
```

core/2d/DrawNode.cpp
```diff
@@ -503,288 +460,852 @@ void DrawNode::drawCardinalSpline(PointArray* config, float tension, unsigned in
...
+    float thickness)
+{
+    if (thickness == 0.0f) return;
 
-    V2F_C4B_T2F_Triangle* triangles = (V2F_C4B_T2F_Triangle*)(_bufferTriangle + _bufferCountTriangle);
-    V2F_C4B_T2F_Triangle triangle0  = {a, b, c};
-    V2F_C4B_T2F_Triangle triangle1  = {a, c, d};
-    triangles[0]                    = triangle0;
-    triangles[1]                    = triangle1;
+    Vec2 line[5] = { {p1}, {p2}, {p3}, {p4}, {p1} };
```
halx99:
```suggestion
    Vec2 line[5] = { p1, p2, p3, p4, p1 };
```

core/2d/DrawNode.cpp
```diff
@@ -503,288 +460,852 @@ void DrawNode::drawCardinalSpline(PointArray* config, float tension, unsigned in
...
+                Vec2 v2 = b - nw;
+                Vec2 v3 = b + nw;
+                Vec2 v4 = a - nw;
+                Vec2 v5 = a + nw;
+                Vec2 v6 = a - (nw - tw);
+                Vec2 v7 = a + (nw + tw);
+
+                {
+                    triangles[ii++] = {
+                        {v0, borderColor, Tex2F(-(n + t))},
```
halx99: since Vec2, Tex2F is same type, no need convert

core/2d/DrawNode.cpp
```diff
@@ -503,288 +460,852 @@ void DrawNode::drawCardinalSpline(PointArray* config, float tension, unsigned in
...
+        else
         {
-            Vec2 v0 = verts[(i - 1 + count) % count];
-            Vec2 v1 = verts[i];
-            Vec2 v2 = verts[(i + 1) % count];
+            struct ExtrudeVerts
+            {
+                Vec2 offset, n;
+            };
+            struct ExtrudeVerts* extrude = (struct ExtrudeVerts*)malloc(sizeof(struct ExtrudeVerts) * count);
```
halx99: recommend to use ax::pod_vector

core/2d/DrawNode.cpp
```diff
@@ -503,288 +460,852 @@ void DrawNode::drawCardinalSpline(PointArray* config, float tension, unsigned in
...
+        };
+        break;
+
+    default:
+        break;
+    }
+
+    // BODY
+    triangles[ii++] = {
+        {v3, color, Tex2F(n)},
```
halx99: Tex2F can be removed

core/2d/DrawNode.cpp
```diff
@@ -503,288 +460,852 @@ void DrawNode::drawCardinalSpline(PointArray* config, float tension, unsigned in
...
+    bool solid,
+    float thickness)
 {
-    Vec2 vertices[] = {origin, Vec2(destination.x, origin.y), destination, Vec2(origin.x, destination.y)};
-    _isConvex = true;  // Fix issue #1546 of UILayout(#1549)
-    drawSolidPoly(vertices, 4, color);
-    _isConvex = false;
+    const float coef = 2.0f * (float)M_PI / segments;
+
+    Vec2* _vertices = new Vec2[segments + 2]; //_abuf.get<Vec2>(segments + 2);
```
halx99: ax::pod_vector  is recommended for alloc temp buffer

core/2d/DrawNode.cpp
```diff
@@ -503,288 +460,852 @@ void DrawNode::drawCardinalSpline(PointArray* config, float tension, unsigned in
...
+
+    // Its a circle?
+    if (MAX((startAngle - endAngle), (endAngle - startAngle)) == DEGREES)
+    {
+        _circle = true;
+    }
 
     const float coef = 2.0f * (float)M_PI / DEGREES;
-    Vec2* vertices   = _abuf.get<Vec2>(DEGREES + 2);
+    Vec2* vertices = new Vec2[DEGREES + 2]; //_abuf.get<Vec2>(DEGREES + 2);
```
halx99: consider use `ax::pod_vector`

core/2d/DrawNode.cpp
```diff
@@ -805,148 +1326,112 @@ void DrawNode::drawPie(const Vec2& center,
...
+            vert[i].x = vert[i].x * properties->scale.x + properties->position.x;
+            vert[i].y = vert[i].y * properties->scale.y + properties->position.y;
+        }
     }
-    else
+
+    if (closedCounter)
     {
-        Node::visit(renderer, parentTransform, parentFlags);
+        vert[count] = vert[0];
```
halx99: merge 2 line to 1 line: `vert[count++]` = vert[0];`

halx99 (changes requested):

core/2d/DrawNode.cpp
aismann: done

core/2d/DrawNode.cpp
aismann: done

core/2d/DrawNode.cpp
aismann: I use ax::any_buffer

core/2d/DrawNode.cpp
aismann: done

core/2d/DrawNode.cpp
aismann: i use ax::any_buffer

core/2d/DrawNode.cpp
aismann: done

core/2d/DrawNode.cpp
aismann: i use ax::any_buffer

core/2d/DrawNode.cpp
aismann: done

core/2d/DrawNode.cpp
aismann: done

core/2d/DrawNode.cpp
aismann: done

aismann:
Opps.
@halx99 
How can I restore it again?
It was my mistake to closed it.

halx99:
on your dev branch:
```
git reset --hard e6ebe6e315207adbc8b2498a37feb746990f3491
git push -f
```

PR #2100, opened 2024-08-24
Remove the NS_AX_BEGIN and NS_AX_END macros.

j-jorge:
Following #2093 I made these changes that actually remove these two macros. This is the result of a scripted search and replace, so maybe there is incorrect replacements that I will fix if someone point them :)


halx99 (approved):

core/base/Config.h
```diff
@@ -346,15 +346,9 @@ THE SOFTWARE.
 /// @name namespace ax
 /// @{
 #ifdef __cplusplus
-#    define NS_AX_BEGIN \
-        namespace ax    \
-        {
-#    define NS_AX_END }
```
smilediver, line 352: I think these should have been just deprecated and left for backwards compatibility. I mean just those macro definitions.

PR #2102, opened 2024-08-26
Node: fix `setPosition()` and `setPositionNormalized()` not switching normalized mode if positions match

smilediver:
This fixes two bugs where `Node::setPosition()` and `Node::setPositionNormalized()` doesn't turn on or off normalized position mode if position being set is same as current position.

### DIFF
--- a/core/2d/Node.cpp
+++ b/core/2d/Node.cpp
@@ -522,7 +522,7 @@ void Node::getPosition(float* x, float* y) const
 
 void Node::setPosition(float x, float y)
 {
-    if (_position.x == x && _position.y == y)
+    if (_position.x == x && _position.y == y && !_usingNormalizedPosition)
         return;
 
     _position.x = x;
@@ -587,7 +587,7 @@ const Vec2& Node::getPositionNormalized() const
 /// position setter
 void Node::setPositionNormalized(const Vec2& position)
 {
-    if (_normalizedPosition.equals(position))
+    if (_normalizedPosition.equals(position) && _usingNormalizedPosition)
         return;
 
     _normalizedPosition      = position;
--- a/tests/unit-tests/CMakeLists.txt
+++ b/tests/unit-tests/CMakeLists.txt
@@ -43,6 +43,8 @@ set(GAME_SOURCE
     Source/AppDelegate.cpp
     Source/TestUtils.cpp
 
+    Source/core/2d/NodeTests.cpp
+
     Source/core/base/MapTests.cpp
     Source/core/base/UTF8Tests.cpp
     Source/core/base/UtilsTests.cpp
--- a/tests/unit-tests/Source/core/2d/NodeTests.cpp
+++ b/tests/unit-tests/Source/core/2d/NodeTests.cpp
@@ -1,0 +1,62 @@
+/****************************************************************************
+ Copyright (c) 2019-present Axmol Engine contributors (see AUTHORS.md).
+
+ https://axmol.dev/
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in
+ all copies or substantial portions of the Software.
+
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ THE SOFTWARE.
+ ****************************************************************************/
+
+#include <doctest.h>
+#include <float.h>
+#include "2d/Node.h"
+
+using namespace ax;
+
+TEST_SUITE("2d/Node") {
+    TEST_CASE("normalized_position") {
+        auto parent = Node();
+        auto node = Node();
+
+        parent.setContentSize(Vec2(200.0f, 100.0f));
+        node.setParent(&parent);
+        CHECK_EQ(0.0f, node.getPosition().x);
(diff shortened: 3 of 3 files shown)


rh101:
This PR has uncovered an issue in `ui::Widget`, where child nodes using normalized positioning would end up at the wrong position.  The issue seems to start in `Widget::setContentSize(const Vec2& contentSize)`, where the final call in that method is to `onSizeChanged();`, which in turn ends up calling `updateSizeAndPosition()` on all child nodes.  

Example code to reproduce the issue:
```
auto baseLayout = ax::ui::Layout::create();
baseLayout->setContentSize(Vec2(400, 400));
contentLayout->setBackGroundColorType(ax::ui::Layout::BackGroundColorType::SOLID);
contentLayout->setBackGroundColor(ax::Color3B::RED);
baseLayout->setLayoutType(ax::ui::Layout::Type::ABSOLUTE);
this->addChild(baseLayout);

auto contentLayout = ax::ui::Layout::create();
contentLayout->setLayoutType(ax::ui::Layout::Type::RELATIVE);
contentLayout->setAnchorPoint(ax::Vec2::ANCHOR_MIDDLE_TOP);
contentLayout->setPositionNormalized(ax::Vec2(0.5f, 1.f));
contentLayout->setBackGroundColorType(ax::ui::Layout::BackGroundColorType::SOLID);
contentLayout->setBackGroundColor(ax::Color3B::GREEN);
baseLayout->addChild(contentLayout);

contentLayout->setContentSize(Vec2(200, 200)); // this triggers the issue
```

The problem is this section:

```
void Widget::updateSizeAndPosition(const Vec2& parentSize)
{
...
        Vec2 absPos = getPosition();
        switch (_positionType)
        {
        case PositionType::ABSOLUTE:
        {
            if (parentSize.width <= 0.0f || parentSize.height <= 0.0f)
            {
                _positionPercent.setZero();
            }
            else
            {
                _positionPercent.set(absPos.x / parentSize.width, absPos.y / parentSize.height);
            }
            break;
        }
        case PositionType::PERCENT:
        {
            absPos.set(parentSize.width * _positionPercent.x, parentSize.height * _positionPercent.y);
            break;
        }
        default:
            break;
        }
        setPosition(absPos);
...
```

On the line `Vec2 absPos = getPosition();`, the actual position is still `0,0`, since the normalized position has not yet been calculated for that widget, so when it gets to `setPosition(absPos);`, it effectively overwrites the position with `0,0`, and sets `_usingNormalizedPosition` to `false`.  So, when it finally gets to this section during rendering, the real normalized position is never used, since `_usingNormalizedPosition` is `false`:

```
uint32_t Node::processParentFlags(const Mat4& parentTransform, uint32_t parentFlags)
{
    if (_usingNormalizedPosition)
    {
        AXASSERT(_parent, "setPositionNormalized() doesn't work with orphan nodes");
        if ((parentFlags & FLAGS_CONTENT_SIZE_DIRTY) || _normalizedPositionDirty)
        {
            auto& s           = _parent->getContentSize();
            _position.x       = _normalizedPosition.x * s.width;
            _position.y       = _normalizedPosition.y * s.height;
            _transformUpdated = _transformDirty = _inverseDirty = true;
            _normalizedPositionDirty                            = false;
        }
    }
...
}
```

Right now I can think of 2 ways to fix this. One is to add logic to detect if the widget is using normalized positioning, and if so, use those methods instead, as follows:
```
void Widget::updateSizeAndPosition(const Vec2& parentSize)
{
...
    // update position & position percent
    if (_usingNormalizedPosition)
    {
        Vec2 absPos = getNormalizedPosition();
        switch (_positionType)
        {
        case PositionType::ABSOLUTE:
        {
            if (parentSize.width <= 0.0f || parentSize.height <= 0.0f)
            {
                _positionPercent.setZero();
            }
            else
            {
                _positionPercent.set(absPos.x, absPos.y);
            }
            break;
        }
        case PositionType::PERCENT:
        {
            absPos.set(_positionPercent.x, _positionPercent.y);
            break;
        }
        default:
            break;
        }
        setPositionNormalized(absPos);
    }
    else
    {
        Vec2 absPos = getPosition();
        switch (_positionType)
        {
        case PositionType::ABSOLUTE:
        {
            if (parentSize.width <= 0.0f || parentSize.height <= 0.0f)
            {
                _positionPercent.setZero();
            }
            else
            {
                _positionPercent.set(absPos.x / parentSize.width, absPos.y / parentSize.height);
            }
            break;
        }
        case PositionType::PERCENT:
        {
            absPos.set(parentSize.width * _positionPercent.x, parentSize.height * _positionPercent.y);
            break;
        }
        default:
            break;
        }
        setPosition(absPos);
    }
}

// and...

void Widget::setPositionPercent(const Vec2& percent)
{
    if (_usingLayoutComponent)
    {
        auto component = this->getOrCreateLayoutComponent();
        component->setPositionPercentX(percent.x);
        component->setPositionPercentY(percent.y);
        component->refreshLayout();
    }
    else
    {
        _positionPercent = percent;
        if (_running)
        {
            Widget* widgetParent = getWidgetParent();
            if (widgetParent)
            {
                Vec2 parentSize = widgetParent->getContentSize();
                if (_usingNormalizedPosition)
                {
                    Vec2 absPos(_positionPercent.x, _positionPercent.y);
                    setPositionNormalized(absPos);
                }
                else
                {
                    Vec2 absPos(parentSize.width * _positionPercent.x, parentSize.height * _positionPercent.y);
                    setPosition(absPos);
                }
            }
        }
    }
}
```

Or, alternatively, we can force the calculation of the position from the normalized values prior to using them, and turn off normalized positioning.  This change would only affect `Widget::updateSizeAndPosition`, with no need to update `Widget::setPositionPercent`:

```
void Widget::updateSizeAndPosition(const Vec2& parentSize)
{
    ...

    Vec2 absPos;
    if (_usingNormalizedPosition && _normalizedPositionDirty)
    {
        absPos.x = _normalizedPosition.x * parentSize.width;
        absPos.y = _normalizedPosition.y * parentSize.height;
    }
    else
    {
        absPos = getPosition();
    }

    // update position & position percent
    switch (_positionType)
    {
    case PositionType::ABSOLUTE:
    {
        if (parentSize.width <= 0.0f || parentSize.height <= 0.0f)
        {
            _positionPercent.setZero();
        }
        else
        {
            _positionPercent.set(absPos.x / parentSize.width, absPos.y / parentSize.height);
        }
        break;
    }
    case PositionType::PERCENT:
    {
        absPos.set(parentSize.width * _positionPercent.x, parentSize.height * _positionPercent.y);
        break;
    }
    default:
        break;
    }
    setPosition(absPos);
}
```

So, the first solution maintains normalized positioning, while the alternative keeps the current behavior, which turns normalized positioning off, but now calculates the values correctly. 

Which is preferred? First or second solution? Or does anyone have an alternative suggestion?

smilediver:
This should get a separate issue.

My first question would be, is it valid to call `setPositionNormalized()` on the widget? It seems that it has its own relative layout mechanism through `setPositionPercent()`.

rh101:
I'll create an issue for this.  As for the validity of calling `setPositionNormalized()`, it's still part of the public interface.  It's a little strange though that it wasn't used for the purpose of the `percentage`, since it seems to represent the same values.

~~Perhaps that is another alternative fix, just overriding `setPositionNormalized()`, and then calling `setPositionPercent()` from it.~~ (Never mind, this won't fix the issue.)

PR #2103, opened 2024-08-26
Remove the USING_NS_AX and NS_AX macros.

j-jorge:
Similar to #2100, this PR removes some unnecessary macros, namely `USING_NS_AX` and `NS_AX`. 

A [comment](https://github.com/axmolengine/axmol/pull/2100#discussion_r1731253889) from @smilediver in the aforementioned PR suggests to deprecate these macros and keep their definition (while still removing their usage). I don't know of any way to mark a macro as deprecated, so I'm afraid that nothing would prevent a programmer to not use them except a line in the release notes. On the other hand I see how such a radical removal could break user code. I have no preference so I would need an authority to tell me if I should restore the definitions :)


rh101:
`USING_NS_AX` is a convenience macro that would currently be used in most projects, so this change is definitely going to break things.  The macro itself should not be used within the engine code, but it may be better to keep the define as it is for the sake of developers using it.

`NS_AX` is a pointless macro, and definitely should be removed.

With `NS_AX_BEGIN` and `NS_AX_END`, the primary use was in the engine code, and it would be rare to use these in a game project, so there is no point for them to remain defined at all.  If a game project was using them, then the impact would be minor, as I can't imagine why a developer would want to add too many things to the axmol namespace in their project.

j-jorge:
Thanks @rh101 :) I have updated the PR following your feedback.

rh101:
Just FYI, section missed for `NS_AX` in `Macros.h`:
image

Issue #2104, opened 2024-08-27
Incorrect output when using custom shaders


rh101:
- axmol version: 15268dc182a4499d965ace98271a89f089d9c55d
- devices test on:
- developing environments
   - NDK version: r23c
   - Xcode version: 14.2+
   - Visual Studio: 
      - VS version: 2022 (17.9+)
      - MSVC version: 19.39+
      - Windows SDK version: 10.0.22621.0+
   - cmake version: 
Steps to Reproduce:

1. Create 4 sprites, and assign the same custom fragment shader to each one.  The custom shader simply changes the color of the sprite based on a supplied value.  
2. The rendered sprites all have the same color, based on the *last* shader program state created, so it is somehow being batched when it should not be, since the color value passed to the shader is different for all sprites.

Expected:
![image](https://github.com/user-attachments/assets/27ce2bac-c66c-4a5a-95cc-6b67d990981b)

Actual:
![image](https://github.com/user-attachments/assets/5c93a298-fbf7-47f2-b231-b57313c07f30)

The sprites are added in order from top left to bottom right, so they should be Red, Green, Blue, Orange. As can be seen in the actual output, the orange color is the last one set, so is being applied to all sprites.  The number of GL calls is 1, which means the sprites are being batched, when they should not be. The correct GL calls should be 4.

This was working in previous versions of Axmol, and I think it may have changed after this commit :
https://github.com/axmolengine/axmol/commit/2d0a63660fe082f2f797702db807d2169901a6c2

I think I have an idea on how to fix it, so should be able to get a PR up soon.

rh101:
PR #2105 shows the expected output (with 4 draw calls):
![image](https://github.com/user-attachments/assets/75485574-354d-48e9-ae55-205ae56776b9)

rh101:
It seems that once a uniform value is updated, `ProgramState::updateBatchId()` needs to be called in order to update the batch ID, and that in turn fixes the issue.  This also ensures that same-value uniforms are batched together.

smilediver:
I think it's a bug. You should not be required to call `updateBatchId()`.

rh101:
The batch ID is equal to the program ID if it is using one of the built in shaders:
```
bool ProgramState::init(Program* program)
{
...
    const auto programId = program->getProgramId();
    if (programId < ProgramType::BUILTIN_COUNT)
        this->_batchId = programId
...
}
```

[...]

PR #2105, opened 2024-08-27
Fix custom shader output

rh101:
## Describe your changes

The batch ID and program ID should be `uint64_t`, but in places they were being used as `int64_t`, causing incorrect results of conditional statements that check if batching should be enabled.

This fixes the issue of incorrect output when the same custom shader is used on multiple sprites.

#2104 

### For each PR
       
   Optional:
          
### For core/new feature PR


rh101:
The changes made in this PR may not be correct, and are not actually required, and the correct solution is to simply call `ProgramState::updateBatchId()` when the uniforms are updated for a specific program state.

PR #2115, opened 2024-08-28
Math optimizations

smilediver:
Mostly moves some trivial and commonly used methods in math classes to headers so that they can be inlined by compiler.


paulocoutinhox:
What is inl file?

halx99 (changes requested):
**I prefer just implement inline APIs in class define directly, don't create individual .inl file if the inline APIs are public**

core/math/Color.inl
halx99, line 1: don't create individual .inl file

core/math/Mat4.inl
halx99, line 1: don't create individual .inl file

core/math/Rect.inl
halx99, line 1: don't create individual .inl file

core/math/Vec4.inl
halx99, line 1: don't create individual .inl file

core/math/Quaternion.inl
halx99, line 1: move to class define body

smilediver:
Should I put definitions with declarations like so:

```cpp
class Color {
    void method() { doStuff(); }
};
```
or keep declarations clean and put definitions at the end of file like this:
```cpp
class Color {
    void method();
};

void Color::method() { doStuff(); }
```
or maybe something in between, like keep shorter definitions with declarations, but keep larger definitions at the end of file?

halx99:
I prefer first one, less code

halx99 (approved):

PR #2136, opened 2024-09-11
Allow turn on/off deprecated features

### LINKED ISSUE #2137: Deprecation policy
### Discussed in https://github.com/axmolengine/axmol/discussions/2111

<div type='discussions-op-text'>

<sup>Originally posted by **smilediver** August 28, 2024</sup>
It would be nice if there was a more clear deprecation policy and more robust management. My proposed plan is:

1. Decide what the lifetime for deprecated features should be. Example: deprecated features are removed after they were deprecated for at least two minor releases: so if it was marked deprecated in 2.1, then it is removed in 2.3.
3. Document that in Readme/Wiki.
4. Mark everything already deprecated with `AX_DEPRECATED(2.1)`. (Or maybe we should already remove old deprecated stuff right now?, cause some stuff is already deprecated since Cocos).
5. Add `AX_DEPRECATED_NOTE(...)` (name could be different) for marking things like macros, that can't have a deprecated attribute.
6. Remove `AX_DEPRECATED_ATTRIBUTE`.
7. Add `AX_USE_DEPRECATED` to turn on/off deprecated features. It can be optionally turned on by user, so they can choose to stay on the bleeding edge or use deprecated features. It would work like this:
```cpp
    #if AX_USE_DEPRECATED
        AX_DEPRECATED(2.1) static Application* sharedApplication();
    #endif
```

Also, it would be nice if there was a more clear overall release schedule, like a new minor release every quarter, or something like that.

What do you think? If there's interest to that, I could take care of the code side of things.</div>

halx99:
## Describe your changes

- Add new cmake option: `AX_CORE_PROFILE`
- Remove deprecated stubs from extension: `cocostudio`
- Remove `AX_DEPRECATED_ATTRIBUTE`

https://github.com/axmolengine/axmol/discussions/2111

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/cocos2d.h
+++ b/core/cocos2d.h
@@ -25,10 +25,14 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
+
+!!! The cocos2d-x compatibility header, please reserve this file, because some
+extensions use it, i.g fairygui, live2d ...
+
 ****************************************************************************/
 
-#ifndef __COCOS2D_H__
-#define __COCOS2D_H__
+#ifndef __AX_COCOS2DX_H__
+#define __AX_COCOS2DX_H__
 
 #include "axmol.h"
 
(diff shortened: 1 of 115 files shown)


solan-solan (approved):
Looks as hard but good decision

aismann:
How will future "AX_DEPRECATED_ATTRIBUTE" features handle?

If I have AX_DISABLE_DEPRECATED = 'true' => all new defined "AX_DEPRECATED_ATTRIBUTE" features will be "removed" directly without planed version?

Ok I have this reading to late:
Deprecation policy #2137
Can be a solution.

core/cocos2d.h
```diff
@@ -60,10 +60,12 @@ namespace cocos2d = ax;
 #define CC_ASSERT                    AX_ASSERT
 #define CC_CONSTRUCTOR_ACCESS        public
 
+#ifndef AX_DISABLE_DEPRECATED
```
smilediver: Maybe everything in this header should be disabled by `AX_DISABLE_DEPRECATED`?

halx99: Many references from extensions, and refactor 3rd extensions is time-wasting

PR #2138, opened 2024-09-12
Fix for incorrect bits per pixel set on texture reload

rh101:
## Describe your changes

~~The pixel format of the FPS texture was being set incorrectly, which ended up crashing Android applications after attempting to recover from a GL context loss.~~

~~The change made simply fixes the pixel format of that texture entry in the texture cache.~~

The true cause of the issue was the bits per pixel setting of the cached texture being overwritten during recovery from the GL context loss.

#2126 

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/base/Director.cpp
+++ b/core/base/Director.cpp
@@ -1325,7 +1325,7 @@ void Director::createStatsLabel()
         return;
     }
 
-    texture = _textureCache->addImage(image, "/ax_fps_images", PixelFormat::RGBA4);
+    texture = _textureCache->addImage(image, "/ax_fps_images", image->getPixelFormat());
     AX_SAFE_RELEASE(image);
 
     /*
--- a/core/base/Director.cpp
+++ b/core/base/Director.cpp
@@ -1325,7 +1325,7 @@ void Director::createStatsLabel()
         return;
     }
 
-    texture = _textureCache->addImage(image, "/ax_fps_images", image->getPixelFormat());
+    texture = _textureCache->addImage(image, "/ax_fps_images", PixelFormat::RGBA4);
     AX_SAFE_RELEASE(image);
 
     /*
--- a/core/renderer/Texture2D.cpp
+++ b/core/renderer/Texture2D.cpp
@@ -334,7 +334,7 @@ bool Texture2D::updateWithMipmaps(MipmapInfo* mipmaps,
     }
 
 #if AX_ENABLE_CACHE_TEXTURE_DATA
-    VolatileTextureMgr::findVolotileTexture(this);
+    VolatileTextureMgr::getOrAddVolatileTexture(this);
 #endif
 
     backend::TextureDescriptor textureDescriptor;
--- a/core/renderer/TextureCache.cpp
+++ b/core/renderer/TextureCache.cpp
@@ -861,7 +861,7 @@ void VolatileTextureMgr::addImage(Texture2D* tt, Image* image)
     }
 }
 
-VolatileTexture* VolatileTextureMgr::findVolotileTexture(Texture2D* tt)
+VolatileTexture* VolatileTextureMgr::getOrAddVolatileTexture(Texture2D* tt)
 {
     VolatileTexture* vt = nullptr;
     for (const auto& texture : _textures)
--- a/core/renderer/TextureCache.cpp
+++ b/core/renderer/TextureCache.cpp
@@ -838,7 +838,7 @@ void VolatileTextureMgr::addImageTexture(Texture2D* tt, std::string_view imageFi
         return;
     }
 
-    VolatileTexture* vt = findVolotileTexture(tt);
+    VolatileTexture* vt = getOrAddVolatileTexture(tt);
 
     vt->_cashedImageType = VolatileTexture::kImageFile;
     vt->_fileName        = imageFileName;
@@ -850,7 +850,7 @@ void VolatileTextureMgr::addImage(Texture2D* tt, Image* image)
     if (tt == nullptr || image == nullptr)
         return;
 
-    VolatileTexture* vt = findVolotileTexture(tt);
+    VolatileTexture* vt = getOrAddVolatileTexture(tt);
 
     if(vt->_uiImage != image) {
         AX_SAFE_RELEASE(vt->_uiImage);
@@ -894,7 +894,7 @@ void VolatileTextureMgr::addDataTexture(Texture2D* tt,
         return;
     }
 
-    VolatileTexture* vt = findVolotileTexture(tt);
+    VolatileTexture* vt = getOrAddVolatileTexture(tt);
 
     vt->_cashedImageType = VolatileTexture::kImageData;
     vt->_textureData     = data;
@@ -910,7 +910,7 @@ void VolatileTextureMgr::addStringTexture(Texture2D* tt, std::string_view text, 
         return;
     }
 
-    VolatileTexture* vt = findVolotileTexture(tt);
+    VolatileTexture* vt = getOrAddVolatileTexture(tt);
 
     vt->_cashedImageType = VolatileTexture::kString;
     vt->_text            = text;
--- a/core/renderer/TextureCache.h
+++ b/core/renderer/TextureCache.h
@@ -313,7 +313,8 @@ public:
 
     // find VolatileTexture by Texture2D*
     // if not found, create a new one
-    static VolatileTexture* findVolotileTexture(Texture2D* tt);
+    AX_DEPRECATED(2.2) VolatileTexture* findVolotileTexture(Texture2D* tt) { return getOrAddVolatileTexture(tt); }
+    VolatileTexture* getOrAddVolatileTexture(Texture2D* tt);
 
 private:
     static void reloadTexture(Texture2D* texture, std::string_view filename, backend::PixelFormat pixelFormat);
--- a/core/renderer/TextureCache.h
+++ b/core/renderer/TextureCache.h
@@ -313,8 +313,8 @@ public:
 
     // find VolatileTexture by Texture2D*
     // if not found, create a new one
-    AX_DEPRECATED(2.2) VolatileTexture* findVolotileTexture(Texture2D* tt) { return getOrAddVolatileTexture(tt); }
-    VolatileTexture* getOrAddVolatileTexture(Texture2D* tt);
+    AX_DEPRECATED(2.2) static VolatileTexture* findVolotileTexture(Texture2D* tt) { return getOrAddVolatileTexture(tt); }
+    static VolatileTexture* getOrAddVolatileTexture(Texture2D* tt);
 
 private:
     static void reloadTexture(Texture2D* texture, std::string_view filename, backend::PixelFormat pixelFormat);
--- a/core/renderer/backend/Texture.cpp
+++ b/core/renderer/backend/Texture.cpp
@@ -36,8 +36,13 @@ void TextureBackend::updateTextureDescriptor(const ax::backend::TextureDescripto
     _textureType   = descriptor.textureType;
     _textureFormat = descriptor.textureFormat;
     _textureUsage  = descriptor.textureUsage;
-    _width         = descriptor.width;
-    _height        = descriptor.height;
+    _width        = (std::max)(descriptor.width, (uint32_t)1);
+    _height       = (std::max)(descriptor.height, (uint32_t)1);
+
+    if (_bitsPerPixel == 0)
+    {
+        _bitsPerPixel = (uint8_t)(8 * 4);
+    }
 }
 
 NS_AX_BACKEND_END
--- a/core/renderer/backend/opengl/TextureGL.cpp
+++ b/core/renderer/backend/opengl/TextureGL.cpp
@@ -168,17 +168,9 @@ void Texture2DGL::initWithZeros()
     // FIXME: Don't call at Texture2DGL::updateTextureDescriptor, when the texture is compressed, initWithZeros will
     // cause GL Error: 0x501 We call at here once to ensure depth buffer works well. Ensure the final data size at least
     // 4 byte
-
-    _width        = (std::max)(_width, (uint32_t)1);
-    _height       = (std::max)(_height, (uint32_t)1);
-
-    constexpr auto maxBitsPerPixel = (uint8_t)(8 * 4);
-    if (_bitsPerPixel == 0)
-    {
-        _bitsPerPixel = maxBitsPerPixel;
-    }
-
     auto size     = _width * _height * _bitsPerPixel / 8;
+    assert(size > 0);
+
     uint8_t* data = (uint8_t*)malloc(size);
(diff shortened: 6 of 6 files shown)


rh101:
Note this may not be the right fix it the intention is to actually use `RGBA4` as the pixel format, and the error may exist elsewhere.  It only happens when the texture is being recreated after a GL context loss:

```
void VolatileTextureMgr::reloadAllTextures()
{
...
    for (auto&& texture : _textures)
    {
        VolatileTexture* vt = texture;

        switch (vt->_cashedImageType)
        {
...
        case VolatileTexture::kImage:
        {
            vt->_texture->initWithImage(vt->_uiImage, vt->_pixelFormat);  // this call
        }
        break;
```

Which ends up here:
```
bool Texture2D::updateWithImage(Image* image, backend::PixelFormat format, int index)
{
    if (image == nullptr)
    {
        __AXLOGWITHFUNCTION("axmol: Texture2D. Can't create Texture. UIImage is nil");
        return false;
    }

    if (this->_filePath.empty())
        this->_filePath = image->getFilePath();

    int imageWidth  = image->getWidth();
    int imageHeight = image->getHeight();

    Configuration* conf = Configuration::getInstance();

    int maxTextureSize = conf->getMaxTextureSize();
    if (imageWidth > maxTextureSize || imageHeight > maxTextureSize)
    {
        AXLOGW("axmol: WARNING: Image ({} x {}) is bigger than the supported {} x {}", imageWidth, imageHeight,
              maxTextureSize, maxTextureSize);
        return false;
    }

    unsigned char* tempData               = image->getData();
    // Vec2 imageSize                        = Vec2((float)imageWidth, (float)imageHeight);
    backend::PixelFormat renderFormat     = (PixelFormat::NONE == format) ? image->getPixelFormat() : format;
    backend::PixelFormat imagePixelFormat = image->getPixelFormat();
    size_t tempDataLen                    = image->getDataLen();

#ifdef AX_USE_METAL
    //! override renderFormat, since some render format is not supported by metal
    switch (renderFormat)
    {
#    if (AX_TARGET_PLATFORM != AX_PLATFORM_IOS || TARGET_OS_SIMULATOR)
    // packed 16 bits pixels only available on iOS
    case PixelFormat::RGB565:
    case PixelFormat::RGB5A1:
    case PixelFormat::RGBA4:
#    endif
    case PixelFormat::R8:
    case PixelFormat::RG8:
    case PixelFormat::RGB8:
        // Note: conversion to RGBA8 will happends
        renderFormat = PixelFormat::RGBA8;
        break;
    default:
        break;
    }
#elif !AX_GLES_PROFILE
    // Non-GLES doesn't support follow render formats, needs convert PixelFormat::RGBA8
    // Note: axmol-1.1 deprecated A8, L8, LA8 as renderFormat, preferred R8, RG8
    switch (renderFormat)
    {
    case PixelFormat::R8:
    case PixelFormat::RG8:
        // Note: conversion to RGBA8 will happends
        renderFormat = PixelFormat::RGBA8;
    }
#endif

    if (image->getNumberOfMipmaps() > 1)
    {
        if (renderFormat != image->getPixelFormat())
        {
            AXLOGW("WARNING: This image has more than 1 mipmaps and we will not convert the data format");
        }

        // pixel format of data is not converted, renderFormat can be different from pixelFormat
        // it will be done later
        updateWithMipmaps(image->getMipmaps(), image->getNumberOfMipmaps(), image->getPixelFormat(), renderFormat, imageHeight, imageWidth, image->hasPremultipliedAlpha(), index);
    }
    else if (image->isCompressed())
    {  // !Only hardware support texture will be compression PixelFormat, otherwise, will convert to RGBA8 duraing image
       // load
        renderFormat = imagePixelFormat;
        updateWithData(tempData, tempDataLen, image->getPixelFormat(), image->getPixelFormat(), imageWidth, imageHeight, image->hasPremultipliedAlpha(), index);
    }
    else
    {
        // after conversion, renderFormat == pixelFormat of data
        updateWithData(tempData, tempDataLen, imagePixelFormat, renderFormat, imageWidth, imageHeight,
                       image->hasPremultipliedAlpha(), index);
    }

    return true;
}
```

The problem may be in `updateWithMipmaps`, which is called from `updateWithData`.  

Looking into this some more to try and figure out why it's happening, since forcing `RGBA4` for the FPS texture makes sense.

rh101:
The true problem seems to have something to do with the `_bitsPerPixel`, where it is 32 when it should be 16:

image

When the FPS texture is first created using pixel format `RGBA4`, everything is fine, and the `_bitsPerPixel` is 16, so the calculation for bytes per row is correct.  On recovery from the GL context loss, `_bitsPerPixel` is never set correctly, which results in an incorrect value for the `GL_UNPACK_ALIGNMENT`, and the crash.

rh101:
The problem has been traced to this method:
```
void Texture2DGL::initWithZeros()
{
    // !!!Only used for depth stencil render buffer
    // For example, a texture used as depth buffer will not invoke updateData(), see cpp-tests 'Effect Basic/Effects
    // Advanced'.
    // FIXME: Don't call at Texture2DGL::updateTextureDescriptor, when the texture is compressed, initWithZeros will
    // cause GL Error: 0x501 We call at here once to ensure depth buffer works well. Ensure the final data size at least
    // 4 byte

    _width        = (std::max)(_width, (uint32_t)1);
    _height       = (std::max)(_height, (uint32_t)1);
    _bitsPerPixel = (std::max)(_bitsPerPixel, (uint8_t)(8 * 4));

    auto size     = _width * _height * _bitsPerPixel / 8;
    uint8_t* data = (uint8_t*)malloc(size);
    memset(data, 0, size);
    updateData(data, _width, _height, 0);
    free(data);
}
```

`_bitsPerPixel` is being set to the max of that statement, so while it may be 16, it ends up changing to 32, which causes the problem.

The `Texture2DGL::initWithZeros` method is called from here when it is recovering from a GL context loss:
```
Texture2DGL::Texture2DGL(const TextureDescriptor& descriptor)
{
    updateTextureDescriptor(descriptor);

#if AX_ENABLE_CACHE_TEXTURE_DATA
    // Listen this event to restored texture id after coming to foreground on Android.
    _rendererRecreatedListener = EventListenerCustom::create(EVENT_RENDERER_RECREATED, [this](EventCustom*) {
        _textureInfo.onRendererRecreated(GL_TEXTURE_2D);
        this->initWithZeros();
    });
    Director::getInstance()->getEventDispatcher()->addEventListenerWithFixedPriority(_rendererRecreatedListener, -1);
#endif
}
```

The Cocos2d-x implementation was this, which never changed the value of the `_bitsPerElement`, and why it never caused the problem:
```
void Texture2DGL::initWithZeros()
{
    auto size = _width * _height * _bitsPerElement / 8;
    uint8_t* data = (uint8_t*)malloc(size);
    memset(data, 0, size);
    updateData(data, _width, _height, 0);
    free(data);
}
```

@halx99 How should this be solved? Should we just use `std::min` for the statement?

```
_bitsPerPixel = (std::min)(_bitsPerPixel, (uint8_t)(8 * 4));
```

or alternatively, avoid setting it if it is already has a valid value (greater than 0, and less or equal to 32):
```
constexpr auto maxBitsPerPixel = (uint8_t)(8 * 4);
if (_bitsPerPixel > maxBitsPerPixel || _bitsPerPixel == 0)
{
    _bitsPerPixel = maxBitsPerPixel;
}
```

rh101:
I've used this as the fix now, unless there are better suggestions:
```
constexpr auto maxBitsPerPixel = (uint8_t)(8 * 4);
if (_bitsPerPixel > maxBitsPerPixel || _bitsPerPixel == 0)
{
    _bitsPerPixel = maxBitsPerPixel;
}
```

This will preserve a valid `_bitsPerPixel` value.  The app no longer crashes on recovery from the GL context loss.

smilediver:
This doesn't seem right, why `initWithZeros()` is modifying `_width`, `_height` or `_bitsPerPixel` at all? Also floating point textures can have more than 32 bits.

halx99:
Maybe use renderbuffer instead texture for depth stencil buffer, no need initWithZeros

rh101:
> This doesn't seem right, why `initWithZeros()` is modifying `_width`, `_height` or `_bitsPerPixel` at all? Also floating point textures can have more than 32 bits.

I'm not sure why the code was changed from the Cocos2d-x implementation, which never adjusted those values.

If floating point texture can have a more than 32 bits, then I'll change the conditional to simply check for 0.

halx99:
better solution is clamp width, height, bitsPerPexel at TextureBackend::updateTextureDescriptor, and don't modify them in initWithZeros

rh101:
> better solution is clamp width, height, bitsPerPexel at TextureBackend::updateTextureDescriptor, and don't modify them in initWithZeros

That can be done, but I'm concerned about this "FIXME" comment in the code:
```
void Texture2DGL::initWithZeros()
{
    // !!!Only used for depth stencil render buffer
    // For example, a texture used as depth buffer will not invoke updateData(), see cpp-tests 'Effect Basic/Effects
    // Advanced'.
    // FIXME: Don't call at Texture2DGL::updateTextureDescriptor, when the texture is compressed, initWithZeros will
    // cause GL Error: 0x501 We call at here once to ensure depth buffer works well. Ensure the final data size at least
    // 4 byte

    _width        = (std::max)(_width, (uint32_t)1);
    _height       = (std::max)(_height, (uint32_t)1);
```

Is this what you were thinking:

```
void TextureBackend::updateTextureDescriptor(const ax::backend::TextureDescriptor& descriptor, int /*index*/)
{
    _bitsPerPixel  = PixelFormatUtils::getBitsPerPixel(descriptor.textureFormat);
    _textureType   = descriptor.textureType;
    _textureFormat = descriptor.textureFormat;
    _textureUsage  = descriptor.textureUsage;
    _width        = (std::max)(descriptor.width, (uint32_t)1);
    _height       = (std::max)(descriptor.height, (uint32_t)1);

    if (_bitsPerPixel == 0)
    {
        _bitsPerPixel = (uint8_t)(8 * 4);
    }
}
```
and 
```
void Texture2DGL::initWithZeros()
{
    // !!!Only used for depth stencil render buffer
    // For example, a texture used as depth buffer will not invoke updateData(), see cpp-tests 'Effect Basic/Effects
    // Advanced'.
    // FIXME: Don't call at Texture2DGL::updateTextureDescriptor, when the texture is compressed, initWithZeros will
    // cause GL Error: 0x501 We call at here once to ensure depth buffer works well. Ensure the final data size at least
    // 4 byte
    auto size     = _width * _height * _bitsPerPixel / 8;
    uint8_t* data = (uint8_t*)malloc(size);
    memset(data, 0, size);
    updateData(data, _width, _height, 0);
    free(data);
}
```

halx99:
The comment should be `The initWIthZeros is only used for RenderTarget attachments`. I'm not sure do we need invoke it when android GLES context resume from lost

halx99:
> > better solution is clamp width, height, bitsPerPexel at TextureBackend::updateTextureDescriptor, and don't modify them in initWithZeros
> 
> That can be done, but I'm concerned about this "FIXME" comment in the code:
> 
> ```
> void Texture2DGL::initWithZeros()
> {
>     // !!!Only used for depth stencil render buffer
>     // For example, a texture used as depth buffer will not invoke updateData(), see cpp-tests 'Effect Basic/Effects
>     // Advanced'.
>     // FIXME: Don't call at Texture2DGL::updateTextureDescriptor, when the texture is compressed, initWithZeros will
>     // cause GL Error: 0x501 We call at here once to ensure depth buffer works well. Ensure the final data size at least
>     // 4 byte
> 
>     _width        = (std::max)(_width, (uint32_t)1);
>     _height       = (std::max)(_height, (uint32_t)1);
> ```
> 
> Is this what you were thinking:
> 
> ```
> void TextureBackend::updateTextureDescriptor(const ax::backend::TextureDescriptor& descriptor, int /*index*/)
> {
>     _bitsPerPixel  = PixelFormatUtils::getBitsPerPixel(descriptor.textureFormat);
>     _textureType   = descriptor.textureType;
>     _textureFormat = descriptor.textureFormat;
>     _textureUsage  = descriptor.textureUsage;
>     _width        = (std::max)(descriptor.width, (uint32_t)1);
>     _height       = (std::max)(descriptor.height, (uint32_t)1);
> 
>     if (_bitsPerPixel == 0)
>     {
>         _bitsPerPixel = (uint8_t)(8 * 4);
>     }
> }
> ```
> 
> and
> 
> ```
> void Texture2DGL::initWithZeros()
> {
>     // !!!Only used for depth stencil render buffer
>     // For example, a texture used as depth buffer will not invoke updateData(), see cpp-tests 'Effect Basic/Effects
>     // Advanced'.
>     // FIXME: Don't call at Texture2DGL::updateTextureDescriptor, when the texture is compressed, initWithZeros will
>     // cause GL Error: 0x501 We call at here once to ensure depth buffer works well. Ensure the final data size at least
>     // 4 byte
>     auto size     = _width * _height * _bitsPerPixel / 8;
>     uint8_t* data = (uint8_t*)malloc(size);
>     memset(data, 0, size);
>     updateData(data, _width, _height, 0);
>     free(data);
> }
> ```

correct

smilediver:
Why is this clamping necessary? Shouldn't these be asserts, since If you have a texture with 0 width, height or bits per pixel, then I'm certain it's a bug?

rh101:
> The comment should be `The initWIthZeros is only used for RenderTarget attachments`. I'm not sure do we need invoke it when android GLES context resume from lost

I'm not sure if it needs to be called on the context loss either.

By the way, would you mind if I make the following change too:

```
AX_DEPRECATED_ATTRIBUTE VolatileTexture* findVolotileTexture(Texture2D* tt) { return getOrAddVolatileTexture(tt); }
VolatileTexture* getOrAddVolatileTexture(Texture2D* tt);
```

`findVolotileTexture` should be "findVolatileTexture", but that method name does not represent what it actually does, which is to add a new entry if it does not find one.  I suggest we deprecate the old method name (or delete it) and use `getOrAddVolatileTexture`.  What do you think about that?

halx99:
> > The comment should be `The initWIthZeros is only used for RenderTarget attachments`. I'm not sure do we need invoke it when android GLES context resume from lost
> 
> I'm not sure if it needs to be called on the context loss either.
> 
> By the way, would you mind if I make the following change too:
> 
> ```
> AX_DEPRECATED_ATTRIBUTE VolatileTexture* findVolotileTexture(Texture2D* tt) { return getOrAddVolatileTexture(tt); }
> VolatileTexture* getOrAddVolatileTexture(Texture2D* tt);
> ```
> 
> `findVolotileTexture` should be "findVolatileTexture", but that method name does not represent what it actually does, which is to add a new entry if it does not find one. I suggest we deprecate the old method name (or delete it) and use `getOrAddVolatileTexture`. What do you think about that?

sure, but please use `AX_DEPRECATED(2.2)` or just rename for internal API without deprecated mark.

halx99 (approved):

Issue #1411, opened 2023-11-01
DrawNode: setLineWitdh() is no longer working


aismann:
Steps to Reproduce:
cpp-tests->31:Node:Draw>GitHub Issue #11942 (axmol #137)"

Wrong:  Latest build:
![image](https://github.com/axmolengine/axmol/assets/8652787/a94dce8e-f398-4de5-87dd-7fc300897ff1)

Works: Thats a older version
![image](https://github.com/axmolengine/axmol/assets/8652787/8bef719b-26fe-4689-bc8e-480c182256c2)

halx99:
Not recommand use setLineWith because the render backend opengl & metal have diferrent behavior:

- GLES2: support glLineWidth
- GLES3: only support glLineWidth (0~1)
- Metal: not support lineWidth

joprice:
I hit this as well and found a workaround by taking the `drawSolidCircle` code and calling `drawPolygon` instead of`drawSolidPoly` and passing thickness along:

```c++
 static void drawCircle(DrawNode &node, const Vec2 &center, float radius, float angle, int segments,
                         float scaleX, float scaleY, const Color4F &color, float thickness) {
    const float coef = 2.0f * (float)M_PI / static_cast<float>(segments);

    Vec2 *vertices = new (std::nothrow) Vec2[segments];
    if (vertices == nullptr)
      return;

    for (unsigned int i = 0; i < segments; i++) {
      float rads = static_cast<float>(i) * coef;
      float j = radius * cosf(rads + angle) * scaleX + center.x;
      float k = radius * sinf(rads + angle) * scaleY + center.y;

      vertices[i].x = j;
      vertices[i].y = k;
    }
    node.drawPolygon(vertices, segments, Color4F(), thickness, color);

    CC_SAFE_DELETE_ARRAY(vertices);
  }
```

Might be of help to someone, or could perhaps become a new helper if it seems useful enough.

PR #2165, opened 2024-09-19
Optimize DrawNode

smilediver:
This PR does some DrawNode optimizations:
* Disabled transform by default, which had significant performance impact by default.
* Use `pod_vector` for storing and growing buffers.
* Don't update actual vertex buffer every time `drawXXX()` is called.
* Avoid unnecessary allocations on some `drawXXX()` methods in `_transform()`.

There's still a lot of `drawXXX()` methods that do unnecessary allocations that have a big impact on performance.

Also there are `_lineWidth` and `_defaultLineWidth` members that can be set by the user but do nothing. Seems like a bug that probably breaks old behavior.

### DIFF
--- a/core/2d/DrawNode.cpp
+++ b/core/2d/DrawNode.cpp
@@ -733,19 +733,19 @@ void DrawNode::drawSolidCircle(const Vec2& center,
     _drawCircle(center, radius, angle, segments, false, 1.0f, 1.0f, Color4B(), color, true);
 }
 
-void DrawNode::drawTriangle(const Vec2* _vertices3, const Color4B& color)
+void DrawNode::drawTriangle(const Vec2* vertices3, const Color4B& color)
 {
-    _drawTriangle(_vertices3, Color4B::TRANSPARENT, color, false, 0.0f);
+    Vec2 vertices[3] = {vertices3[0], vertices3[1], vertices3[2]};
+    _drawTriangle(vertices, Color4B::TRANSPARENT, color, false, 0.0f);
 }
 
 void DrawNode::drawTriangle(const Vec2& p1, const Vec2& p2, const Vec2& p3, const Color4B& color)
 {
-
-    Vec2 _vertices3[3] = {p1, p2, p3};
-    _drawTriangle(_vertices3, Color4B::TRANSPARENT, color, false, 0.0f);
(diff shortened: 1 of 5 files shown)


aismann:
> There's still a lot of `drawXXX()` methods that do unnecessary allocations that have a big impact on performance.
> 
@smilediver 
Can you add comments like `// To do: Improve: unnecessary allocations ...` or better can you improve this methods too?

> Also there are `_lineWidth` and `_defaultLineWidth` members that can be set by the user but do nothing. Seems like a bug that probably breaks old behavior.

See here:
[DrawNode: setLineWitdh() is no longer working](https://github.com/axmolengine/axmol/issues/1411#top)#1411
See here for more info:
https://github.com/axmolengine/axmol/issues/1411#issuecomment-1788550519

@halx99 @smilediver:
 `setLineWidth()` should be deprecated.

smilediver:
> Can you add comments like `// To do: Improve: unnecessary allocations ...` or better can you improve this methods too?

It would be better if you did that, since you wrote it and know the code better.

> `setLineWidth()` should be deprecated.

Maybe it's better to remove it completely, since it's not doing anything anyway, and will mislead people.

aismann:
> > Can you add comments like `// To do: Improve: unnecessary allocations ...` or better can you improve this methods too?
> 
> It would be better if you did that, since you wrote it and know the code better.
> 
> > `setLineWidth()` should be deprecated.
> 
> Maybe it's better to remove it completely, since it's not doing anything anyway, and will mislead people.

I will look after merge.

halx99 (approved):

halx99:
@aismann could we delete https://github.com/axmolengine/axmol/tree/dev/extensions/DrawNodeEx/src/DrawNodeEx?

aismann:
@smilediver (@halx99)
Can you explain me what this means:
`    bool _trianglesDirty: 1 = false;`

aismann:
> @aismann could we delete https://github.com/axmolengine/axmol/tree/dev/extensions/DrawNodeEx/src/DrawNodeEx?

I read to late, yes can deleted.

smilediver:
> Can you explain me what this means: ` bool _trianglesDirty: 1 = false;`

It's a bit field: https://en.cppreference.com/w/cpp/language/bit_field

PR #2185, opened 2024-09-26
`AxmolActivity` refactoring and fixes

smilediver:
This PR refactors and simplifies how the state of the app activity is managed and also:

* Fixes `nativeOnResume()` not being called after a system popup is displayed. (Easy repro: turn on bluetooth in quick settings menu and wait couple seconds for a popup to show up, then close it).
* Fix `AxmolRenderer` forgetting paused state when `AxmolActivity` is relaunched, during which the old objects are destroyed and recreated. (Easy repro: in Android settings change global font size).

I have tested this extensively on several devices, and everything seems to work. But it would be nice if more people tested it.

### DIFF
--- a/core/platform/android/java/src/org/axmol/lib/AxmolGLSurfaceView.java
+++ b/core/platform/android/java/src/org/axmol/lib/AxmolGLSurfaceView.java
@@ -37,6 +37,8 @@ import android.view.MotionEvent;
 import android.view.View;
 import android.view.inputmethod.InputMethodManager;
 
+import java.util.concurrent.CountDownLatch;
+
 public class AxmolGLSurfaceView extends GLSurfaceView {
     // ===========================================================
     // Constants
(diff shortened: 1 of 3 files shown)


AlexandreK38:
I also faced the popup issue and started a fix but saw this PR; I tried and it seems to work fine

core/platform/android/java/src/org/axmol/lib/AxmolGLSurfaceView.java
```diff
@@ -427,6 +401,38 @@ public void run() {
...
+            }
+        });
+    }
+
+    public void waitForPauseToComplete() {
+        while (mNativePauseComplete.getCount() > 0) {
+            try {
+                mNativePauseComplete.await();
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
```
halx99: Maybe just ignore the exception since you wait while latch count == 0

halx99 (changes requested):

core/platform/android/java/src/org/axmol/lib/AxmolGLSurfaceView.java
smilediver: Yes, nice catch! It was added by autocomplete and I missed it...

halx99:
lgtm, merged

PR #2218, opened 2024-10-23
Update to locate Microsoft Build Tools

martinking71:
Locate build tool only installs, for when no full Visual Studio install

## Describe your changes
Updated command line option to locate BuildTool only installs of the Visual Studio Build tools - for when there is no full install of visual studio - CI server or when running other IDE's like Jetbrains.


halx99:
will fixed by #2252

PR #2262, opened 2024-12-10
Remove "old" example files (not used and cpp-tests/ParticleTest has the "same" tests)

aismann:
## Describe your changes
Remove (not used and test examples be part of the cpp-tests/ParticleTest too):
```
core/2d/ParticleExamples.h
core/2d/ParticleExamples.cpp
```

### For each PR
       
   Optional:
          
### For core/new feature PR


aismann:
Opps. Too early submitted

rh101:
Are you sure it's a good idea to remove these files?  They're not actually examples, but implementations of common particle effects that developers may make use of in their applications.

The filename "ParticleExamples" may just be a little misleading.

aismann:
> Are you sure it's a good idea to remove these files? They're not actually examples, but implementations of common particle effects that developers may use. The filename "ParticleExamples" may just be a little misleading.

Yes my foult, the file name did me going in a wrong direction.

PR #2275, opened 2024-12-14
Add printLeaks call in appropriate location to help developers

rh101:
## Describe your changes

If the `printLeaks()` is called in the wrong location, then it could lead to confusion, since it would be reporting memory leaks that aren't actually valid leaks.  The changes in this PR just ensure that it is called in an appropriate location, and enabled by the existing `AX_OBJECT_LEAK_DETECTION` define.

### For each PR
       
   Optional:
          
### For core/new feature PR


grif-on:
Ultra speed 👍 
Thanks again .

PR #2284, opened 2024-12-24
fix crash when close app (PC platform)

NgVThangBz:
on PC platform. when using (X) to close,
`director->mainLoop();` right after `director->end();` seems redundant and causes a crash,
after removing `director->mainLoop();` my application works fine,
no crashes because of the order of deleting objects

### DIFF
--- a/core/platform/linux/Application-linux.cpp
+++ b/core/platform/linux/Application-linux.cpp
@@ -93,7 +93,6 @@ int Application::run()
     if (glView->isOpenGLReady())
     {
         director->end();
-        director->mainLoop();
         director = nullptr;
     }
     glView->release();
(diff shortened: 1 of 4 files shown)


rh101:
> on PC platform. when using (X) to close, `director->mainLoop();` right after `director->end();` seems redundant and causes a crash

The current changes in this PR are not correct.  

`Director::end()` simply does the following:

```
void Director::end()
{
    _cleanupDirectorInNextLoop = true;
}
```

It requires a call to `Director::mainLoop()` in order to actually clean-up and free all objects etc..

While the changes in this PR may seem to fix the issue for you, it is more than likely that it is covering up the true source of the crash. 

> after removing `director->mainLoop();` my application works fine, no crashes because of the order of deleting objects

At a guess, the issue is in your project code.  

Submitting a PR like this without actually diagnosing the problem and providing detailed information makes little sense.

PR #2291, opened 2024-12-26
Fix renderer test case batching

rh101:
## Describe your changes

In cpp-tests -> Renderer -> RendererUniformBatch test, the drawn sprites should be batched, but they are not, so instead of 9 draw calls, it results in over 80.  The changes in this PR ensure that the batch ID of each of the custom shaders is calculated so that the sprites are correctly batched, which fixes the issue.

The other related issue is this:
```
bool ProgramState::init(Program* program)
{
...
    const auto programId = program->getProgramId();
    if (programId < ProgramType::BUILTIN_COUNT)
        this->_batchId = programId;

    return true;
}
```

`Program::getProgramId()` was returning an `int64_t`, which could return a negative number, resulting in the conditional statement being true, when it should not be:
```
if (programId < ProgramType::BUILTIN_COUNT)
```

Unless I'm mistaken, all Program IDs are `uint64_t`, so the result of `Program::getProgramId()` has been updated accordingly, along with the key types of the `std::unordered_map` used to store the `Program` instances.

Added `ProgramState::isValidBatchId()` to check if a valid batch ID has been set.  The reason is that the hashing methods can return any value, even `uint64_t(-1)` (please correct me if I'm wrong on this):

`_batchId = XXH64(_uniformBuffers.data(), _uniformBuffers.size(), _program->getProgramId());`

So, doing a check like the following may result in incorrectly skipping batching:
```
if (_batchId == -1)
    setSkipBatching(true);
```

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/renderer/TrianglesCommand.cpp
+++ b/core/renderer/TrianglesCommand.cpp
@@ -68,7 +68,7 @@ void TrianglesCommand::init(float globalOrder,
         blendDescriptor.sourceRGBBlendFactor = blendDescriptor.sourceAlphaBlendFactor = blendType.src;
         blendDescriptor.destinationRGBBlendFactor = blendDescriptor.destinationAlphaBlendFactor = blendType.dst;
 
-        if (_batchId == -1)
+        if (!_pipelineDescriptor.programState->isValidBatchId())
             setSkipBatching(true);
 
         if (!isSkipBatching())
--- a/core/renderer/TrianglesCommand.cpp
+++ b/core/renderer/TrianglesCommand.cpp
@@ -68,7 +68,7 @@ void TrianglesCommand::init(float globalOrder,
         blendDescriptor.sourceRGBBlendFactor = blendDescriptor.sourceAlphaBlendFactor = blendType.src;
         blendDescriptor.destinationRGBBlendFactor = blendDescriptor.destinationAlphaBlendFactor = blendType.dst;
 
-        if (!_pipelineDescriptor.programState->isValidBatchId())
+        if (!_pipelineDescriptor.programState->isBatchable())
             setSkipBatching(true);
 
         if (!isSkipBatching())
--- a/core/renderer/backend/Program.h
+++ b/core/renderer/backend/Program.h
@@ -152,7 +152,7 @@ public:
      * Get program id.
      * @return The program id.
      */
-    int64_t getProgramId() const { return _programId; }
+    uint64_t getProgramId() const { return _programId; }
 
     /**
      * Get uniform buffer size in bytes that can hold all the uniforms.
--- a/core/renderer/backend/ProgramManager.h
+++ b/core/renderer/backend/ProgramManager.h
@@ -157,9 +157,9 @@ protected:
     };
 
     BuiltinRegInfo _builtinRegistry[(int)backend::ProgramType::BUILTIN_COUNT];
-    std::unordered_map<int64_t, BuiltinRegInfo> _customRegistry;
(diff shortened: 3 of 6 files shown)


halx99:
does the changes cause custom program can't be batched?

halx99:
`isValidBatchId` --> `isBatchable` better?

rh101:
> does the changes cause custom program can't be batched?

The behavior does not change, but this PR fixes the logic in the `ProgramState::init(Program* program)`.

From what I understand, the program ID for built-in programs is just their `ProgramType` enum (uint32_t) value.  This value is greater than 0, and less than `ProgramType::BUILTIN_COUNT` (`0x1000`).

In the following code, it checks the returned program ID, and if it is within the range 0 to `ProgramType::BUILTIN_COUNT`, it sets the batch ID to the same as the program ID, so it is batched.
```
bool ProgramState::init(Program* program)
{
...
    const auto programId = program->getProgramId();
    if (programId < ProgramType::BUILTIN_COUNT)
        this->_batchId = programId;

    return true;
}
```

The problem though was for custom programs, where the program ID is calculated by the following:
```
uint64_t ProgramManager::computeProgramId(std::string_view vsName, std::string_view fsName)
{
    XXH64_reset(_programIdGen, 0);
    XXH64_update(_programIdGen, vsName.data(), vsName.length());
    XXH64_update(_programIdGen, fsName.data(), fsName.length());
    return XXH64_digest(_programIdGen);
}
```

That program ID result is an unsigned 64 bit value, and if I understand how the hash works, it can be anything from 0 to 2^64.

The problem was that the program ID was being returned as a signed 64 bit value, so values about `2^63` would result in negatives:
`int64_t getProgramId() const { return _programId; }`

This negative value was being used in the conditional statement `if (programId < ProgramType::BUILTIN_COUNT)`, so the result would be `true`, which is incorrect, based on the context of that check.  This means *some* ProgramState instances were being batched, but unexpectedly so, and even then, they may have been incorrectly batched (especially if they used different uniform data).

I presume that to be the reason why `ProgramState::updateBatchId()` was added, which meant developers are in control of updating the batch ID for program states that may have different uniform data.

The changes in this PR only fix the issue of `if (programId < ProgramType::BUILTIN_COUNT)` not being logically correct.  Custom programs will not be batched unless `ProgramState::updateBatchId()` is called on the ProgramState instances.

Would you prefer if we do have custom programs batched by default, and still allow developers to adjust the batch ID when required (for different uniform data)?  All we would need to change this:
```
bool ProgramState::init(Program* program)
{
...
    const auto programId = program->getProgramId();
    if (programId < ProgramType::BUILTIN_COUNT)
        this->_batchId = programId;

    return true;
}
```
To this:
```
bool ProgramState::init(Program* program)
{
...
    this->_batchId = program->getProgramId();
    return true;
}
```

This will automatically batch the ProgramState, and if the same program is used, but with different uniform data, then `ProgramState::updateBatchId()` would still need to be used.  This is left up to the developer, as it is right now.  So, the only difference is that custom programs will be auto-batched, compared to the current implementation, where they are not.  Do you want this change too?

halx99:
> This will automatically batch the ProgramState, and if the same program is used, but with different uniform data, then `ProgramState::updateBatchId()` would still need to be used. This is left up to the developer, as it is right now. So, the only difference is that custom programs will be auto-batched, compared to the current implementation, where they are not. Do you want this change too?

Let user control custom program batch fully may be better, so don't change

halx99 (approved):

PR #2318, opened 2025-01-09
Distinguishing between Simplified Chinese and Traditional Chinese in the Win32 Platform

MartinLutherSu:
The Win32 platform will now support distinguishing between Simplified and Traditional Chinese, allowing developers to build PC games in both languages for their players.


halx99 (approved):

halx99:
Thanks, I do merge to `dev` and `3.x` manually

PR #2319, opened 2025-01-09
Update DevSetup.md for Android

rh101:
## Describe your changes

The instructions in the DevSetup state that r23c is the minimum, which may make developer assume that it would be set as the default, which is not the case, since the default configured NDK is r23d in https://github.com/axmolengine/axmol/blob/dev/1k/build.profiles#L46

To avoid this confusion, instructions have been updated to ensue developers run the `setup.ps1 -p android` in order to download NDK r23d, just so that they can get up and running with a new project without issues.

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/docs/DevSetup.md
+++ b/docs/DevSetup.md
@@ -14,8 +14,9 @@
 
   1. Download or clone Axmol from GitHub ([https://github.com/axmolengine/axmol](https://github.com/axmolengine/axmol)).
   2. Enter `axmol` root directory.
-  3. Run `setup.ps1` in windows powershell or (macOS/Linux/ArchLinux terminal). Restart the console after it has finished for environment variables to take effect.
-  4. Ensure that the C / C++ compiler toolset is installed on your host machine.
+  3. Run `setup.ps1` in windows powershell or (macOS/Linux/ArchLinux terminal).
+  4. Restart the console after it has finished for environment variables to take effect.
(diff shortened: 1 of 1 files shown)


rh101:
@halx99 There may be an issue on MacOS. If `setup.ps1` is run first, then `setup.ps1 -p android` is run afterwards, it does not download the NDK.  It's happened to a few developers so far, and all seem to be using MacOS.  Could there be something that is preventing the command from running correctly (such as a cached value etc.)?

halx99:
> @halx99 There may be an issue on MacOS. If `setup.ps1` is run first, then `setup.ps1 -p android` is run afterwards, it does not download the NDK. It's happened to a few developers so far, and all seem to be using MacOS. Could there be something that is preventing the command from running correctly (such as a cached value etc.)?

I will take a look

rh101:
Sorry, I didn't realise it created the branch in this repo instead mine, so I'll pay more attention to that next time.

halx99:
> > @halx99 There may be an issue on MacOS. If `setup.ps1` is run first, then `setup.ps1 -p android` is run afterwards, it does not download the NDK. It's happened to a few developers so far, and all seem to be using MacOS. Could there be something that is preventing the command from running correctly (such as a cached value etc.)?
> 
> I will take a look

I can't reproduce it

rh101:
> I can't reproduce it

OK! Then it may just be a configuration issue in their own environment.

PR #2321, opened 2025-01-10
Add support for the creation of layers that are set to invisible in tilemap

### LINKED ISSUE #2320: Saving tilemap (.tmx) having a layer visibility off, makes that layer nullptr in axmol.
- axmol version: 2.2.1
- devices test on: Windows 10
- developing environments
   - Visual Studio: 
      - VS version: 2022 (17.9+)

Steps to Reproduce:
1. Create a tile map with Tiled, add a layer and set  its visibility off. You can use this project https://github.com/rudiHammad/tilemap-axmol-test/tree/main
2. replace the existing init code with this init:
```
auto tileMap = new ax::FastTMXTiledMap();
tileMap->initWithTMXFile("tilesets/myOtherTilemap.tmx");   
tileset myTileset.tsx
auto layer = tileMap->getLayer("Tile Layer 1");
layer->setGlobalZOrder(false);
addChild(tileMap);
```
The error happens in `layer->setGlobalZOrder(false);` since layer is considered nullptr.
In a tile map , the collision layer(s) visibility are usually off. So currently we are forced to have all the visibilities on in Tiled, and then
set them off in Axmol.

rh101:
## Describe your changes
The current implementation does not allow the creation of layer nodes from layers that are set to invisible in the tilemap file.  There is no way to create them, so functionality has been added to allow invisible layers to be created.  

The default option is that they are not created, so it does not change existing behavior.

#2320 

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/2d/FastTMXTiledMap.cpp
+++ b/core/2d/FastTMXTiledMap.cpp
@@ -35,10 +35,10 @@ namespace ax
 
 // implementation FastTMXTiledMap
 
-FastTMXTiledMap* FastTMXTiledMap::create(std::string_view tmxFile)
+FastTMXTiledMap* FastTMXTiledMap::create(std::string_view tmxFile, bool allowInvisibleLayers)
 {
     FastTMXTiledMap* ret = new FastTMXTiledMap();
-    if (ret->initWithTMXFile(tmxFile))
+    if (ret->initWithTMXFile(tmxFile, allowInvisibleLayers))
     {
         ret->autorelease();
         return ret;
@@ -47,10 +47,12 @@ FastTMXTiledMap* FastTMXTiledMap::create(std::string_view tmxFile)
     return nullptr;
 }
 
-FastTMXTiledMap* FastTMXTiledMap::createWithXML(std::string_view tmxString, std::string_view resourcePath)
+FastTMXTiledMap* FastTMXTiledMap::createWithXML(std::string_view tmxString,
+                                                std::string_view resourcePath,
+                                                bool allowInvisibleLayers)
 {
     FastTMXTiledMap* ret = new FastTMXTiledMap();
-    if (ret->initWithXML(tmxString, resourcePath))
+    if (ret->initWithXML(tmxString, resourcePath, allowInvisibleLayers))
     {
         ret->autorelease();
         return ret;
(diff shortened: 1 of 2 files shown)


rudiHammad:
That works for me . Thanks

rudiHammad:
Hello again.
I've been testing this a bit more. It does work, but it turns out that if the layer which visibility is off,  is used for tile collisions, then this errors. int FastTMXLayer::getTileGIDAt(const Vec2& tileCoordinate, TMXTileFlags* flags /* = nullptr*/).
So don't worry about it, I think everybody else has no issue with this so it's okey.
Sorry about that, I was setting the visibility off in the wrong layers.

rh101:
> Hello again.
> I've been testing this a bit more. It does work, but it turns out that if the layer which visibility is off, is used for tile collisions, then this errors. int FastTMXLayer::getTileGIDAt(const Vec2& tileCoordinate, TMXTileFlags* flags /* = nullptr*/).
> So don't worry about it, I think everybody else has no issue with this so it's okey.
> Sorry about that, I was setting the visibility off in the wrong layers.

You edited your post to add the last sentence, but left everything else in there as it is, so overall it made no sense.  If something has changed, then just leave a new comment, but otherwise if you're changing the issue, then use something like the strike-out text option to remove the text that no longer applies, ~~like this~~, or just some indicator that the rest of the text is no longer valid/no longer applies.

So, from your post above, the only part that is relevant is the "Sorry about that, I was setting the visibility off in the wrong layers.", but what does that mean, is there still an issue or not?  If there is a crash, then be more specific about the crash, and include a lot more information, including how to reproduce it.

rudiHammad:
In the last sentence I meant to say: "sorry about that, there is still the same issue, but I thought it was solved because I was testing the wrong layers. "
Sorry if I wasn't clear.

I don't want to be a bother with this. So we can close this issue and leave as it was. Meaning as it was by default before the changes.

rh101:
> I don't want to be a bother with this. So we can close this issue and leave as it was. Meaning as it was by default before the changes.

The PR doesn't impact existing behavior, and still gives the developers the option of creating tile layers that are turned off in the input TMX file, so there is no need to roll back the changes.

PR #2327, opened 2025-01-11
fix drawLine wrong when thickness!=1

NgVThangBz:
fix drawLine with thickness other than 1
factor = (FrameSize /designResolutionSize)*2
and the smallest thickness that axmol can create is 1, 
so in all cases where the thickness is <1 I think we should draw a line with thickness = 1

See detailed error at #2324

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/2d/DrawNode.cpp
+++ b/core/2d/DrawNode.cpp
@@ -1058,7 +1058,7 @@ void DrawNode::_drawSegment(const Vec2& from,
     Vec2 vertices[2] = {from, to};
     applyTransform(vertices, vertices, 2);
 
-    if (thickness == 1.0f && !properties.drawOrder)
+    if (thickness <= 1.0f && !properties.drawOrder)
     {
         auto line = expandBufferAndGetPointer(_lines, 2);
         _linesDirty = true;
@@ -1068,12 +1068,16 @@ void DrawNode::_drawSegment(const Vec2& from,
     }
     else
     {
+        float fS = Director::getInstance()->getGLView()->getFrameSize().width;
+        float dS = Director::getInstance()->getGLView()->getDesignResolutionSize().width;
+        float factor = 2 * fS / dS;
+
         Vec2 a  = vertices[0];
         Vec2 b  = vertices[1];
         Vec2 n  = ((b - a).getPerp()).getNormalized();
         Vec2 t  = n.getPerp();
-        Vec2 nw = n * thickness;
-        Vec2 tw = t * thickness;
+        Vec2 nw = n * thickness/ factor;
+        Vec2 tw = t * thickness/ factor;
         Vec2 v0 = b - (nw + tw);
         Vec2 v1 = b + (nw - tw);
         Vec2 v2 = b - nw;


aismann:
@NgVThangBz 
Please move this to the constructor of DrawNode.cpp
```
        _factor2 = 2 * AX_CONTENT_SCALE_FACTOR(); 
```
DrawNode .h:
`float _factor2;` 

Maybe this is better:
```
    float width =  thickness / _factor2;
    Vec2 nw = n * width ;
    Vec2 tw = t * width ;
```
Please also add a cpp tester  which test different factors or at least the
correct draw of `Label->enableUnderline();`

aismann:
@NgVThangBz 
If you give me some more time I will make a PR for this issue (So you can close your PR when you want)
Currently writing the cpp-tests tester for at least Label->enableUnderline();
image

rh101:
> @NgVThangBz Please move this to the constructor of DrawNode.cpp
> 
> ```
>         _factor2 = 2 * AX_CONTENT_SCALE_FACTOR(); 
> ```

@aismann If the content scale factor is changed at runtime (such as due to screen resolution changes), then the above cannot be in the constructor.  If it changes after the DrawNode is created, the resulting output will be incorrect.

NgVThangBz:
@aismann factor != AX_CONTENT_SCALE_FACTOR()
factor = 2*FrameSize()/DesignResolutionSize()


        float fS = Director::getInstance()->getGLView()->getFrameSize().width;

        float dS = Director::getInstance()->getGLView()->getDesignResolutionSize().width;

        float factor = 2 * fS / dS;

and as @rh101  mentioned, if set in constructor it will not be accurate

rh101:
@NgVThangBz It may be better to do as @aismann suggested, which is to close this PR and let him submit one.  The reason is that @aismann is very familiar with the DrawNode implementation, and a change like this has a wide impact, so it may require more test cases, which you have not added to this PR.

aismann:
> @aismann factor != AX_CONTENT_SCALE_FACTOR() factor = 2*FrameSize()/DesignResolutionSize()
> 
> ```
>     float fS = Director::getInstance()->getGLView()->getFrameSize().width;
> 
>     float dS = Director::getInstance()->getGLView()->getDesignResolutionSize().width;
> 
>     float factor = 2 * fS / dS;
> ```
> 
> and as @rh101 mentioned, if set in constructor it will not be accurate

@NgVThangBz 
cpp-tests use '*.height' as input values:
`           director->setContentScaleFactor(g_resourceSize.height / g_designSize.height);`

You switch to "width":
>     float fS = Director::getInstance()->getGLView()->getFrameSize().width;
>     float dS = Director::getInstance()->getGLView()->getDesignResolutionSize().width;

What is your reason to use 'width' instead of 'height'?

I think your solution is sometimes not a correct value because the factor between 
     getFrameSize().width / getDesignResolutionSize().width  
     and 
     getFrameSize().height/ getDesignResolutionSize().height
     can be totaly different.
     I belive define/changing the factor is part of the developer.
     And DrawNode should add a getFactor/setFactor API  and a default value = (AX_CONTENT_SCALE_FACTOR()*0.5f).

NgVThangBz:
I will close my PR, and  @aismann will continue

PR #2273, opened 2024-12-13
Improve build profiles

halx99:
1. Add `1k/build.profiles` for android ndk, sdk tools, emsdk ...
2. Android targetSdk and minSdk also control by 1k/build.profiles
3. You can override some build profiles in `.axproj` file, i.e android ndk, android target_sdk and min_sdk and others

### For each PR
- [ ] Add Copyright if it missed:   
      - `"Copyright (c) 2019-present Axmol Engine contributors (see AUTHORS.md)."`
- [ ] I have performed a self-review of my code.
       
   Optional:
   - [ ] I have checked readme and add important infos to this PR.
   - [ ] I have added/adapted some tests too.
          
### For core/new feature PR
- [ ] I have checked readme and add important infos to this PR.
- [ ] I have added thorough tests.

PR #2361, opened 2025-02-01
[Android] Allow project to specify Axmol engine root path

rh101:
Allow Android project to specify Axmol engine path in case the engine is embedded into the project.

The path is now read from `System.properties.AX_ROOT` first instead of `System.env.AX_ROOT`.  If the `System.properties.AX_ROOT` value is not set, then it will default to using `System.env.AX_ROOT` in `axutils.gradle`.

This may help solve the issue mentioned in this [post](https://github.com/axmolengine/axmol/pull/2273#issuecomment-2626991359)

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/platform/android/libaxmol/axutils.gradle
+++ b/core/platform/android/libaxmol/axutils.gradle
@@ -178,11 +178,19 @@ class AxmolUtils {
         return buildProfiles
     }
 
+    private static getEngineRoot() {
+        def axRoot = System.getProperty("AX_ROOT");
+        if (axRoot == null) {
+            axRoot = System.getenv("AX_ROOT");
+        }
+        return axRoot;
+    }
+
     private static Properties loadProfiles(project, isAxmolAppProj) {
         // build.profiles in axmol engine
         def profiles = new Properties()
         try {
-            profiles.load(new File(Paths.get("${System.env.AX_ROOT}/1k/build.profiles").toUri()).newDataInputStream())
+            profiles.load(new File(Paths.get("${getEngineRoot()}/1k/build.profiles").toUri()).newDataInputStream())
         }
         catch(ignored) {
         }
@@ -311,7 +319,7 @@ class AxmolUtils {
             cmakeBinDirs.add(cmakeBinDir)
         }
 
-        cmakeBinDir = joinPath(Paths.get("${System.env.AX_ROOT}").toAbsolutePath().toString(), 'tools', 'external', 'cmake', 'bin')
+        cmakeBinDir = joinPath(Paths.get("${getEngineRoot()}").toAbsolutePath().toString(), 'tools', 'external', 'cmake', 'bin')
         cmakeBinDirs.add(cmakeBinDir)
 
         // find in cmakeBinDirs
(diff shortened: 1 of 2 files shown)


halx99:
where the value of System.properties.AX_ROOT comes from?

rh101:
> where the value of System.properties.AX_ROOT comes from?

From the project `{project_dir}/proj.android/settings.gradle`, where it will have this:

```
def embeddedAxmol = false
if (embeddedAxmol) {
    System.setProperty("AX_ROOT", "${settingsDir}/../axmol")
} else {
    System.setProperty("AX_ROOT", "${System.env.AX_ROOT}")
}

include ':libaxmol'
project(':libaxmol').projectDir = new File(Paths.get("${System.properties.AX_ROOT}/core/platform/android/libaxmol").toUri())
```

If `embeddedAxmol` is set to true, then it will use the custom path that the developer sets, which defaults to `${settingsDir}/../axmol`, being `{project_dir}/axmol`.  That default path is the same as what is used in the `{project_dir}/CMakeLists.txt` when `axmol` folder exists in project:

```
set(_is_axmol_embed FALSE)
if (IS_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/axmol")
    set(_AX_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/axmol")
    set(_is_axmol_embed TRUE)
    message(STATUS "Building isolated project: ${APP_NAME}")
else()
...
endif()
```


`axutils.gradle` won't really be affected by it, since it will always default to `System.env.AX_ROOT` if the `System.properties.AX_ROOT` is not set:
```
private static getEngineRoot() {
    def axRoot = System.getProperty("AX_ROOT");
    if (axRoot == null) {
        axRoot = System.getenv("AX_ROOT");
    }
    return axRoot;
}
```

rh101:
@halx99 Maybe this would be better to match how the `CMakeLists.txt` handles the embedded axmol folder:

```
def folder = new File("${settingsDir}/../axmol")
if (folder.exists()) {
    System.setProperty("AX_ROOT", folder.path)
} else {
    System.setProperty("AX_ROOT", "${System.env.AX_ROOT}")
}
```
instead of:
```
def embeddedAxmol = false
if (embeddedAxmol) {
    System.setProperty("AX_ROOT", "${settingsDir}/../axmol")
} else {
    System.setProperty("AX_ROOT", "${System.env.AX_ROOT}")
}
```

That way no need to manually set `embeddedAxmol`.

halx99:
> @halx99 Maybe this would be better to match how the `CMakeLists.txt` handles the embedded axmol folder:
> 
> ```
> def folder = new File("${settingsDir}/../axmol")
> if (folder.exists()) {
>     System.setProperty("AX_ROOT", folder.path)
> } else {
>     System.setProperty("AX_ROOT", "${System.env.AX_ROOT}")
> }
> ```
> 
> instead of:
> 
> ```
> def embeddedAxmol = false
> if (embeddedAxmol) {
>     System.setProperty("AX_ROOT", "${settingsDir}/../axmol")
> } else {
>     System.setProperty("AX_ROOT", "${System.env.AX_ROOT}")
> }
> ```
> 
> That way no need to manually set `embeddedAxmol`.

Yes, I think so.

AlexandreK38:
@rh101 wow i didn’t expect a fix for this so fast thanks 😅🙏
One question on the code though, if the System.properties AX_ROOT is set in settings.gradle from value or env AX_ROOT why do you check in axutils.gradle again?

rh101:
> One question on the code though, if the System.properties AX_ROOT is set in settings.gradle from value or env AX_ROOT why do you check in axutils.gradle again?

The check in `axutils.gradle` ensures that the changes are transparent, making them backwards compatible with existing projects.  The changes to `settings.gradle` are not mandatory for anyone not needing to have a copy of Axmol within the project.

PR #2367, opened 2025-02-04
Add Android event for Warm Start

AlexandreK38:
## Describe your changes
When an Opengl context is lost, it can be because of the Surface view releasing the context, or the app that is making a Cold Start meaning that a new activity is created along with a new surface view, a new renderer and so on.
In case of a warm start, any UI elements that have been attached to the previous activity are destroyed so we want to be able to know that we have to recreate these elements again (for example a banner ad) 
This PR adds an event `EVENT_APP_WARM_START` that we can listen for in such case.

### For each PR
       
   Optional:
          
### For core/new feature PR


core/platform/android/javaactivity-android.cpp
```diff
@@ -107,7 +107,7 @@ JNIEXPORT void JNICALL Java_dev_axmol_lib_AxmolRenderer_nativeInit(JNIEnv*, jcla
     }
 }
 
-JNIEXPORT void JNICALL Java_dev_axmol_lib_AxmolRenderer_nativeOnContextLost(JNIEnv*, jclass)
+JNIEXPORT void JNICALL Java_org_axmol_lib_AxmolRenderer_nativeOnContextLost(JNIEnv*, jclass, jboolean isColdStart)
```
halx99:
```suggestion
JNIEXPORT void JNICALL Java_dev_axmol_lib_AxmolRenderer_nativeOnContextLost(JNIEnv*, jclass, jboolean isColdStart)
```

AlexandreK38:
I still need to clean one method added by mistake
Update: done

AlexandreK38:
All good for me, thanks for fixing the merge issue @halx99

core/platform/android/java/src/dev/axmol/lib/AxmolRenderer.java
```diff
@@ -77,9 +77,12 @@ public void onSurfaceCreated(final GL10 GL10, final EGLConfig EGLConfig) {
         AxmolRenderer.nativeInit(this.mScreenWidth, this.mScreenHeight);
         this.mLastTickInNanoSeconds = System.nanoTime();
 
+        boolean isColdStart = mNativeInitCompleted;
```
rh101: Is this correct? `mNativeInitComplete` is initially false, which is when the activity is started or recreated (since it's not static), indicating a cold start.  If it's true, it means that it has already been set in a previous run, and hasn't been recreated, which indicates that it is a warm start.  So, should this be:

`boolean isColdStart = !mNativeInitCompleted;`

EDIT: Actually, is the `isColdStart` flag related to the application cold start, or is it just related to a cold start for the activity? It just seems to be a little ambiguous at the moment with regards to the naming of that variable, and the name of the event, `EVENT_APP_COLD_START`, which implies an entire app cold start.

AlexandreK38: I was definitely not well waken up when I did this; you are right, I made a fix. After looking the better definition closer it's a Warm start so I renamed in consequence

core/platform/android/java/src/dev/axmol/lib/AxmolRenderer.java
```diff
@@ -47,9 +47,9 @@ public class AxmolRenderer implements GLSurfaceView.Renderer {
     private long mLastTickInNanoSeconds;
     private int mScreenWidth;
     private int mScreenHeight;
-
     private static boolean gNativeInitialized = false;
     private static boolean gNativeIsPaused = false;
+    private boolean mNativeInitCompleted = false;
```
halx99: the variable `gNativeInitialized` already indicate whether native inited.

AlexandreK38: @halx99 
yeah `gNativeInitialized` is static so at first launch the AppActivity will be created along with the SurfaceView and the Renderer and the value will be true afterward no matter what. 
Now when a Warm start occurs, a new AppActivity will be instanciated, creating a new SurfaceView and of course a new Renderer, which means the static `gNativeInitialized` 
 will still be true but `mNativeInitCompleted` will be false as it's a new instance, that's how we figure out a Warm start as occured.
I renamed the variable to `mSurfaceCreated`

PR #2369, opened 2025-02-04
Image flip Raw data vertically

AlexandreK38:
## Describe your changes

Add a new method `flipRawData()` to Image to flip raw data vertically. As for the `initWithRawData` it only works for RGBA8888 format.

### For each PR
       
   Optional:
          
### For core/new feature PR


core/platform/Image.cpp
```diff
@@ -748,6 +748,30 @@ bool Image::initWithRawData(const uint8_t* data,
     return ret;
 }
 
+void Image::flipRawData()
+{
+    // only RGBA8888 supported
+
```
halx99: should check pixel format

AlexandreK38: I added a check

core/platform/Image.cpp
```diff
@@ -748,6 +748,34 @@ bool Image::initWithRawData(const uint8_t* data,
...
 
+void Image::flipRawData()
+{
+    // only RGBA8888 supported
+    if(_pixelFormat != backend::PixelFormat::RGBA8)
+    {
+        return;
+    }
+    
+    uint8_t temp;
```
halx99, line 760: please correct the code indent

halx99: which editor do you use, the vscode or visual studio should support format axmol engine code by clang-format

halx99: you don't need correct the code indent manually if you use modern vs2022 or vscode, just select the code block, then use correct indent shortcut to auto format them, usually:
visual studio: `Ctrl+K+F`
vscode: context menu: `Format Selection`

core/platform/Image.cpp
```diff
@@ -748,6 +748,34 @@ bool Image::initWithRawData(const uint8_t* data,
...
+    // only RGBA8888 supported
+    if (_pixelFormat != backend::PixelFormat::RGBA8)
+    {
+        return;
+    }
+    
+    uint8_t temp;
+    int idx1, idx2;
+
+    for (int w=0; w<_width; w++)
```
halx99: The code indent style still not match with format config file: `axmol/.clang-format`

core/platform/Image.cpp
aismann: Can you plz replace: 
```
    // only RGBA8888 supported
    if(_pixelFormat != backend::PixelFormat::RGBA8)
    {
        return;
    }
```

with one like this code example:
`    AXASSERT(_pixelFormat == backend::PixelFormat::RGBA8, "only RGBA8888 can be flipped");`

rh101: > Can you plz replace:
> 
> ```
>     // only RGBA8888 supported
>     if(_pixelFormat != backend::PixelFormat::RGBA8)
>     {
>         return;
>     }
> ```
> 
> with one like this code example: ` AXASSERT(_pixelFormat == backend::PixelFormat::RGBA8, "only RGBA8888 can be flipped");`

This is not entirely correct.  AXASSERT would only be valid if `_AX_DEBUG > 0`, and if that is not the case, and there is a possibility that this will result in a crash.  So, if possible, it should still retain a validity check to handle the case when running without AX_DEBUG enabled to avoid crashing the application.  For this method, where it simply flips data, it would be better to have it as something like the following:

```
AXASSERT(_pixelFormat == backend::PixelFormat::RGBA8, "only RGBA8888 can be flipped");
if (_pixelFormat != backend::PixelFormat::RGBA8)
{
    AXLOGE("Cannot flip image. Unsupported pixel format.");
    return;
}
```

halx99 (approved):

PR #2381, opened 2025-02-10
Added methods to modify audio  pitch

Janglee123:
## Describe your changes
Added functions to get/set audio pitch.

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/audio/AudioEngine.cpp
+++ b/core/audio/AudioEngine.cpp
@@ -216,6 +216,28 @@ void AudioEngine::setVolume(AUDIO_ID audioID, float volume)
     }
 }
 
+void AudioEngine::setPitch(AUDIO_ID audioID, float pitch)
+{
+    auto it = _audioIDInfoMap.find(audioID);
+    if (it != _audioIDInfoMap.end())
+    {
+        if (pitch < 0.5f)
+        {
+            pitch = 0.5f;
+        }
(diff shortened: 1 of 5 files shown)


core/audio/AudioEngine.cpp
```diff
@@ -216,6 +216,28 @@ void AudioEngine::setVolume(AUDIO_ID audioID, float volume)
     }
 }
 
+void AudioEngine::setPitch(AUDIO_ID audioID, float pitch)
+{
+    auto it = _audioIDInfoMap.find(audioID);
+    if (it != _audioIDInfoMap.end())
+    {
+        if (pitch < 0.5f)
```
halx99: better code: `pitch = std::clamp(pitch, 0.5f, 2.0f);`

Janglee123: I was following the pattern of [`setVolume`](https://github.com/axmolengine/axmol/blob/53811d6df6ac768763e8e88126085f2329e9b3ae/core/audio/AudioEngine.cpp#L202) method. Let me know if you still want to go with `std::clamp`.

rh101: There is no need to follow the patterns of existing code when it comes to things like this, and any new code should use the best practice at the time it is written.

PR #2204, opened 2024-10-04
3.x draft (for ci check only)

halx99:
## Describe your changes

- [x] Remove all deprecated stubs
- [x] Implement drawArraysInstanced and make instanced draw more common use
- [x] box2d-3.x
    - [x] Make Box2dTest of `cpp-tests` works on box2d v3
    - [x] Make Box2dTestBed of `cpp-tests` works on box2d v3
- [x] Remove chipmunk
    - [x] Remove `core/physics` based on chipmunk
    - [x] Remove ChipmunkTest from `cpp-tests`
    - [x] Remove ChipmuniTestbed from `cpp-tests`
- ~~Change underlaying color type to Color4F aka `ax::Color`~~
- [x] Improve DrawNode, don't create backend::Buffer always
- [x] Upgrade iOS deploy target: 12.0 ==> 13.0
- [x] Upgrade android minsdk: 17(Android 4.2) ==> 21(Android 5.1)
- [x] Upgrade android ndk: r23d ==> r27c
- [x] Due to android,ios toolchain update to support c++20 full features, `std::views::split` can be use to improve performance
- [x] Remove  3rdparty ghc filesystem, since both ios13 and android5.1 support `std::filesystem`
- [x] Build and run wasm64 support, current need enable chrome flag: `chrome://flags/#enable-webassembly-memory64`
- [ ] drop rapidjson
- [ ] re-select json lib: yyjson > boost.json > rapidjson
    simdjson for parse only and simdjson build in-progress: https://github.com/simdjson/experimental_json_builder
    finally solution: `yyjson`  + `simdjson`
- [x] `Color4B` ==> `Color32`
- [ ] Remove `Color3B`
- [ ] drop tolua, use sol2
- [ ] Reimplement core/physics with box2d v3, not all API compatible with axmol-2.x
### For each PR
- [ ] Add Copyright if it missed:   
      - `"Copyright (c) 2019-present Axmol Engine contributors (see AUTHORS.md)."`
- [ ] I have performed a self-review of my code.
       
   Optional:
   - [ ] I have checked readme and add important infos to this PR.
   - [ ] I have added/adapted some tests too.
          
### For core/new feature PR
- [ ] I have checked readme and add important infos to this PR.
- [ ] I have added thorough tests.

## Note

*Currently, the branch `3.x` is in draft mode, that means it's not stable and will force-push frequently*

PR #2387, opened 2025-02-11
Working in-progress v3

halx99:
## Describe your changes

Previous working PR  #2204 was closed due to branch name renamed from `3.x` to `wip-v3`

- [x] ISO c++23 standard
- [x] Texture resource improvment, add texture array support, remove CPU array implement since all graphics API support texture array
- [x] Remove sampler from Texture, implement texture sampler object and cache it, since GLES3/OGL-3.3, mtl, d3d support sampler object
- [x] Update lua to 5.5.x
- [x] Remove the default pixel format, fix rendering of grayscale PNGs by @j-jorge in https://github.com/axmolengine/axmol/pull/2573
- [x] Remove all deprecated stubs
- [x] Implement drawArraysInstanced and make instanced draw more common use
- [x] box2d-3.x
    - [x] Make Box2dTest of `cpp-tests` works on box2d v3
    - [x] Make Box2dTestBed of `cpp-tests` works on box2d v3
- [x] Remove chipmunk
    - [x] Remove `core/physics` based on chipmunk
    - [x] Remove ChipmunkTest from `cpp-tests`
    - [x] Remove ChipmuniTestbed from `cpp-tests`
- [x] Improve DrawNode, don't create backend::Buffer always
- [x] Upgrade iOS deploy target: 12.0 => 13.0
- [x] Upgrade android minsdk: 17(Android 4.2) => 21(Android 5.1)
- [x] Upgrade android ndk: r23 => latest LTS release r27
- [x] Due to android,ios toolchain update to support c++20 full features, `std::views::split` can be use to improve performance
- [x] Remove  3rdparty ghc filesystem, since both ios13 and android5.1 support `std::filesystem`
- [x] Build and run wasm64 support, current need enable chrome flag: `chrome://flags/#enable-webassembly-memory64`
- [x] `Color4B` => `Color32`
- [x] Remove `Color3B`
- [x] Remove deprecated StringUtils::format
- [x] drop GLES 2.0 support since all Android 5.1+ devices support `opengl-es-3.0`
- [x] Improve axslcc preprocessors, remove `METAL`, `GLES2`, then add `AXSLC_TARGET_[GLSL,HLSL,MSL,GLES]`
- [x] Improve VertexLayout management
- [x] Improve `sampler2D texs[4]` support, 2.x only support OpenGL, now support all RHI: d3d11, metal, opengl(es)
- [ ] Reimplement core/physics with box2d v3, not all API compatible with axmol-2.x (in-progress)
     - [x] basic functionals 
     - [x] Rename `PhysicsJointXXX::construct` => `PhysicsJointXXX::instantiate`
     - [ ] b2ContactListener
- [ ] re-select json lib: yyjson > boost.json > rapidjson
    simdjson for parse only and simdjson build in-progress: https://github.com/simdjson/experimental_json_builder
    finally solution: `yyjson`  + `simdjson`
- [ ] drop tolua, use sol2
- [ ] Choose shader for more render objects like `Sprite`
    - [x] ParticleSystemQuad

### For each PR
       
   Optional:
          
### For core/new feature PR


paulocoutinhox:
Version 3 don't will be published with **PhysicsBody** replacement?

image

halx99:
version 3 still in early status, and should contains physics2d when publish.

paulocoutinhox:
Will be nice if we have a single json library easy to use, like `nlohmann/json` instead of two.

halx99:
> Will be nice if we have a single json library easy to use, like `nlohmann/json` instead of two.

The `nlohmann/json` easy to use, but performance not good.

danialias:
I need rapidjson performance as well, please do not remove 😅

paulocoutinhox:
I don't understand how a JSON lib in a game engine needs such an absurd performance, since things like that are used in few moments, like loading or saving things at specific times.

In our app (Ubook) that uses JSON all the time and processes thousands of JSON data with thousands of records, the processing is instantaneous, even in huge JSON data and we use nloman/json. Before we used rapid, but it doesn't have as many features, requiring other libs to be used and more things to be maintained, decreasing efficiency in exchange for NOTHING.

For me, it would make sense to keep a single lib that is easy to use and that for our scenario here at Axmol, will not affect anything in practice.

Here is my opinion based on my reality as someone who uses this in practice and in large volumes, both on Android and iOS, as well as on Apple Watch and Apple TV and also on the desktop.

rh101:
> I need rapidjson performance as well, please do not remove 😅

If it is removed from Axmol, then you can easily add it into your own projects with a few lines of CMake code, and nothing else has to change in your projects.

rh101:
> I don't understand how a JSON lib in a game engine needs such an absurd performance, since things like that are used in few moments, like loading or saving things at specific times.

Your projects may not require the speed of libraries like rapidjson, but that may not apply to projects by others, since they may have completely different requirements.

> In our app (Ubook) that uses JSON all the time and processes thousands of JSON data with thousands of records, the processing is instantaneous, even in huge JSON data and we use nloman/json. Before we used rapid, but it doesn't have as many features, requiring other libs to be used and more things to be maintained, decreasing efficiency in exchange for NOTHING.
 
There are cases where the speed of parsing needs to be as quick as possible, such as in event handlers, comms response handlers, etc. etc..  In these cases, speed definitely does matter, and it is quite noticeable if the parsing takes too long.

> For me, it would make sense to keep a single lib that is easy to use and that for our scenario here at Axmol, will not affect anything in practice.

That may be the case for your own projects, but again, it may not apply to projects by others.  Also, just because a specific library works for you, it doesn't mean that it is appropriate for it to be used in Axmol.

> Here is my opinion based on my reality as someone who uses this in practice and in large volumes, both on Android and iOS, as well as on Apple Watch and Apple TV and also on the desktop.

That's all well and good, and as you stated, your opinion is based on the requirements of your own projects, but it's not really appropriate to make statements about what other developer should or shouldn't use in their projects, because you really don't know what their requirements are.  

The engine should bundle whatever json library (or libraries) it requires, but that in no way forces the developers using Axmol to also use those libraries.  Developers are free to use whatever library they want to use.

On a side note, many developers who use Axmol for their released projects have never posted on this discussion board, and may not even read these threads, so unfortunately, it would be hard to get their feedback on such things.

halx99:
Finally, maybe `yyjson`, speed fast and support read/write and dom style APIs, axmol can make a thin  c++ wrapper

iAndyHD3:
The problem with nlohman/json is not only its lack of runtime speed but it's also a huge >10k thousand long header only library, making compilation also slower. its really one of the worst options you can choose. In modern C++ there is a clear new winner in my opinion:  [glaze](https://github.com/stephenberry/glaze) is a modern C++ json library with automatic serialization/deserialization, making it really convenient for simple use cases. The performance is on par with simdjson. For non reflection uses, there is [generic json](https://github.com/stephenberry/glaze/blob/main/docs/generic-json.md) which is similar to nlohman in that regard.

There are a few cons, thought:
it only supports c++23
Needs recent compiler versions
Not sure if it has been tested in platforms like wasm, ios

paulocoutinhox:
Hi,

Glaze is nice too. Their syntax is like nlohman/json and is easy to use.

My concern is always to keep the engine as easy as possible for people to use. It's better to use Glaze, which already does everything, than to have to adopt several libraries and increase complexity, lowering the barrier to entry.

halx99:
I tried glaze, but compile fails
EDIT: seems it fix msvc compile issues recently

WUCJ638:
I hope DragonBones extensions is retained, for it eases my creating skeleton animation, and cares for those without access to Spine PRO like me, though it's no more under maintenance.

Yet it's said that currently a NEW DragonBone (aka "Loongbones") has been published and under developement, so I believe it's still reasonable to keep it around.🤔

rh101:
> I hope DragonBones extensions is retained, for it eases my creating skeleton animation, and cares for those without access to Spine PRO like me, though it's no more under maintenance.
> 
> Yet it's said that currently a NEW DragonBone (aka "Loongbones") has been published and under developement, so I believe it's still reasonable to keep it around.🤔

You can easily add it directly to your game project, since the engine doesn't actually interact with DragonBones, so it does not need to exist in the engine source tree.

WUCJ638:
> You can easily add it directly to your game project, since the engine doesn't actually interact with DragonBones, so it does not need to exist in the engine source tree.

That is I manually add it to my project or the Engine in my computer...🤔acceptable. Thank you😊.

halx99:
actually, we can reserve all exists axmol  extensions

WUCJ638:
> actually, we can reserve all exists axmol extensions

Great, enabling me to code smoothly. Thanks!😉

aismann:
@halx99 
Just a suggestion:
About: _Renaming Color4B to Color32_

Isn't `Color256` more self-explanatory and therefore better?

halx99:
> @halx99 Just a suggestion: About: _Renaming Color4B to Color32_
> 
> Isn't `Color256` more self-explanatory and therefore better?

Why 256， it's 32bit

rh101:
> Isn't `Color256` more self-explanatory and therefore better?

Color256 does not give you any information at all on the contents of that structure. It really makes no sense at all, given that the actual size is 2^32, not 2^8.  

The only naming that works is either ColorXX, where XX is the number of bits, or as it is right now, ColorXY, where X is the number of values, and Y is the type, such as Color4B/Color4F etc. etc..

aismann:
@halx99 @rh101 
Thanks for explanations. 

I thought Color256 because of the used values 0-255.
Color32 is ok for me too.

aismann:
[Box2D Releases](https://github.com/erincatto/box2d/releases) [v3.1.0](https://github.com/erincatto/box2d/releases/tag/v3.1.0)

aismann:
@halx99 
[Box2D Release v3.1.1 · erincatto/box2d](https://github.com/erincatto/box2d/releases/tag/v3.1.1)
This is a bug fix release. There are a few minor API changes. Making way for 3.2.0 work.
There is no documentation updates with this releases.

What's Changed

- Warning as error off by default by @erincatto in https://github.com/erincatto/box2d/pull/914
- functions to get and set surface material by @erincatto in https://github.com/erincatto/box2d/pull/921
- Fixed another shape distance assert by @erincatto in https://github.com/erincatto/box2d/pull/926
- Revolute and prismatic targets by @erincatto in https://github.com/erincatto/box2d/pull/929
- Cast initial overlap by @erincatto in https://github.com/erincatto/box2d/pull/933
- Tunable joint stiffness by @erincatto in https://github.com/erincatto/box2d/pull/934
- _aligned_free for mingw by @erincatto in https://github.com/erincatto/box2d/pull/939
- Removed MSVC static library setting by @erincatto in https://github.com/erincatto/box2d/pull/940

Full Changelog: https://github.com/erincatto/box2d/compare/v3.1.0...v3.1.1

aismann:
@halx99 
`ax::LayerGradient `is not working, see also
```
10::LayerGradientTest
11::LayerGradientTest2
```

halx99:
> @halx99 `ax::LayerGradient `is not working, see also
> 
> ```
> 10::LayerGradientTest
> 11::LayerGradientTest2
> ```

Thanks, will take a look later

aismann:
Here a yasio bug (
image
I got it running the "Start AutoTest" on cpp-tests

halx99:
> Here a yasio bug ( image I got it running the "Start AutoTest" on cpp-tests

fixed

halx99:
> > @halx99 `ax::LayerGradient `is not working, see also
> > ```
> > 10::LayerGradientTest
> > 11::LayerGradientTest2
> > ```
> 
> Thanks, will take a look later

fixed

core/renderer/backend/ProgramManager.cpp
```diff
@@ -55,6 +55,22 @@ void ProgramManager::destroyInstance()
...
+    switch (pixelFormat)
+    {
+    case PixelFormat::R8:
+        return backend::ProgramType::POSITION_TEXTURE_GRAY;
+    case PixelFormat::RG8:
+        return backend::ProgramType::POSITION_TEXTURE_GRAY_ALPHA;
+    case PixelFormat::RGBA8:
+        return backend::ProgramType::POSITION_TEXTURE_COLOR;
+    default:
+        AXLOGW("Warning: Sprite::chooseSpriteShader() unhandled pixel format {}", (int)pixelFormat);
```
j-jorge: Bad log, it's not `Sprite::chooseSpriteShader()` anymore :)

j-jorge:
What do you think of taking advantage of the break introduced by v3 to remove the axmol.h file? This is the kind of header that explodes the build duration :)

halx99:
> What do you think of taking advantage of the break introduced by v3 to remove the axmol.h file? This is the kind of header that explodes the build duration :)

Need consider long term.

aismann:
> > What do you think of taking advantage of the break introduced by v3 to remove the axmol.h file? This is the kind of header that explodes the build duration :)
> 
> Need consider long term.

Maybe a text like the "copyright hint" (pull_request_template.md) is a "start" on each PR?

[ ]  Check the 'axmol.h' and replace it with the needed headers.

halx99:
> > > What do you think of taking advantage of the break introduced by v3 to remove the axmol.h file? This is the kind of header that explodes the build duration :)
> > 
> > 
> > Need consider long term.
> 
> Maybe a text like the "copyright hint" (pull_request_template.md) is a "start" on each PR?
> 
> [ ] Check the 'axmol.h' and replace it with the needed headers.

I agree

j-jorge:
I'm not sure I understand your feedback correctly, but if we keep the axmol.h file maybe we can put a deprecation warning message in it, and at least remove the `#include <axmol.h>` from this repo? Then we'll be able to remove it in v3.x or v4.

aismann:
image

image

aismann:
> > > > What do you think of taking advantage of the break introduced by v3 to remove the axmol.h file? This is the kind of header that explodes the build duration :)
> > > 
> > > 
> > > Need consider long term.
> > 
> > 
> > Maybe a text like the "copyright hint" (pull_request_template.md) is a "start" on each PR?
> > [ ] Check the 'axmol.h' and replace it with the needed headers.
> 
> I agree

I will change the `pull_request_template.md`
@halx99 
For the 3.0 branch, right?

aismann:
> I'm not sure I understand your feedback correctly, but if we keep the axmol.h file maybe we can put a deprecation warning > message in it, and at least remove the `#include <axmol.h>` from this repo? Then we'll be able to remove it in v3.x or v4.

@j-jorge 
We will start with v3.x.

aismann:
@halx99 
There is a missing assign on `void Label::setTextColor(const Color32& color)`
Correct is:
`    _textColor = _textColor32 = color;`

Maybe all `_textColor` have to rename to `_textColor32`?

halx99:
> @halx99 There is a missing assign on `void Label::setTextColor(const Color32& color)` Correct is: ` _textColor = _textColor32 = color;`
> 
> Maybe all `_textColor` have to rename to `_textColor32`?

fixed by https://github.com/axmolengine/axmol/pull/2387/commits/284c22989e1ca635eb877883e90bf07c5d97816a

aismann:
Assertion failed!

Program: D:\_git\axmol\build\bin\cpp-tests\Debug\cpp-tests.exe
D:\_git\axmol\core\rhi\opengl\ProgramGL.cpp
Line: 143

Expression: vertShader != 0 && fragShader != 0

image

halx99:
> Assertion failed!
> 
> Program: D:_git\axmol\build\bin\cpp-tests\Debug\cpp-tests.exe File: D:_git\axmol\core\rhi\opengl\ProgramGL.cpp Line: 143
> 
> Expression: vertShader != 0 && fragShader != 0
> 
> image

I run on my local computer, works fine, you should `pull --rebase`, and do cmake regenerate

aismann:
> I run on my local computer, works fine, you should `pull --rebase`, and do cmake regenerate


It was a "fresh" cloned repository and I tried also the artifact from the build (RelWithDebInfo) which is like a "standalone installation"
Artifact download URL: https://github.com/axmolengine/axmol/actions/runs/16961100151/artifacts/3763459927

aismann:
Found the "error":
axsc => correct axslc
image

halx99:
> Found the "error": axsc => correct axslc image

should fixed by latest push, please check, thanks

aismann:
> 
> should fixed by latest push, please check, thanks

Works. Thanks. And I like the 3.0!

j-jorge:
I was wondering if v3 wouldn't be a good time for more breaking changes, like getting rid of the `.h` extension for the headers in favor of `.hpp` or something like this, such that we can finally know that it is C++ without opening the file. What do you think?

In the same vein, what do you think of enforcing the code formatting in the CI by running a clang-format check on each PR?

halx99:
This PR is for ci check only, since we try to update branch mangement: enable ci for dev/*, release/*, this PR can closed, v3 roadmap, please see: https://github.com/axmolengine/axmol/discussions/2650

halx99:
> I was wondering if v3 wouldn't be a good time for more breaking changes, like getting rid of the `.h` extension for the headers in favor of `.hpp` or something like this, such that we can finally know that it is C++ without opening the file. What do you think?
> 
> In the same vein, what do you think of enforcing the code formatting in the CI by running a clang-format check on each PR?

auto run clang-foramt check is a good idea.

halx99:
> > I was wondering if v3 wouldn't be a good time for more breaking changes, like getting rid of the `.h` extension for the headers in favor of `.hpp` or something like this, such that we can finally know that it is C++ without opening the file. What do you think?
> > In the same vein, what do you think of enforcing the code formatting in the CI by running a clang-format check on each PR?
> 
> auto run clang-foramt check is a good idea.

should finish by PR: #2654

j-jorge:
If I understood #2654 correctly it creates a new commit which formats the code on each PR. This is imho a bad solutions as those commits will pollute the history with meaningless changes, complexifying  `git blame` and `git bisect`, for example. Still imho, the pipeline should just check that the code is formatted, and let the committer rewrite its commits with adequately formatted code.

halx99:
> If I understood #2654 correctly it creates a new commit which formats the code on each PR. This is imho a bad solutions as those commits will pollute the history with meaningless changes, complexifying `git blame` and `git bisect`, for example. Still imho, the pipeline should just check that the code is formatted, and let the committer rewrite its commits with adequately formatted code.

at least, first time, we can format all sources

PR #2398, opened 2025-02-16
Calculate correct positions for underline and strikethrough 

### LINKED ISSUE #2397: Label underline and strikethrough positioning incorrect when using Overflow::SHRINK
- axmol version: 
- devices test on:
- developing environments
   - NDK version: r23c
   - Xcode version: 14.2+
   - Visual Studio: 
      - VS version: 2022 (17.9+)
      - MSVC version: 19.39+
   - cmake version: 
Steps to Reproduce:

```
auto underlineLabel = Label::createWithTTF("UNDERLINE\nfghijkl0123456789", "fonts/arial.ttf", 32, Size(200, 64));
underlineLabel->setTextColor(Color4B::WHITE);
underlineLabel->enableUnderline();
underlineLabel->setOverflow(Label::Overflow::SHRINK);
underlineLabel->setPositionNormalized(Vec2(0.5f,0.75f));
addChild(underlineLabel);

auto strikethroughLabel = Label::createWithTTF("STRIKETHROUGH\nfghijkl0123456789", "fonts/arial.ttf", 32, Size(200, 64));
strikethroughLabel->setTextColor(Color4B::YELLOW);
strikethroughLabel->enableStrikethrough();
strikethroughLabel->setOverflow(Label::Overflow::SHRINK);
strikethroughLabel->setPositionNormalized(Vec2(0.5f, 0.45f));
addChild(strikethroughLabel);
```

Actual output (TTF):

Image

Actual output (System Font):

Image

Expected output:

Image

rh101:
## Describe your changes

Changes to calculate the correct positions for undereline and strike-through lines when using `setOverflow(Label::Overflow::SHRINK);`.

For TTF, the changes in this PR will ensure the output is accurate.  

For system fonts, we do not have access to information that would give the best output, so it's a rough estimate of where the lines should be.

TTF:
image

System Font:
image

#2397 

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/2d/Label.cpp
+++ b/core/2d/Label.cpp
@@ -1753,28 +1753,32 @@ void Label::updateContent()
 
         _lineDrawNode->clear();
 
-        if (_numberOfLines && _currentLabelType != LabelType::STRING_TEXTURE)
+        if (_numberOfLines > 0 && _currentLabelType != LabelType::STRING_TEXTURE)
         {
             // This is the logic for TTF fonts
-            const float charheight = (_textDesiredHeight / _numberOfLines);
-            float thickness        = charheight / 6;
+            float thickness         = (_lineHeight * _fontScale) / 6;
+            float nextTokenY        = _letterOffsetY;
+            auto contentScaleFactor = AX_CONTENT_SCALE_FACTOR();
+            float lineSpacing       = _lineSpacing * contentScaleFactor;
 
             // atlas font
             for (int i = 0; i < _numberOfLines; ++i)
             {
                 if (_strikethroughEnabled)
                 {
-                    float y = (_numberOfLines - i - 1) * charheight + charheight / 2;
+                    auto y = nextTokenY / contentScaleFactor - _lineHeight * _fontScale / 2;
                     _lineDrawNode->drawLine(Vec2(_linesOffsetX[i], y), Vec2(_linesWidth[i] + _linesOffsetX[i], y),
                                             Color4F(lineColor), thickness);
                 }
 
                 if (_underlineEnabled)
                 {
-                    float y = (_numberOfLines - i - 1) * charheight;
+                    auto y = (nextTokenY * _fontScale) / contentScaleFactor - _lineHeight * _fontScale;
                     _lineDrawNode->drawLine(Vec2(_linesOffsetX[i], y), Vec2(_linesWidth[i] + _linesOffsetX[i], y),
                                             Color4F(lineColor), thickness);
                 }
+
+                nextTokenY -= _lineHeight * _fontScale + lineSpacing;
             }
         }
         else if (_textSprite) // ...and is the logic for System fonts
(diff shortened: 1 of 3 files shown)


rh101:
Adjusted the line thickness to 12% of line height, and made it a multiplication for clarity. Previously the divide by 6 would have resulted in a 16.7% thickness.

The 12% output is as follows:
image

16% gives the following, which is a little too thick:
image

Is the 12% thickness acceptable?

EDIT: @aismann The post is fixed, just please be more careful when replying that you don't accidentally edit someone else's post.

rh101:
@aismann Be careful when you try to reply to posts, because you ended up editing my post:
image

I'll manually fix it.

aismann:
@rh101 
Want you add some tester too?
Im not sure it is important having tester about this too..

rh101:
> @rh101 Want you add some tester too? Im not sure it is important having tester about this too..

I'll check if the current cases cover it, and if they don't, I'll add one in there.

rh101:
Test 59 will cover the `Overflow::SHRINK` mode with the underline and strike-through

image

halx99 (approved):

aismann:
@rh101 
Looks it works not correct with system fonts 
image

rh101:
> @rh101 Looks it works not correct with system fonts 

That specific test _never_ worked with the code prior to this PR as well, and it was never noticed because there was no option to turn underline and strike-through on in that test.  Also, as I mentioned at the start of this thread, "For system fonts, we do not have access to information that would give the best output, so it's a rough estimate of where the lines should be.".

In general, it does work correctly when the label dimension is the same as the label text size, meaning if the label boundary matches the boundary of the text.  In the screenshot for this test, you can see that the label dimensions are a lot larger than the true content size:
image

This is the screenshot with the Label changes from this PR reverted, with label test code changes still in to show the underline and strike-through:
image

Now, back to the latest code from this PR, with label content size the same as the label dimensions (the left slider in that test):

Underline:
image

Strikethrough:
image

The left slider in that test changes the dimension size via `Label::setDimensions()`:
```
    slider->addEventListener([this, slider, winSize](Object* /*sender*/, Slider::EventType event) {
        float percent     = slider->getPercent();
        auto labelSize    = _label->getContentSize();
        auto drawNodeSize = Size(percent / 100.0 * winSize.width, labelSize.height);
        if (drawNodeSize.width <= 0)
        {
            drawNodeSize.width = 0.1f;
        }
        _label->setDimensions(drawNodeSize.width, drawNodeSize.height);
        this->updateDrawNodeSize(drawNodeSize);
    });
```
So, in summary, the problem seen in test 63 has always existed, and it occurs when `Label::setDimensions()` is called with a size that is larger than the max text height.

~~Actually, that gives me an idea; I'll see if it is possible to use the `_labelDimensions` in the calculation to try to get a better result.~~ EDIT: this isn't possible either, since the `_contentSize` is eventually set to the same as the `_labelDimensions`, so I can't find any way to get the true text boundary compared to the texture boundary (meaning not including the empty space above and below the text etc..).

aismann:
- A (new) setter method for the pos of the line is also an idea
```
<label>::setLinePos(Vec2 pos, lineType);
enum lineType 
{
 underline,
 strikethrough,
};
```

rh101:
> * A (new) setter method for the pos of the line is also an idea
> 
> 
> ```
> setLinePos(Vec2 pos, lineType = )
> enum lineType 
> {
>  underline,
>  strikethrough,
> };
> ```

Sorry, I don't quite follow. What do you mean?

aismann:
> > * A (new) setter method for the pos of the line is also an idea
> > 
> > ```
> > setLinePos(Vec2 pos, lineType = )
> > enum lineType 
> > {
> >  underline,
> >  strikethrough,
> > };
> > ```
> 
> Sorry, I don't quite follow. What do you mean?

The developer can set itself the line height:
`myLabel->setLinePos `or `myLabel->setLineHeightOffset `

rh101:
> The developer can set itself the line height

The line height isn't the issue.  The actual problem is the offset between the top of the label texture (for system fonts only) to where the text actually starts in that texture.  There is no way for the developer to know what it is, since there is no way to actually figure it out, at least from what I've seen.

aismann:
> > The developer can set itself the line height
> 
> The line height isn't the issue. The actual problem is the offset between the top of the label texture (for system fonts only) to where the text actually starts in that texture. There is no way for the developer to know what it is, since there is no way to actually figure it out, at least from what I've seen.

Ok I understand.
Should be part of the wiki thats system font not used for  underline, strikethrough.

rh101:
> Should be part of the wiki thats system font not used for underline, strikethrough.

System font underline and strikethrough does work fine, as long as the no dimensions are set for the label, since it auto-sizes based on the text content.

Perhaps a wiki post mentioning that if custom dimensions are set, underline and strikethrough may not display correctly when using system fonts.

PR #2403, opened 2025-02-17
Fix tilemap culling

DelinWorks:
## Describe your changes
I've reverted the changes to the old ones for now but this seems to break `cpp_tests`. as for normal projects like the sample one provided by @rudiHammad seems to work, I noticed that in cpp_tests even with the cocos2dx implementation, the tiles seem to be culled twice the top and right. there's a discrepancy that I couldn't find somewhere, I'll try to find the old sample I was working with in #1134  

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/2d/FastTMXLayer.cpp
+++ b/core/2d/FastTMXLayer.cpp
@@ -135,16 +135,14 @@ void FastTMXLayer::draw(Renderer* renderer, const Mat4& transform, uint32_t flag
     updateTotalQuads();
 
     auto cam = Camera::getVisitingCamera();
-    if (flags != 0 || _dirty || _quadsDirty || !_cameraPositionDirty.fuzzyEquals(cam->getPosition(), _tileSet->_tileSize.x) ||
-        _cameraZoomDirty != cam->getZoom())
+    auto proj = Director::getInstance()->getProjection();
+    if (flags != 0 || _dirty || _quadsDirty)
     {
-        _cameraPositionDirty = cam->getPosition();
-        auto zoom            = _cameraZoomDirty = cam->getZoom();
-        Vec2 s               = _director->getVisibleSize();
-        const Vec2& anchor   = getAnchorPoint();
-        auto rect            = Rect(cam->getPositionX() - getPositionX() - s.width * zoom * 0.5,
-                                    cam->getPositionY() - getPositionY() - s.height * zoom * 0.5,
-                                    s.width * zoom, s.height * zoom);
+        Vec2 s             = _director->getVisibleSize();
+        const Vec2& anchor = getAnchorPoint();
+        auto rect = Rect(Camera::getVisitingCamera()->getPositionX() - s.width * (anchor.x == 0.0f ? 0.5f : anchor.x),
+                         Camera::getVisitingCamera()->getPositionY() - s.height * (anchor.y == 0.0f ? 0.5f : anchor.y),
+                         s.width, s.height);
 
         Mat4 inv = transform;
         inv.inverse();
--- a/core/2d/FastTMXLayer.cpp
+++ b/core/2d/FastTMXLayer.cpp
@@ -135,7 +135,6 @@ void FastTMXLayer::draw(Renderer* renderer, const Mat4& transform, uint32_t flag
     updateTotalQuads();
 
     auto cam = Camera::getVisitingCamera();
-    auto proj = Director::getInstance()->getProjection();
     if (flags != 0 || _dirty || _quadsDirty)
     {
         Vec2 s             = _director->getVisibleSize();
--- a/core/2d/FastTMXLayer.cpp
+++ b/core/2d/FastTMXLayer.cpp
@@ -135,13 +135,17 @@ void FastTMXLayer::draw(Renderer* renderer, const Mat4& transform, uint32_t flag
     updateTotalQuads();
 
     auto cam = Camera::getVisitingCamera();
-    if (flags != 0 || _dirty || _quadsDirty)
+    if (flags != 0 || _dirty || _quadsDirty ||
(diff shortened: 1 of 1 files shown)


rh101:
You didn't need to revert the entire code, since it seemed to be working with the zoom values.  The only part that needed to be removed was the section:

```
rect.origin.x -= _tileSet->_tileSize.x;
rect.origin.y -= _tileSet->_tileSize.y;
rect.size.x += s.x * zoom / 2 + _tileSet->_tileSize.x * zoom;
rect.size.y += s.y * zoom / 2 + _tileSet->_tileSize.y * zoom;
```

Do you still see issues with cpp-tests if you leave it as this (with the 4 lines removed):
```
if (flags != 0 || _dirty || _quadsDirty || !_cameraPositionDirty.fuzzyEquals(cam->getPosition(), _tileSet->_tileSize.x) ||
    _cameraZoomDirty != cam->getZoom())
{
    _cameraPositionDirty = cam->getPosition();
    auto zoom            = _cameraZoomDirty = cam->getZoom();
    Vec2 s               = _director->getVisibleSize();
    const Vec2& anchor   = getAnchorPoint();
    auto rect = Rect(cam->getPositionX() - s.width * zoom * (anchor.x == 0.0f ? 0.5f : anchor.x),
                     cam->getPositionY() - s.height * zoom * (anchor.y == 0.0f ? 0.5f : anchor.y),
                     s.width * zoom, s.height * zoom);

    Mat4 inv = transform;
    inv.inverse();
    rect = RectApplyTransform(rect, inv);

    updateTiles(rect);
    updateIndexBuffer();
    updatePrimitives();
    _dirty = false;
}
```

DelinWorks:
did you test it for the 2 "samples" (rudiHammad one and cpp_tests)? because the solution you put seems to solve only one as well as the cocos2dx one
let me see if there's a discrepancy 

EDIT: yes there's a discrepancy in the 2 different samples

rh101:
> did you test it for the 2 "samples" (rudiHammad one and cpp_tests)? because the solution you put seems to solve only one as well as the cocos2dx one let me see if there's a discrepancy

Specifically which tests in cpp-tests are you referring to?

DelinWorks:
sample 31 in FastTileMap

rh101:
> sample 31 in FastTileMap

OK, and what is exactly do you see with that test? I'm running it now on Windows, and I don't see any strange issues with it when removing the 4 lines from my previous post.  I don't know what I'm looking for.

DelinWorks:
hmm,, let me see

DelinWorks:
here @rh101, the first test where it's all red tiles the culling works perfectly whereas the cpp_tests one seems broken

rh101:
I'm not sure I understand what is meant by "the tiles seem to be culled twice the top and right.". What I do see is that if the map moves off the screen to the top or the right, it is still being drawn, even with the 4 lines removed.

Left side:
image

Right side:
image


606 vs 1398 GL verts.  You would have to move it much further out to the right when it is well off the screen to see any change.

DelinWorks:
sorry I meant twice "x2" the width and height, which is the incorrect behaviour the first issue described by rudiHammad, I might've not explained it properly

DelinWorks:
> 606 vs 1398 GL verts. You would have to move it much further out to the right when it is well off the screen to see any change.

Yes exactly, now.. no matter what you do there's always one that will fail resulting in the case you said where it would have to move much further to the right or top to be culled

rh101:
> here @rh101, the first test where it's all red tiles the culling works perfectly whereas the cpp_tests one seems broken

OK, so we're referring to the same thing. I understand now; initially thought you were referring to some flickering issue.

Anyhow, as long as there are no graphical glitches, removing those 4 lines for now should be enough.  Working out the issue with culling off the screen to the top and right isn't something urgent.

DelinWorks:
> Working out the issue with culling off the screen to the top and right isn't something urgent.

that's what the original issue wanted to fix but I guess as long as it doesn't cause visual culling to the user then it's fine, better to have more tiles than less tiles that disappear in front of the user, will in case of performance that's the opposite 😅

rh101:
@DelinWorks I think I just found the problem, by accident. Can you please verify if this is the cause:

```
void FastTMXLayer::updateTiles(const Rect& culledRect)
{
    Rect visibleTiles        = Rect(culledRect.origin, culledRect.size * _director->getContentScaleFactor()); <<<< this multiplication of the size
...
}
```

Should be:
```
void FastTMXLayer::updateTiles(const Rect& culledRect)
{
    Rect visibleTiles        = Rect(culledRect.origin, culledRect.size);
...
```

DelinWorks:
oh good catch!! I'll see if it fixes it

DelinWorks:
yep it fixed it for both! very nice job @rh101

aismann:
> yep it fixed it for both! very nice job @rh101

@rh101 is one of the best Contributers here ;)

DelinWorks:
> > yep it fixed it for both! very nice job @rh101
> 
> @rh101 is one of the best Contributers here ;)

strong agree

rudiHammad:
Awesome, great work.

rh101:
I'm really glad this issue is sorted now.  I spent some time on it last night trying to figure out what was wrong with the `FastTMXLayer::draw` method, as that was the only thing that had changed, and leaving the `FastTMXLayer::updateTiles` section alone, since it hadn't changed from Cocos2d-x... and now it turns out this issue has been in there from Cocos2d-x this whole time.  I should have paid more attention to it.

DelinWorks:
yeah me too.. I thought everything was perfect and that the error must've been the culling rect calculations I made lol
this bug got me pulling out 2 devices 💀💀

halx99 (approved):

PR #2410, opened 2025-02-20
[Android] Add a way to disable interactions on the SurfaceView

AlexandreK38:
## Describe your changes
I needed a way to disable the interactions on Android, but nor `surfaceView.setEnabled(false)` nor `surfaceView.setClickable(false);` is working as it was in the past.
I added a Getter and Setter to be able to disable interactions on the events on the surface view directly:
- Key up & down
- Touch Down only so any gestures already started could end

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/platform/android/java/src/dev/axmol/lib/AxmolGLSurfaceView.java
+++ b/core/platform/android/java/src/dev/axmol/lib/AxmolGLSurfaceView.java
@@ -64,7 +64,7 @@ public class AxmolGLSurfaceView extends GLSurfaceView {
 
     private boolean mSoftKeyboardShown = false;
     private boolean mMultipleTouchEnabled = true;
-    private boolean mInteractiveEnabled = true;
+    private boolean mInteractive = true;
 
     private CountDownLatch mNativePauseComplete;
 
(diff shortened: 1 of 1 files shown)


core/platform/android/java/src/dev/axmol/lib/AxmolGLSurfaceView.java
```diff
@@ -83,6 +84,14 @@ public void setMultipleTouchEnabled(boolean multipleTouchEnabled) {
         this.mMultipleTouchEnabled = multipleTouchEnabled;
     }
 
+    public boolean areInteractionsEnabled() {
```
halx99:
```suggestion
    public boolean isInteractionsEnabled() {
```

AlexandreK38: done 👍

rh101:
Sorry to chime in so late. The word "interactivity" may be more suitable for this, since it describes exactly what the purpose of these methods are, to enable or disable interactivity.  So, how about `setInteractivityEnabled()`, `isInteractivityEnabled`, and `mInteractivityEnabled `? 

The current "interactions" implies setting multiple instances of an interaction, which isn't actually the case.

halx99:
While `setInteractiveEnabled` `isInteractiveEnabled` may be better

AlexandreK38:
English is not my first language, so whatever you prefer and seems more suitable to describe what we do

rh101:
> While `setInteractiveEnabled` `isInteractiveEnabled` may be better

Yes, that would also work. The "enabled" word isn't actually required in that, since "interactive" is self-descriptive, so "setInteractive()" and "isInteractive()" have the correct meaning.

For instance, `setInteractive(true)` and `setInteractive(false)` make sense for anyone reading the code.

halx99:
Yes, `setInteractive` should be better and simpler, search on github global: 
- `setInteractionEnabled`: 18K results
- `setInteractive`: 84K results

AlexandreK38:
We should be good with `setInteractive`

halx99 (approved):

PR #2426, opened 2025-03-03
Simplify project CMakeLists.txt by moving code sections to cmake modules

### LINKED ISSUE #2373: Improve updating old axmol projects to new axmol versions

When updating an old axmol project the root CMakeLists.txt must be updated too, because it might have gotten changes since then. Not updating it with the most recent template CMakeLists.txt might result in compilation errors. This is not an annoying process but it must be noticed by the user first and it might not be entirely obvious, and the problem with updating it manually is that you can't just copy paste the template cmakelists.txt, you need to change all "Dummy" strings to your projects name. 

**Describe the solution you'd like**
Add a setup_axmol_application function in axmol cmake itself so the template can be as short as 

```cmake 
set(APP_NAME Dummy)
project(${APP_NAME}) 
setup_axmol_project(${APP_NAME})
```
Advantages:
updating axmol project does not require user specific action
clean cmake on the user side
easier to spot project specific cmake code

**Describe alternatives you've considered**
Adding an update command to the axmol command line that fetches and updates cmakelists.txt replacing project name

**To consider**
The setup cmake function would need to handle lua project aswell

rh101:
## Describe your changes

@iAndyHD3 put forward the idea of simplifying the project `CMakeLists.txt` in order to reduce the effort required for the process of updating to new versions of Axmol.  This is a good idea, one that also helps new developers who are starting to use Axmol, since it makes the `CMakeLists.txt` less overwhelming, and much easier to modify for their projects.

This PR has an example of how the project `CMakeLists.txt` can be simplified.  A lot of the code in the existing `CMakeLists.txt` would never be modified by the project developer, so those sections have been moved out to cmake module files.  If there are unique requirements to modify the code in the modules, then the developer can do so easily, by either modifying the code in the module, or just not including the module, and instead copying the code over to the project `CMakeLists.txt`

#2373 

### For each PR
       
   Optional:
          
### For core/new feature PR


rh101:
This is what a C++ project `CMakeLists.txt` would look like when a new project is created:
```
cmake_minimum_required(VERSION 3.20)

set(APP_NAME MyProject)

project(${APP_NAME})

if(XCODE)
    set(CMAKE_XCODE_GENERATE_TOP_LEVEL_PROJECT_ONLY TRUE)
endif()

set(_is_axmol_embed FALSE)
set(_AX_USE_PREBUILT FALSE)

set(CMAKE_MODULE_PATH
    ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules
)

include(AXGameEngineSetup)

# The common cross-platforms source files and header files
file(GLOB_RECURSE GAME_HEADER
    Source/*.h Source/*.hpp
)

file(GLOB_RECURSE GAME_SOURCE
    Source/*.cpp Source/*.c
)

set(GAME_INC_DIRS
    "${CMAKE_CURRENT_SOURCE_DIR}/Source"
)

set(content_folder
    "${CMAKE_CURRENT_SOURCE_DIR}/Content"
)

if(APPLE)
    ax_mark_multi_resources(common_content_files RES_TO "Resources" FOLDERS ${content_folder})
elseif(WINDOWS)
    ax_mark_multi_resources(common_content_files RES_TO "Content" FOLDERS ${content_folder})
endif()

include(AXGameSourceSetup)

# mark app complie info and libs info
set(APP_SOURCES
    ${GAME_HEADER}
    ${GAME_SOURCE}
    )

include(AXGameTargetSetup)

# mark app resources, resource will be copy auto after mark
ax_setup_app_config(${APP_NAME})

# Add any libraries you need to link to the project after this point

# Default Platform-specific setup
include(AXGamePlatformSetup)

# Make sure the AXGameFinalSetup is included at the end of this file
include(AXGameFinalSetup)
```

templates/cpp/CMakeLists.txt
```diff
@@ -33,209 +33,59 @@ if(XCODE)
...
-if (_AX_USE_PREBUILT) # support windows and linux
-    use_ax_compile_define(${APP_NAME})
+# Default Apple-specific setup
+include(AXGameAppleSetup)
 
-    include(AXLinkHelpers)
-    ax_link_cxx_prebuilt(${APP_NAME} ${_AX_ROOT} ${AX_PREBUILT_DIR})
-endif()
+# Default Windows-specific setup
+include(AXGameWindowsSetup)
```
halx99: could we automatically check platform-spec setup?

rh101: Do you mean move the platform checks out of the modules and have them like this?

```
if(APPLE)
    # Default Apple-specific setup
    include(AXGameAppleSetup)
elseif(WINDOWS)
    # Default Windows-specific setup
    include(AXGameWindowsSetup)
endif()
```

halx99: move into , not out, looks like `include(AXGamePlatformSetup)` in project `CMakeLists.txt`

rh101: Oh, I understand.  The reason I split it was because the Apple section tends to be heavily modified by developers, so I had it in its own file.  I'll merge them together though, since developers can just edit that module directly with their changes.

halx99:
Verified by: https://github.com/axmolengine/axmol/actions/runs/13701800613

halx99:
@rh101  the winuwp platform verify fail, could you take a look, steps:
1. axmol new HelloCpp
2. axmol -p winuwp -d HelloCpp

halx99:
> @rh101 the winuwp platform verify fail, could you take a look, steps:
> 
> 1. axmol new HelloCpp
> 2. axmol -p winuwp -d HelloCpp

I also check on local machine, same link error:

image

halx99:
Without this PR, works fine: https://github.com/simdsoft/axmol/actions/runs/13704010552

rh101:
> > @rh101 the winuwp platform verify fail, could you take a look, steps:
> > 
> > 1. axmol new HelloCpp
> > 2. axmol -p winuwp -d HelloCpp
> 
> I also check on local machine, same link error:
> 
Sorry, I should have been more thorough with the checks.  I'll sort it out today.

rh101:
The issue was to do with `ax_setup_winrt_sources()`, since the call was moved to a module in `{project}/modules/cmake/`, the path from `CMAKE_CURRENT_LIST_DIR` is based on where the function is called from, so the resulting path would have been `{project}/modules/cmake/proj.winrt/App.xaml`, which is not correct.  

Changing it to `CMAKE_CURRENT_SOURCE_DIR` resolves the issue, since that would be the path of the `CMakeLists.txt` that includes the modules, so it would end up with `{project}/proj.winrt/App.xaml`.

halx99:
EDIT: Wait verification action success: https://github.com/axmolengine/axmol/actions/runs/13723857352

halx99:
> EDIT: Wait verification action success: https://github.com/axmolengine/axmol/actions/runs/13723857352

New error:

image

rh101:
> New error:

What command was used that resulted in that error? 

EIDT: I've tried several commands and new projects, but can't seem to reproduce this.  Is there something different about that DemoApp, like a different configuration set inside the `CMakeLists.txt`?

halx99:
> > New error:
> 
> What command was used that resulted in that error?
> 
> EIDT: I've tried several commands and new projects, but can't seem to reproduce this. Is there something different about that DemoApp, like a different configuration set inside the `CMakeLists.txt`?

see: https://github.com/axmolengine/axmol/actions/runs/13723857352/job/38385481612#step:3:790

halx99:
Caused by concurrent execute Content & axslc resources, it not 100% occured, the issue not caused by this PR

PR #2452, opened 2025-03-13
Prevent getCurrentTime() from returning 0 at the end of the audio track playback

### LINKED ISSUE #2451: AudioEngineImpl::getCurrentTime never gets to 100% of sound duration
- axmol version:
- devices test on: Windows 10 x64 build
- developing environments
   - NDK version: r23c
   - Xcode version: 14.2+
   - Visual Studio: 
      - VS version: 2022 (17.9+)
      - MSVC version: 19.39+
   - cmake version: 
Steps to Reproduce:

When playing an audio track, and calling `getCurrentTime()` for that audio ID every loop, it should go from 0.0 to max sound duration (in seconds), but the issue is that, depending on the track length, it will start returning 0 just before it gets to end.   If using the audio playback percentage to synchronize other things, such as animations, then `getCurrentTime()` suddenly returning 0 when it should return the total sound time an issue.

In the `getCurrentTime` method,  `player->_ready` becomes `false` when the audio track finishes, so it returns 0.0, but technically if the audio was playing correctly, and did reach the end, then it should return sound duration.  If the audio never started playing, then a return value of 0.0 would be expected.
```
float AudioEngineImpl::getCurrentTime(AUDIO_ID audioID)
{
    std::unique_lock<std::recursive_mutex> lck(_threadMutex);
    auto it = _audioPlayers.find(audioID);
    if (it == _audioPlayers.end())
        return 0.0f;

    float ret   = 0.0f;
    auto player = it->second;
    if (player->_ready)
    {
        if (player->_streamingSource)
        {
            ret = player->getTime();
        }
        else
        {
            alGetSourcef(player->_alSource, AL_SEC_OFFSET, &ret);

            auto error = alGetError();
            if (error != AL_NO_ERROR)
            {
                AXLOGE("{}, audio id:{},error code:{:#x}", __FUNCTION__, audioID, error);
            }
        }
    }

    return ret;
}
```

Would it be possible to track the state of the `AudioPlayer` a bit differently, so if the following conditions can be checked:

1. Audio did start playing
2. Audio reached the end

If the above 2 conditions are met, then `AudioEngineImpl::getCurrentTime()` should return sound duration.  If the sound never started playing, then it should return 0.0.  If the sound was playing, but there is some fault, then what it returns is undefined.

At the moment the only thing that is tracked is the `_ready` flag, but it's not enough to know if the audio did start playing, or never started playing, so that is not enough information for `getCurrentTime()` to work out if it should return 0.0 or sound duration.

rh101:
## Describe your changes

If the audio track started playing, and it has reached the end, then `getCurrentTime()` should return the total sound duration instead of 0.  The changes in this PR are a possible solution to the problem.  I'm not too familiar with that code to work out if there is a better way to fix this issue.

#2451

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/audio/AudioEngineImpl.cpp
+++ b/core/audio/AudioEngineImpl.cpp
@@ -829,6 +829,10 @@ float AudioEngineImpl::getCurrentTime(AUDIO_ID audioID)
             {
                 AXLOGE("{}, audio id:{},error code:{:#x}", __FUNCTION__, audioID, error);
             }
+            else if (ret == 0.0f && player->isFinished())
+            {
+                ret = player->_audioCache->_duration;
+            }
         }
     }
 
--- a/core/audio/AudioEngineImpl.cpp
+++ b/core/audio/AudioEngineImpl.cpp
@@ -829,10 +829,11 @@ float AudioEngineImpl::getCurrentTime(AUDIO_ID audioID)
             {
                 AXLOGE("{}, audio id:{},error code:{:#x}", __FUNCTION__, audioID, error);
             }
-            else if (ret == 0.0f && player->isFinished())
-            {
-                ret = player->_audioCache->_duration;
-            }
+        }
+
+        if (ret == 0.0f && player->isFinished())
+        {
+            ret = player->_audioCache->_duration;
         }
     }
 


rh101:
Would it be better to have it this way instead:
```
float AudioEngineImpl::getCurrentTime(AUDIO_ID audioID)
{
    std::unique_lock<std::recursive_mutex> lck(_threadMutex);
    auto it = _audioPlayers.find(audioID);
    if (it == _audioPlayers.end())
        return 0.0f;

    float ret   = 0.0f;
    auto player = it->second;
    if (player->_ready)
    {
        if (player->_streamingSource)
        {
            ret = player->getTime();
            if (ret == 0.0f && player->isFinished())
            {
                ret = player->_audioCache->_duration;
            }
        }
        else
        {
            alGetSourcef(player->_alSource, AL_SEC_OFFSET, &ret);

            auto error = alGetError();
            if (error != AL_NO_ERROR)
            {
                AXLOGE("{}, audio id:{},error code:{:#x}", __FUNCTION__, audioID, error);
            } 
            else if (ret == 0.0f && player->isFinished())
            {
                ret = player->_audioCache->_duration;
            }
        }
    }

    return ret;
}
```

That way, if `alGetSourcef` does generate an error, it won't try to call `isFinished()`, which would result in more calls to OpenAL when it is already in an error state:
```
bool AudioPlayer::isFinished() const
{
    if (_streamingSource)
        return _isRotateThreadExited;
    else
    {
        ALint sourceState;
        alGetSourcei(_alSource, AL_SOURCE_STATE, &sourceState);
        return sourceState == AL_STOPPED;
    }
}
```

PR #2465, opened 2025-03-19
Add interactions control support on GLView

AlexandreK38:
## Describe your changes

Hi!
I added a way to disable touch events on the GLView, and when you disable the events I cancel all current touches.
Any feedback is welcome.
I also plan to remove the same behavior I did on the AxmolSurfaceView as this solution is working for any platform. Does it work for you?

### For each PR
- [ ] Add Copyright if it missed:   
      - `"Copyright (c) 2019-present Axmol Engine contributors (see AUTHORS.md)."`
- [ ] I have performed a self-review of my code.
       
   Optional:
   - [ ] I have checked readme and add important infos to this PR.
   - [ ] I have added/adapted some tests too.
          
### For core/new feature PR
- [ ] I have checked readme and add important infos to this PR.
- [ ] I have added thorough tests.

PR #2472, opened 2025-03-22
Revert "[Android] Add a way to disable interactions on the SurfaceView"

halx99:
Reverts axmolengine/axmol#2410


halx99:
see also: #2465

AlexandreK38:
Thanks for the revert 🙏

PR #2483, opened 2025-03-28
[WASM] set _APP_RES_FOLDER if not defined to be able to specify directories of Resources

AlexandreK38:
## Describe your changes

Currently there is no way to specify the directory/ies of the resources for Wasm (it is "Content/")
~~I added `AX_WASM_RES_FOLDER` to be able to specify the directories~~

I changed to check if _APP_RES_FOLDER is defined before setting it so we can set it to whatever directories we need for Wasm resources

### For each PR
       
   Optional:
          
### For core/new feature PR


paulocoutinhox:
I don't understand why we need a specific variable for wasm `AX_WASM_RES_FOLDER` and not use a generic variable `_APP_RES_FOLDER`.

AlexandreK38:
`_APP_RES_FOLDER` was only used and defined here, another solution would be to use `content_folder` so as for APPLE and pass it as param to the macro

AlexandreK38:
Updated to use `_APP_RES_FOLDER` and set it if not defined

Issue #2513, opened 2025-04-28
Label: underline/strikthrough is not correct clipped on 'Clamp Content Test'


aismann:
Works not if wrap is off:
![Image](https://github.com/user-attachments/assets/8e97d351-afa7-4170-b161-4553790c0a5e)

wrap enabled works:
![Image](https://github.com/user-attachments/assets/3d5a2354-1f3a-493a-b783-d860d0ed1611)

Also the last char is not correct clipped sometimes: 
![Image](https://github.com/user-attachments/assets/c9258d5f-9b84-4881-b579-dbff17f745f5)

aismann:
Works correct here  (test issue with 56/57 ?) :

![Image](https://github.com/user-attachments/assets/5f8df3e7-9425-4f89-badd-7c1a470be8c4)

rh101:
For tests 56/57, clamping the line works by not adding the character to the draw list, as you can see in `Label::updateQuads()`.  The issue is that the line length is still the full length without clamping, so when the underline or strikethrough are being drawn, they don't factor in the clamp cut-off.  That's always been the case, all the way back to Cocos2d-x.

For test 61, it works correctly because the label dimensions are being updated, so the line lengths match the actual displayed string length.  It's a completely different operation than clamping, so can't be compared with tests 56/57.

rh101:
A quick fix would be to change this:
```
if (_strikethroughEnabled)
{
    auto y = nextY - lineHeight / 2;
    _lineDrawNode->drawLine(Vec2(_linesOffsetX[i], y), Vec2(_linesWidth[i] + _linesOffsetX[i], y),
                            Color4F(lineColor), thickness);
}

if (_underlineEnabled)
{
    auto y = nextY - lineHeight;
    _lineDrawNode->drawLine(Vec2(_linesOffsetX[i], y), Vec2(_linesWidth[i] + _linesOffsetX[i], y),
                            Color4F(lineColor), thickness);
}
```

to this:
```
float startOffset = 0;
float endOffset  = 0;

if (_overflow == Overflow::CLAMP)
{
    startOffset = std::max(0.f, _linesOffsetX[i]);
    if (_hAlignment == TextHAlignment::LEFT || _hAlignment == TextHAlignment::CENTER)
    {
        endOffset = std::min(_linesWidth[i] + startOffset, _labelDimensions.width - startOffset);
    }
    else // _hAlignment == TextHAlignment::RIGHT
    {
[...]

PR #2515, opened 2025-04-29
Fix character, underline and strikethrough drawing over label boundary for Overflow::CLAMP mode

rh101:
## Describe your changes

Ensure that underline and strike-through lines are clamped correctly when overflow is set to `Overflow::CLAMP`

This also fixes the issue of characters being drawn over the label boundary when using `Overflow::CLAMP`.

Label tests 56 and 57 cover this in the cpp-tests project.

#2513 
#2516

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/2d/Label.cpp
+++ b/core/2d/Label.cpp
@@ -1181,10 +1181,10 @@ bool Label::computeHorizontalKernings(const std::u32string& stringToRender)
         return true;
 }
 
-bool Label::isHorizontalClamped(float letterPositionX, int lineIndex)
+bool Label::isHorizontalClamped(float letterPositionX, float letterWidth, int lineIndex)
 {
     auto wordWidth       = this->_linesWidth[lineIndex];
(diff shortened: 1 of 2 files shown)


aismann:
For backwards compatilibty and "no API change":
` bool isHorizontalClamped(float letterPositionX, int lineIndex,  float letterWidth = 0);`

rh101:
> For backwards compatilibty and "no API change": ` bool isHorizontalClamped(float letterPositionX, int lineIndex, float letterWidth = 0);`

There is no change to the API, since the method is not virtual, and also not part of the public API, so it will not be called outside of Label.

~~If update it regardless of that, but it really is unnecessary.~~
Actually, `letterWidth` cannot be set to 0 as default, since that would not behave as it did previously, so it must remain as-is.  There is no API change, so it's a non-issue.

PR #2518, opened 2025-04-29
Fix for clamped and left-aligned text with certain character sets

### LINKED ISSUE #2516: Label: incorrect character clipping,..has to do with calls to isHorizontalClamped
see also #2513
Also the last char is not correct clipped sometimes:
Image

rh101:
## Describe your changes

With left-alignment, it seems that certain characters can result in negative X position, even though the left-aligned text should be at 0.

This change just partially reverts the calculation to how it was previously, but only for the left position of the text, using half the width to check if it is out bounds, `(letterPositionX + letterWidth / 2) < 0`. 

#2516 

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/core/2d/Label.cpp
+++ b/core/2d/Label.cpp
@@ -1184,7 +1184,7 @@ bool Label::computeHorizontalKernings(const std::u32string& stringToRender)
 bool Label::isHorizontalClamped(float letterPositionX, float letterWidth, int lineIndex)
 {
     auto wordWidth       = this->_linesWidth[lineIndex];
-    bool letterOverClamp = ((letterPositionX + letterWidth) > _contentSize.width || letterPositionX < 0);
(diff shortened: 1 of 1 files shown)


aismann:
@rh101 
Is a "comment line' usefull for fix/know there should be a deepher check why the first character is negative sometimes?

rh101:
> @rh101
> Is a "comment line' usefull for fix/know there should be a deepher check why the first character is negative sometimes?

The reason why it's negative is because for that character, the `FontAtlas::prepareLetterDefinitions` call ends up with this calculation:
image

The eventual result (when factoring in the content scale factor), is the `-1.xx` offset position.  

The actual calculation has nothing to do with the Label class, but rather `FontAtlas.cpp`, so perhaps it's best not to clutter the code with comments that are unrelated to the Label implementation, and don't serve any real purpose.  The negative value seems to be correct, but that's for someone with more experience with the font code to decide.  Either way, it's documented here.

PR #2532, opened 2025-05-09
[FairyGUI] Add support for blending

### LINKED ISSUE #2486: FairyGUI: Unable to Use BlendMode due to The Lack of Its Implementation
- Axmol version:2.4.0
- Devices test on: Windows 11 24H2 (26100.3476)
- Developing environments
   - (Only in) Visual Studio: 
      - VS version: 2022 (17.13.5)
      - MSVC version: 19.39+
   - CMake version: 3.31.3

## Description of the problem:

Blend Mode isn't working, since **"It's not implemented"**. Inside the source file "GObject.cpp" we can find that the BlendMode is able to load, YET ignored ultimately(on line 872).

Image

## Steps to Reproduce:

It's easy to reproduce this bug, by creating an Object with BlendMode in a component, and loading it in Axmol Engine.

1. Open FairyGUI, and create an Object with any kinds of BlendMode;
2. Save our project, publish it, and normally load it afterwards in Axmol Engine;
3. Since BlendMode is ignored on loading the package, **every object using BlendMode looks the same as the one without using BlendMode**.

Here's how it presented in my computer.

1. Create a GGraph "test" with BlendMode "Add", color "#00CC00" inside a GComponent named "Dashboard", a member of the package "SearchScene";

Image

2. Save and publish the package normally.
3. Load the package in Axmol Engine, and create "Dashboard":

```cpp
void SearchScene::init(){
    ...
    UIPackage::addPackage("SearchScene");
    _dashboard=UIPackage::createObject("SearchScene", "Dashboard")
        ->as<GComponent>();
    _GROOT->addChild(_dashboard);
    ...
}
```

4. Compile and run the project, and the GGraph isn't using BlendMode.

Image

## Another Issue:

Stroke for GTextField isn't working, returning **"Stroke Currently only supported on iOS and Android!"** in console. This is occurred on loading/changing RichText.

Image
Image

rh101:
## Describe your changes

 `DrawNode` now inherits from `BlendProtocol`, since it does support blending.  Also, ensure that the set `BlendFunc` is actually used, because it was hard-coded previously, so never used the correct blending modes. The change should not affect existing usage at all, since the default `BlendFunc` was `BlendFunc::ALPHA_PREMULTIPLIED`, which is `{backend::BlendFactor::ONE, backend::BlendFactor::ONE_MINUS_SRC_ALPHA}`, the same as what was being hard-coded:

```
blendDescriptor.sourceRGBBlendFactor        = backend::BlendFactor::ONE;
blendDescriptor.destinationRGBBlendFactor   = backend::BlendFactor::ONE_MINUS_SRC_ALPHA;
blendDescriptor.sourceAlphaBlendFactor      = backend::BlendFactor::ONE;
blendDescriptor.destinationAlphaBlendFactor = backend::BlendFactor::ONE_MINUS_SRC_ALPHA;
```
Which is now:
```
blendDescriptor.sourceRGBBlendFactor        = _blendFunc.src;
blendDescriptor.destinationRGBBlendFactor   = _blendFunc.dst;
blendDescriptor.sourceAlphaBlendFactor      = _blendFunc.src;
blendDescriptor.destinationAlphaBlendFactor = _blendFunc.dst;
```

Nothing else changes, as this is primarily an enhancement to allow blending when using FairyGUI components.  The FairyGUI code changes are based of FairyGUI libraries for other engines, so the API is similar to their other implementations.

#2486 

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/extensions/fairygui/src/fairygui/display/BlendMode.cpp
+++ b/extensions/fairygui/src/fairygui/display/BlendMode.cpp
@@ -1,0 +1,65 @@
+#include "BlendMode.h"
+#include "renderer/backend/Enums.h"
+
+NS_FGUI_BEGIN
+using namespace ax;
+
+namespace
+{
+std::vector<BlendFunc> blendModes {
+    {backend::BlendFactor::SRC_ALPHA, backend::BlendFactor::ONE_MINUS_SRC_ALPHA},  // normal
+    {backend::BlendFactor::ONE, backend::BlendFactor::ONE},                        // none
+    {backend::BlendFactor::SRC_ALPHA, backend::BlendFactor::ONE},                  // add
+    {backend::BlendFactor::DST_COLOR, backend::BlendFactor::ONE_MINUS_SRC_ALPHA},  // mul
+    {backend::BlendFactor::ONE, backend::BlendFactor::ONE_MINUS_SRC_COLOR},        // screen
+    {backend::BlendFactor::ZERO, backend::BlendFactor::ONE_MINUS_SRC_ALPHA},       // erase
+    {backend::BlendFactor::ZERO, backend::BlendFactor::SRC_ALPHA},                 // mask
+    {backend::BlendFactor::ONE_MINUS_DST_ALPHA, backend::BlendFactor::DST_ALPHA},  // below
+    {backend::BlendFactor::ONE, backend::BlendFactor::ZERO},                       // off
+    {backend::BlendFactor::SRC_ALPHA, backend::BlendFactor::ONE_MINUS_SRC_ALPHA},  // custom1
+    {backend::BlendFactor::SRC_ALPHA, backend::BlendFactor::ONE_MINUS_SRC_ALPHA},  // custom2
+    {backend::BlendFactor::SRC_ALPHA, backend::BlendFactor::ONE_MINUS_SRC_ALPHA},  // custom2
+};
+}
+
+void BlendModeUtils::apply(ax::Node* node, BlendMode blendMode)
+{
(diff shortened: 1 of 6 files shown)


extensions/fairygui/src/fairygui/display/BlendMode.cpp
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    auto&& children = node->getChildren();
+    for (auto&& child : children)
+    {
+        apply(child, blendFunc);
+    }
+}
+
+void BlendModeUtils::override(BlendMode blendMode,
```
halx99: should avoid use c++ keyword as symbol, maybe rename to `update`

halx99 (changes requested):

extensions/fairygui/src/fairygui/display/BlendMode.cpp
rh101: Yes, I noticed that, since I was just following their current implementation from [here](https://github.com/fairygui/FairyGUI-cocoscreator/blob/ccc3.0/source/src/display/BlendMode.ts).

I'll change it now to avoid the confusion.

rh101: Actually, how about `overrideBlendMode`, since it states the purpose of the method?

halx99: > Actually, how about `overrideBlendMode`, since it states the purpose of the method?

it's ok

PR #2551, opened 2025-05-25
Update build.gradle - simplify, remove coping and pointing directly to assets folders

crazyhappygame:
## Describe your changes

simplify gradle script and removing copy assests functionality

If there is needed for different assets for different platform configuration having separate folder(s) is one of the options to achieve that (assestsAndroid/asssestWin etc)

### For each PR
       
   Optional:
          
### For core/new feature PR

### DIFF
--- a/templates/common/proj.android/app/build.gradle
+++ b/templates/common/proj.android/app/build.gradle
@@ -44,7 +44,7 @@ android {
         java.srcDir "src"
         res.srcDir "res"
         manifest.srcFile "AndroidManifest.xml"
-        assets.srcDirs = ["../../Content", "${projectDir}/build/runtime/axslc"]
+        assets.srcDirs = ["../../Content", "${projectDir}/build/runtime"]
     }
 
     externalNativeBuild {
--- a/templates/common/proj.android/app/build.gradle
+++ b/templates/common/proj.android/app/build.gradle
@@ -44,7 +44,7 @@ android {
         java.srcDir "src"
         res.srcDir "res"
         manifest.srcFile "AndroidManifest.xml"
-        assets.srcDir "build/assets"
+        assets.srcDirs = ["../../Content", "${projectDir}/build/runtime/axslc"]
     }
 
     externalNativeBuild {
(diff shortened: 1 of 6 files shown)


rh101:
This PR is not correct unless you have a way to exclude files.  For instance, the existing implementation handles this in the copy command:
```
copy {
    from "${projectDir}/../../Content"
    into "${projectDir}/build/assets"
    exclude "**/*.gz"
}
```

crazyhappygame:
> This PR is not correct unless you have a way to exclude files. For instance, the existing implementation handles this in the copy command:
> 
> ```
> copy {
>     from "${projectDir}/../../Content"
>     into "${projectDir}/build/assets"
>     exclude "**/*.gz"
> }
> ```

Could you explain to me why do we need any exclude rule if there is no *.gz files in Content folder?

rh101:
> Could you explain to me why do we need any exclude rule if there is no *.gz files in Content folder?

It's not about the `*.gz` files, but it's about the ability to exclude files from the assets folder for Android.  For example, developers may be using this to exclude files that are not for Android, or files that are just for certain builds, such as files they may not want in the release build etc. etc..

crazyhappygame:
I see your point but ... I would challenge that.
I think that engine should provide as simple as possible configuration by default. 
By default games should use all resoruces for all platforms. (that what cpp tests does at the moment and all other tests)

If there is needed for different assets for different platform configuration I would say that having separate folder(s) is the simplest way to go (assestsAndroid/asssestWin etc).

I there is need for more sophisticated setup then maybe even exclude could be not enough.

Feel free to close this PR if you do not like this simplification

rh101:
> I see your point but ... I would challenge that. I think that engine should provide as simple as possible configuration by default. By default games should use all resoruces for all platforms. (that what cpp tests does at the moment and all other tests)
> 

I understand, and I do agree with that, but from my perspective, it's more about existing projects using Axmol that may utilize the exclude functionality.  Providing an alternate solution while also cleaning up the existing implementation would be the way forward.

> If there is needed for different assets for different platform configuration I would say that having separate folder(s) is the simplest way to go (assestsAndroid/asssestWin etc).

This is a reasonable solution, so mentioning it in this PR would help any developers that need this functionality.

> Feel free to close this PR if you do not like this simplification

It's not about liking or disliking changes, but about considering the impact that they may have.  It always helps to add some details to the PR as to why the change is happening, and possible alternatives for changes in the functionality.  If someone were to look up this PR to see why the changes were made, at least then they would understand the reasoning behind it.

crazyhappygame:
fair points. Updated title and description

crazyhappygame:
Hmm it looks like that we really do have *.gz files in assests
https://github.com/axmolengine/axmol/actions/runs/15237605216/job/42853547053?pr=2551
```
 The input changes require a full rebuild for incremental task ':CppTests:processReleaseResources'.
Aapt output file /home/runner/work/axmol/axmol/tests/cpp-tests/proj.android/app/build/intermediates/linked_resources_proto_format/release/processReleaseResources/linked-resources-proto-format-release.ap_
ERROR: [Images/test_image_rgba4444.pvr] /home/runner/work/axmol/axmol/tests/cpp-tests/Content/Images/test_image_rgba4444.pvr [Images/test_image_rgba4444.pvr] /home/runner/work/axmol/axmol/tests/cpp-tests/Content/Images/test_image_rgba4444.pvr.gz: Resource and asset merger: Duplicate resources

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':CppTests:mergeReleaseAssets'.
> [Images/test_image_rgba4444.pvr] /home/runner/work/axmol/axmol/tests/cpp-tests/Content/Images/test_image_rgba4444.pvr	[Images/test_image_rgba4444.pvr] /home/runner/work/axmol/axmol/tests/cpp-tests/Content/Images/test_image_rgba4444.pvr.gz: Error: Duplicate resources
```
I am will close this PR

PR #2553, opened 2025-05-28
Add build options to game cmake module for easier access

rh101:
## Describe your changes

Every so often questions are asked related to turning off extensions and code sections that are not used.  The default options are not well documented, so developers would need to hunt down all the options that they can set.  This PR introduces a more centralized and accessible way for developers to enable and disable code that they do not require,

The default settings in the new file should be the same as what is currently set in the engine cmake files, but if they are not, then let me know to update them.

### For each PR
       
   Optional:
          
### For core/new feature PR


halx99:
The modifications make developer can't control cmake option at cmdline due to you set all options   
variables to cache force.

rh101:
> The modifications make developer can't control cmake option at cmdline due to you set all options variables to cache force.

I didn't think of that.  In that case, it would be best to have all the options overrides commented out, just so developers are aware of the available options.  I'll change it now.

Would that suffice?

PR #2575, opened 2025-06-23
[WASM] change wasm initial memory to 100MB, allow memory increase

AlexandreK38:
With EMSCRIPTEN, the `INITIAL_MEMORY` is the amount of bytes that will be required at launch in the browser no matter it is needed or not, and the app won't be able to ask for more than this amount too. While it works on high performance devices, it won't on medium or low devices because they don't have that much memory.
The solution is to lower the initial value (to 100MB which seems reasonable), and allow the memory to grow by using `ALLOW_MEMORY_GROWTH`.

Note: it will be allowed to grow until `MAXIMUM_MEMORY` which is 2GB by default


cmake/Modules/AXConfigDefine.cmake
```diff
@@ -224,8 +224,8 @@ if(EMSCRIPTEN)
         list(APPEND _AX_EM_LD_FLAGS -pthread -sPTHREAD_POOL_SIZE=${AX_WASM_THREADS})
     endif()
 
-    set(AX_WASM_INITIAL_MEMORY "1024MB" CACHE STRING "")
-    list(APPEND _AX_EM_LD_FLAGS -sINITIAL_MEMORY=${AX_WASM_INITIAL_MEMORY})
+    set(AX_WASM_INITIAL_MEMORY "100MB" CACHE STRING "")
```
halx99: why 100MB? maybe 128MB?

AlexandreK38: I don't mind putting 128, I'm not sure why a power of two would change much :)

PR #2578, opened 2025-06-26
[WASM] Option to use main loop timing 'timeout' instead of 'requestAnimationFrame' (RAF)

AlexandreK38:
In Webgl , there are several timings to run the main loop. Current one implemented in Axmol is the `requestAnimationFrame`  one (RAF) but if the app loses focus the loop is paused.
If we want the app to run with or without focus, for example in a multiplayer game, we must use the `timeout` timing.

This PR adds the `AX_WASM_TIMING_USE_TIMEOUT` option to use the timeout with the FPS set by `Application::setAnimationInterval`.


halx99 (approved):

core/platform/wasm/Application-wasm.cpp
```diff
@@ -196,7 +196,12 @@ int Application::run()
     as the browser’s requestAnimationFrame will make sure you render at a proper smooth rate
     that lines up properly with the browser and monitor.
     */
+#if AX_WASM_TIMING_USE_TIMEOUT
+    double fps = ceil(1.0 / (static_cast<double>(s_animationInterval) / NANOSECONDSPERSECOND));
+    emscripten_set_main_loop(updateFrame, static_cast<int>(fps), false);
+#else
     emscripten_set_main_loop(updateFrame, -1, false);
+#endif
```
halx99, line 205: ### ！，，

PR #2580, opened 2025-06-26
Add mouse events listener for Widgets

AlexandreK38:
Hi,
I wanted to add a mouse scroll listener on the UIScrollView to be able to scroll with the mouse. Basically it works, but if we have several scroll views then they will all move, even if the mouse is not over them.

To be able to have only the top widget to handle the mouse scroll and not others, we need to implement the same touches events mechanism but for mouse events. So this will work for scroll event, but also press, released, and moved.
We can also, as for touches, add a listener to our own widgets.

### DIFF
--- a/core/base/EventDispatcher.cpp
+++ b/core/base/EventDispatcher.cpp
@@ -1222,8 +1222,6 @@ void EventDispatcher::dispatchMouseEvent(EventMouse* event)
     if (nullptr == listeners)
         return;
 
-    bool isSwallowed = false;
-
     auto onMouseEvent = [&](EventListener* l) -> bool {  // Return true to break
         EventListenerMouse* listener = static_cast<EventListenerMouse*>(l);
 
--- a/core/base/EventDispatcher.cpp
+++ b/core/base/EventDispatcher.cpp
@@ -975,6 +975,11 @@ void EventDispatcher::dispatchEvent(Event* event, bool forced)
         dispatchTouchEvent(static_cast<EventTouch*>(event));
         return;
     }
+    else if (event->getType() == Event::Type::MOUSE)
+    {
+        dispatchMouseEvent(static_cast<EventMouse*>(event));
+        return;
+    }
 
     auto listenerID = __getListenerID(event);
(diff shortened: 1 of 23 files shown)


AlexandreK38:
I forgot about the extensions, I'll need to fix them with the new code

core/base/EventDispatcher.cpp
```diff
@@ -1207,6 +1212,88 @@ void EventDispatcher::dispatchTouchEvent(EventTouch* event)
...
+{
+    sortEventListeners(EventListenerMouse::LISTENER_ID);
+
+    auto listeners = getListeners(EventListenerMouse::LISTENER_ID);
+
+    // If there aren't any mouse listeners, return directly.
+    if (nullptr == listeners)
+        return;
+
+    bool isSwallowed = false;
```
halx99: The local variable `isSwallowed` unused

core/base/EventListenerMouse.h
```diff
@@ -54,17 +54,33 @@ class AX_DLL EventListenerMouse : public EventListener
...
+
     /// Overrides
     virtual EventListenerMouse* clone() override;
     virtual bool checkAvailable() override;
 
-    std::function<void(EventMouse* event)> onMouseDown;
-    std::function<void(EventMouse* event)> onMouseUp;
-    std::function<void(EventMouse* event)> onMouseMove;
-    std::function<void(EventMouse* event)> onMouseScroll;
+    std::function<bool(EventMouse* event)> onMouseDown;
```
halx99, line 72: could you explain the return type means?

halx99 (approved):

core/base/EventListenerMouse.h
AlexandreK38: It's exactly the same as the onTouchBegan for the touches
We return true if the user want to handle the mouse event, false otherwise
As the mouse events are not dépendant between themselves we need the bool for all, as opposed to the touch where you only handle the began

halx99 (approved):

PR #2581, opened 2025-06-28
Fixed statsLabel not show after Director::restart()

remtori:
I recently have a use case where i need to do Director::restart() and notice that the statsLabel is missing after a restart, after digging around a little bit its seem the issue is we dont set  _isStatusLabelUpdated so the statsLabel never get recreated.

### DIFF
--- a/core/base/Director.cpp
+++ b/core/base/Director.cpp
@@ -1048,6 +1048,7 @@ void Director::reset()
     AX_SAFE_RELEASE_NULL(_FPSLabel);
     AX_SAFE_RELEASE_NULL(_drawnBatchesLabel);
     AX_SAFE_RELEASE_NULL(_drawnVerticesLabel);
+    _isStatusLabelUpdated = true;
 
     // purge bitmap cache
     FontFNT::purgeCachedData();


core/base/Director.cpp
```diff
@@ -1077,13 +1078,6 @@ void Director::cleanupDirector()
     reset();
 
     // cleanup graphics before release glView, otherwise, will cause crash on linux
-    AX_SAFE_RELEASE(_FPSLabel);
```
halx99: Why remove these release statements?

remtori: You can see that this function call reset() which already release all of those already, so these release statements is kind of redundant.

However, if you dont want unrelated code change in 1 PR, i totally understand and will do this in a seperate commit / PR.

rh101: While you're at it, `_notificationNode` is being released and null assigned twice in `reset()`:

```
void Director::reset()
{
...
    if (_notificationNode)
    {
        _notificationNode->onExit();
        _notificationNode->cleanup();
        _notificationNode->release();
    }

    _notificationNode = nullptr;

...
    stopAnimation();    

    AX_SAFE_RELEASE_NULL(_notificationNode);
...
}
```

The `AX_SAFE_RELEASE_NULL(_notificationNode);` line can be safely deleted.

halx99: > You can see that this function call reset() which already release all of those already, so these release statements is kind of redundant.
> 
> However, if you dont want unrelated code change in 1 PR, i totally understand and will do this in a seperate commit / PR.

Yes, revert unrelated changes is better

remtori: I removed the unrelated changes

halx99 (approved):

halx99:
Thanks, merged, and I push the commit: https://github.com/axmolengine/axmol/commit/4eab68f73dfead8ac7463355f883247b86c954bf to remove redundant release statements

PR #2654, opened 2025-08-27
Auto execute clang-format lint for per pull-request

halx99:
## Describe your changes

Finally: The clang-format lint action in `dev/v3` branch support lint_mode: `check_only`, `create_pr`, `auto_commit`, and every pull request is `check_only`

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,6 +1,8 @@
 name: clang-format
 
 on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
   workflow_dispatch:
     inputs:
       commit_changes:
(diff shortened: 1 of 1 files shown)


paulocoutinhox:
I think that `inplace` need be `False`, since the CI only will check if it is formatted and fail. The `inplace` will change files.

aismann:
Running cpp-tests (Windows works!)
Looked on some diffs ( I like the new code style, same look and "feel" on all Axmol source files)

aismann (commented):
looks good for me

halx99:
> I think that `inplace` need be `False`, since the CI only will check if it is formatted and fail. The `inplace` will change files.

I change to lint only for PR, the clang-format lint action also support auto commit changes when trigger manually

PR #2676, opened 2025-09-01
Fix wasm downloader

AlexandreK38:
## Describe your changes

The Wasm downloader doesn't call any of the callbacks because there is an issue in emscripten with the id of the fetch given by `emscripten_fetch()` function, it always return 0 - except for one case not used in our downloader - You can learn more here: https://github.com/emscripten-core/emscripten/issues/23124)

The solution is to use the fetch address instead of the id for the context map. As said in the issue thread, the address could be reused after the download is finished but in our case we remove the entry from the map once done too, so it's not problematic.

I also fixed the data of the task progress info as they weren't set at all


halx99 (approved):

halx99:
lgtm, thanks

PR #2719, opened 2025-09-12
Add support for audio panning

rh101:
## Describe your changes

 Add support for audio panning relative to the listener, from left to right, -60° to +60°.

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/core/audio/AudioEngineImpl.cpp
+++ b/core/audio/AudioEngineImpl.cpp
@@ -936,14 +936,22 @@ void AudioEngineImpl::setPan(AUDIO_ID audioId, float value)
     }
 }
 
+float AudioEngineImpl::getPan(int audioId)
+{
+    std::unique_lock<std::recursive_mutex> lck(_threadMutex);
+    auto iter = _audioPlayers.find(audioId);
+    if (iter == _audioPlayers.end())
+        return 0.f;
+
+    auto player = iter->second;
+    lck.unlock();
+
+    return player->_pan;
+}
+
 bool AudioEngineImpl::isExtensionPresent(const char* extensionId)
 {
-    if (alIsExtensionPresent(extensionId))
-    {
-        return true;
-    }
-
-    return false;
+    return alIsExtensionPresent(extensionId);
 }
 
 void AudioEngineImpl::checkExtensions()
--- a/core/audio/AudioEngineImpl.cpp
+++ b/core/audio/AudioEngineImpl.cpp
@@ -931,7 +931,7 @@ void AudioEngineImpl::setPan(AUDIO_ID audioId, float value)
         panAngles[0] = (1.0f - player->_pan) * angle;
         panAngles[1] = (1.0f + player->_pan) * -angle;
 
-        alSourcei(player->_alSource, AL_SOURCE_RELATIVE, TRUE); // relative to listener
+        alSourcei(player->_alSource, AL_SOURCE_RELATIVE, AL_TRUE); // relative to listener
         alSourcefv(player->_alSource, AL_STEREO_ANGLES, panAngles);
     }
 }
--- a/core/audio/AudioEngineImpl.cpp
+++ b/core/audio/AudioEngineImpl.cpp
@@ -932,7 +932,7 @@ void AudioEngineImpl::setPan(AUDIO_ID audioId, float value)
         panAngles[1] = (1.0f + player->_pan) * -angle;
 
         alSourcei(player->_alSource, AL_SOURCE_RELATIVE, AL_TRUE); // relative to listener
-        alSourcefv(player->_alSource, AL_STEREO_ANGLES, panAngles);
+        alSourcefv(player->_alSource, 0x1030, panAngles); // AL_STEREO_ANGLES = 0x1030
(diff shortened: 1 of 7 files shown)


halx99:
will your also add 3d API: `int AudioEngine::play3d(const std::string& filePath, const Vec3& position, bool loop = false, float volume = 1.0f, const AudioProfile* profile = nullptr);` ?

halx99:
> will your also add 3d API: `int AudioEngine::play3d(const std::string& filePath, const Vec3& position, bool loop = false, float volume = 1.0f, const AudioProfile* profile = nullptr);` ?

never mind, the `audio panning` still for 2d, right?

rh101:
> > will your also add 3d API: `int AudioEngine::play3d(const std::string& filePath, const Vec3& position, bool loop = false, float volume = 1.0f, const AudioProfile* profile = nullptr);` ?
> 
> never mind, the `audio panning` still for 2d, right?

This implementation is only for 2D. Do you want me to try and add 3D panning to this PR?

halx99:
> will your also add 3d API: `int AudioEngine::play3d(const std::string& filePath, const Vec3& position, bool loop = false, float volume = 1.0f, const AudioProfile* profile = nullptr);` ?



> > > will your also add 3d API: `int AudioEngine::play3d(const std::string& filePath, const Vec3& position, bool loop = false, float volume = 1.0f, const AudioProfile* profile = nullptr);` ?
> > 
> > 
> > never mind, the `audio panning` still for 2d, right?
> 
> This implementation is only for 2D. Do you want me to try and add 3D panning to this PR?

Maybe a separated PR is better due to current only have play2d API

rh101:
> Maybe a separated PR is better due to current only have play2d API

I can take a look at adding `play3d` sometime next week.

core/audio/AudioEngineImpl.cpp
```diff
@@ -907,6 +909,48 @@ void AudioEngineImpl::update(float /*dt*/)
...
+        panAngles[1] = (1.0f + player->_pan) * -angle;
+
+        alSourcei(player->_alSource, AL_SOURCE_RELATIVE, AL_TRUE); // relative to listener
+        alSourcefv(player->_alSource, 0x1030, panAngles); // AL_STEREO_ANGLES = 0x1030
+    }
+}
+
+bool AudioEngineImpl::isExtensionPresent(const char* extensionId)
+{
+    if (alIsExtensionPresent(extensionId))
```
halx99: just `return alIsExtensionPresent(extensionId)` ?

halx99 (changes requested):

core/audio/AudioEngineImpl.cpp
rh101: Yep I'll update that and also add a `getPan()` method to get the current value

rh101:
Using a value of `-0.5` for the Z position, instead of `-1.f`, seems to give a better result for 2D audio panning, where it seems to completely silence the opposite side.

So this:
`alSource3f(player->_alSource, AL_POSITION, player->_pan, 0.0f, -1.0f);`
changes to:
`alSource3f(player->_alSource, AL_POSITION, player->_pan, 0.0f, -0.5f);`

Instead of hard-coding it, how about we introduce a new parameter to the `setPan` method, "distance", which is actually the Z value:
`void AudioEngineImpl::setPan(AUDIO_ID audioId, float value, float distance = -0.5f)`

so it would end up as;
`alSource3f(player->_alSource, AL_POSITION, player->_pan, 0.0f, distance);`

This way the developer can set it for the specific panning effect they want if the default is not suitable.

Also, we should add a method to get the current audio source position, such as:

`ax::Vec3 AudioEngineImpl::getSourcePosition(int audioId)`

This would cover both 2D and 3D audio.

@halx99 What is your opinion of this?

halx99:
If ax::Vec3 AudioEngineImpl::getSourcePosition(int audioId) added, shoud we also add setSourcePosition ?

rh101:
> If ax::Vec3 AudioEngineImpl::getSourcePosition(int audioId) added, shoud we also add setSourcePosition ?

Yes, I'll add that too.

rh101:
@halx99 I'm not quite sure why the iOS and OSX tests are failing, unless it's something I'm not seeing in the logs.

halx99:
> @halx99 I'm not quite sure why the iOS and OSX tests are failing, unless it's something I'm not seeing in the logs.

here: /Users/runner/work/axmol/axmol/core/audio/AudioPlayer.h:78:5: error: unknown type name 'Vec3'
   78 |     Vec3 _sourcePosition; 
   
   seems you miss include "math/Vec3.h" in `AudioPlayer.h`

rh101:
> here: /Users/runner/work/axmol/axmol/core/audio/AudioPlayer.h:78:5: error: unknown type name 'Vec3' 78 | Vec3 _sourcePosition;
> 
> seems you miss include "math/Vec3.h" in `AudioPlayer.h`

Thank you. I totally missed that in the logs.

halx99 (approved):

PR #2727, opened 2025-09-12
added onJustKeyPressed event

rudiHammad:
## Describe your changes
Implemented a new keyboard event in EventListenerKeyboard.h and EventListenerKeyboard.cpp
To the existing:
```
    std::function<void(EventKeyboard::KeyCode, Event*)> onKeyPressed;
    std::function<void(EventKeyboard::KeyCode, Event*)> onKeyReleased;
```
I added 
`std::function<void(EventKeyboard::KeyCode, Event*)> onKeyJustPressed;`

Unlike onKeyPressed, which fires continuously while a key is held down, onKeyJustPressed triggers only once at the moment the key is pressed. This simplifies typical gameplay code (e.g., jumping, shooting) by removing the need for extra boolean flags to prevent repeated actions.

### DIFF
--- a/axmol/base/EventListenerKeyboard.cpp
+++ b/axmol/base/EventListenerKeyboard.cpp
@@ -34,7 +34,7 @@ const std::string EventListenerKeyboard::LISTENER_ID = "__ax_keyboard";
 
 bool EventListenerKeyboard::checkAvailable()
 {
-    if (onKeyPressed == nullptr && onKeyReleased == nullptr)
+    if (onKeyPressed == nullptr && onKeyReleased == nullptr && onKeyJustPressed == nullptr)
     {
         AXASSERT(false, "Invalid EventListenerKeyboard!");
         return false;
@@ -63,8 +63,9 @@ EventListenerKeyboard* EventListenerKeyboard::clone()
     if (ret->init())
     {
         ret->autorelease();
-        ret->onKeyPressed  = onKeyPressed;
-        ret->onKeyReleased = onKeyReleased;
+        ret->onKeyPressed       = onKeyPressed;
+        ret->onKeyReleased      = onKeyReleased;
+        ret->onKeyJustPressed   = onKeyJustPressed;
     }
     else
     {
@@ -73,30 +74,42 @@ EventListenerKeyboard* EventListenerKeyboard::clone()
     return ret;
 }
 
-EventListenerKeyboard::EventListenerKeyboard() : onKeyPressed(nullptr), onKeyReleased(nullptr) {}
+EventListenerKeyboard::EventListenerKeyboard() : onKeyPressed(nullptr), onKeyReleased(nullptr),  onKeyJustPressed(nullptr) {}
+
 
 bool EventListenerKeyboard::init()
 {
     auto listener = [this](Event* event) {
-        auto keyboardEvent = static_cast<EventKeyboard*>(event);
+        auto keyboardEvent             = static_cast<EventKeyboard*>(event);
+        EventKeyboard::KeyCode keyCode = keyboardEvent->_keyCode;
         if (keyboardEvent->_isPressed)
         {
+            if (!keyStates[keyCode])
+            {                                     // Key was not pressed before
+                justPressedKeys[keyCode] = true;  // Mark as just pressed
+                if (onKeyJustPressed != nullptr)
+                {
+                    onKeyJustPressed(keyCode, event);
+                }
+            }
+            keyStates[keyCode] = true;  // Mark key as pressed
             if (onKeyPressed != nullptr)
-                onKeyPressed(keyboardEvent->_keyCode, event);
+            {
+                onKeyPressed(keyCode, event);
+            }
         }
         else
-        {
+        {  // Key released
+            keyStates[keyCode]       = false;
+            justPressedKeys[keyCode] = false;  // Reset just pressed on release
             if (onKeyReleased != nullptr)
-                onKeyReleased(keyboardEvent->_keyCode, event);
+            {
+                onKeyReleased(keyCode, event);
+            }
         }
     };
 
-    if (EventListener::init(Type::KEYBOARD, LISTENER_ID, listener))
-    {
-        return true;
-    }
-
-    return false;
+    return EventListener::init(Type::KEYBOARD, LISTENER_ID, listener);
 }
 
 }  // namespace ax
--- a/axmol/base/EventListenerKeyboard.h
+++ b/axmol/base/EventListenerKeyboard.h
@@ -59,8 +59,14 @@ public:
 
     std::function<void(EventKeyboard::KeyCode, Event*)> onKeyPressed;
     std::function<void(EventKeyboard::KeyCode, Event*)> onKeyReleased;
+    std::function<void(EventKeyboard::KeyCode, Event*)> onKeyJustPressed;
     EventListenerKeyboard();
     bool init();
+
+private:
+    std::unordered_map<EventKeyboard::KeyCode, bool> keyStates;
+    std::unordered_map<EventKeyboard::KeyCode, bool> justPressedKeys;
+
 };
 
 }  // namespace ax


halx99:
I prefer add a property to EventKeyboard indicate whether keyRepeat

rudiHammad:
> I prefer add a property to EventKeyboard indicate whether keyRepeat

fair enough. The point was precisely to get it working out of the box without having to add anything extra.
You can still add a property as you use to, this simply implements an extra functionality. But if no one else find it you use
we can cancel the PR.

halx99:
Or you don't need add custom map to save key state, the glfw keydown callback function already have a parameter `action` to indicate whether the key press is repeat

rudiHammad:
> Or you don't need add custom map to save key state, the glfw keydown callback function already have a parameter `action` to indicate whether the key press is repeat

okey, I guess you are refereing to this `void key_callback(GLFWwindow* window, int key, int scancode, int action, int mods)`.
You mention that as if it is something obvious to implement, but I don't know how to bridge the gap from this to the implementation in axmol. That's why I did a map. It's probably me not having enough experience.
Anyways, if it something that no one else find useful it fine, we can cancel the PR.

rh101:
> > Or you don't need add custom map to save key state, the glfw keydown callback function already have a parameter `action` to indicate whether the key press is repeat
> 
> okey, I guess you are refereing to this `void key_callback(GLFWwindow* window, int key, int scancode, int action, int mods)`. You mention that as if it is something obvious to implement, but I don't know how to bridge the gap from this to the implementation in axmol. That's why I did a map. It's probably me not having enough experience. Anyways, if it something that no one else find useful it fine, we can cancel the PR.

If `action == GLFW_REPEAT`, then it indicates that the callback is a button being held down, so there is no need to keep track of the presses in a map.  Adding a field to `EventKeyboard` to indicate if it is a repeated press would allow the developer to check for this, without the need for a new callback like `onKeyJustPressed`. 

So, perhaps something like adding `_isRepeat` to `EventKeyboard` in `EventKeyboard.h`:

```
class AX_DLL EventKeyboard : public Event
{
public:
    /**
     * KeyCode The key (code).
     */
    enum class KeyCode
    {
...
...
    };

    /** Constructor.
     *
     * @param keyCode A given keycode.
     * @param isPressed True if the key is pressed.
     * @param isRepeat True if the key is repeated, meaning it is held down.
     */
    EventKeyboard(KeyCode keyCode, bool isPressed, bool isRepeat);

    bool isRepeat() const { return _isRepeat; }

private:
    KeyCode _keyCode;
    bool _isPressed;
    bool _isRepeat;

    friend class EventListenerKeyboard;
};
```

and implementation:
```
EventKeyboard::EventKeyboard(KeyCode keyCode, bool isPressed, bool isRepeat)
    : Event(Type::KEYBOARD), _keyCode(keyCode), _isPressed(isPressed), _isRepeat(isRepeat)
{}
```

Then change the event creation line in `RenderViewImpl::onGLFWKeyCallback` to indicate if it is repeating:

```
void RenderViewImpl::onGLFWKeyCallback(GLFWwindow* /*window*/, int key, int /*scancode*/, int action, int /*mods*/)
{
    // x-studio spec, for repeat press key support.
    EventKeyboard event(g_keyCodeMap[key], action, action == GLFW_REPEAT);
    auto dispatcher = Director::getInstance()->getEventDispatcher();
    dispatcher->dispatchEvent(&event);
...
...
}
```

rudiHammad:
> > > Or you don't need add custom map to save key state, the glfw keydown callback function already have a parameter `action` to indicate whether the key press is repeat
> > 
> > 
> > okey, I guess you are refereing to this `void key_callback(GLFWwindow* window, int key, int scancode, int action, int mods)`. You mention that as if it is something obvious to implement, but I don't know how to bridge the gap from this to the implementation in axmol. That's why I did a map. It's probably me not having enough experience. Anyways, if it something that no one else find useful it fine, we can cancel the PR.
> 
> If `action == GLFW_REPEAT`, then it indicates that the callback is a button being held down, so there is no need to keep track of the presses in a map. Adding a field to `EventKeyboard` to indicate if it is a repeated press would allow the developer to check for this, without the need for a new callback like `onKeyJustPressed`.
> 
> So, perhaps something like adding `_isRepeat` to `EventKeyboard` in `EventKeyboard.h`:
> 
> ```
> class AX_DLL EventKeyboard : public Event
> {
> public:
>     /**
>      * KeyCode The key (code).
>      */
>     enum class KeyCode
>     {
> ...
> ...
>     };
> 
>     /** Constructor.
>      *
>      * @param keyCode A given keycode.
>      * @param isPressed True if the key is pressed.
>      * @param isRepeat True if the key is repeated, meaning it is held down.
>      */
>     EventKeyboard(KeyCode keyCode, bool isPressed, bool isRepeat);
> 
> private:
>     KeyCode _keyCode;
>     bool _isPressed;
>     bool _isRepeat;
> 
>     friend class EventListenerKeyboard;
> };
> ```
> 
> and implementation:
> 
> ```
> EventKeyboard::EventKeyboard(KeyCode keyCode, bool isPressed, bool isRepeat)
>     : Event(Type::KEYBOARD), _keyCode(keyCode), _isPressed(isPressed), _isRepeat(isRepeat)
> {}
> ```
> 
> Then change the event creation line in `RenderViewImpl::onGLFWKeyCallback` to indicate if it is repeating:
> 
> ```
> void RenderViewImpl::onGLFWKeyCallback(GLFWwindow* /*window*/, int key, int /*scancode*/, int action, int /*mods*/)
> {
>     // x-studio spec, for repeat press key support.
>     EventKeyboard event(g_keyCodeMap[key], action, action == GLFW_REPEAT);
>     auto dispatcher = Director::getInstance()->getEventDispatcher();
>     dispatcher->dispatchEvent(&event);
> ...
> ...
> }
> ```
well, That is pretty much self-explanatory XD. okey, I'll try that later today. Thanks

rh101:
@halx99 The current callbacks both receive `Event` instead of `EventKeyboard`, which is a little strange.  I'm not sure if that was done for a specific purpose, but if not, would it be better to change the event parameters to be `EventKeyboard`?

So, this:
```
std::function<void(EventKeyboard::KeyCode, Event*)> onKeyPressed;
std::function<void(EventKeyboard::KeyCode, Event*)> onKeyReleased;
```
change to:
```
std::function<void(EventKeyboard::KeyCode, EventKeyboard*)> onKeyPressed;
std::function<void(EventKeyboard::KeyCode, EventKeyboard*)> onKeyReleased;
```

halx99:
> @halx99 The current callbacks both receive `Event` instead of `EventKeyboard`, which is a little strange. I'm not sure if that was done for a specific purpose, but if not, would it be better to change the event parameters to be `EventKeyboard`?
> 
> So, this:
> 
> ```
> std::function<void(EventKeyboard::KeyCode, Event*)> onKeyPressed;
> std::function<void(EventKeyboard::KeyCode, Event*)> onKeyReleased;
> ```
> 
> change to:
> 
> ```
> std::function<void(EventKeyboard::KeyCode, EventKeyboard*)> onKeyPressed;
> std::function<void(EventKeyboard::KeyCode, EventKeyboard*)> onKeyReleased;
> ```

The EventKeyboard is inherit from Event, I think it's ok.

rudiHammad:
> > @halx99 The current callbacks both receive `Event` instead of `EventKeyboard`, which is a little strange. I'm not sure if that was done for a specific purpose, but if not, would it be better to change the event parameters to be `EventKeyboard`?
> > So, this:
> > ```
> > std::function<void(EventKeyboard::KeyCode, Event*)> onKeyPressed;
> > std::function<void(EventKeyboard::KeyCode, Event*)> onKeyReleased;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > change to:
> > ```
> > std::function<void(EventKeyboard::KeyCode, EventKeyboard*)> onKeyPressed;
> > std::function<void(EventKeyboard::KeyCode, EventKeyboard*)> onKeyReleased;
> > ```
> 
> The EventKeyboard is inherit from Event, I think it's ok.

Okey, I did the changes and got it working, but I think that what @rh101 suggested is right (i don't know if it would break anything though), because unless I am missing something,  I had to cast the event into EventKeyboard like so:

```
void MySceneTest::onKeyPressed(EventKeyboard::KeyCode code, Event* event)
{
    keys[int(code)] = true;
    auto keyboardEvent = static_cast<EventKeyboard*>(event);
    if (code == EventKeyboard::KeyCode::KEY_SPACE && !keyboardEvent->isRepeat()) {AXLOG("jump once");}
}
```
which is not very intuitive. If we did  std::function<void(EventKeyboard::KeyCode, EventKeyboard*)> onKeyPressed;`
we could do instead this right?
```
void MySceneTest::onKeyPressed(EventKeyboard::KeyCode code, Event* event)
{
    keys[int(code)] = true;
    if (code == EventKeyboard::KeyCode::KEY_SPACE && !event->isRepeat()) {AXLOG("jump once");}
}
```

rh101:
> which is not very intuitive. If we did std::function<void(EventKeyboard::KeyCode, EventKeyboard*)> onKeyPressed;` we could do instead this right?
> 
> ```
> void MySceneTest::onKeyPressed(EventKeyboard::KeyCode code, Event* event)
> {
>     keys[int(code)] = true;
>     if (code == EventKeyboard::KeyCode::KEY_SPACE && !event->isRepeat()) {AXLOG("jump once");}
> }
> ```

As @halx99 noted, EventKeyboard inherits from Event, so it doesn't break anything to change those 2 callbacks to use EventKeyboard instead of Event.  If you do change them, update any code that uses those 2 events etc..  It may be better to close this PR and create a new PR for the all the new changes.

Also, it may be better to target `release/2.x` instead of `dev` for the new PR, since it would be nice to have this change as part of the next Axmol v2 release.

rudiHammad:
> > which is not very intuitive. If we did std::function<void(EventKeyboard::KeyCode, EventKeyboard*)> onKeyPressed;` we could do instead this right?
> > ```
> > void MySceneTest::onKeyPressed(EventKeyboard::KeyCode code, Event* event)
> > {
> >     keys[int(code)] = true;
> >     if (code == EventKeyboard::KeyCode::KEY_SPACE && !event->isRepeat()) {AXLOG("jump once");}
> > }
> > ```
> 
> As @halx99 noted, EventKeyboard inherits from Event, so it doesn't break anything to change those 2 callbacks to use EventKeyboard instead of Event. If you do change them, update any code that uses those 2 events etc.. It may be better to close this PR and create a new PR for the all the new changes.
> 
> Also, it may be better to target `release/2.x` instead of `dev` for the new PR, since it would be nice to have this change as part of the next Axmol v2 release.

Ok, we can close this pr, and I'll target release/2.x and use EventKeyboard in the callbacks. I will be out most of the day tomorrow but I'll try to do it as soon as can.

PR #2730, opened 2025-09-13
Improve the CMakeList.txt on templates folder

aismann:
## Describe your changes

Add subfolders to the studio:
image

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/templates/cpp/CMakeLists.txt
+++ b/templates/cpp/CMakeLists.txt
@@ -61,8 +61,11 @@ set(content_folder
   "${CMAKE_CURRENT_SOURCE_DIR}/Content"
 )
 
-# Find all header files recursively
-file(GLOB_RECURSE HEADER_FILES "${CMAKE_CURRENT_SOURCE_DIR}/Source/*.h")
+# Find all .h and .hpp files recursively
+file(GLOB_RECURSE HEADER_FILES
+  "${CMAKE_CURRENT_SOURCE_DIR}/Source/*.h"
+  "${CMAKE_CURRENT_SOURCE_DIR}/Source/*.hpp"
+)
 
 # Extract unique directories from those headers
 set(GAME_INC_DIRS "")
--- a/templates/cpp/CMakeLists.txt
+++ b/templates/cpp/CMakeLists.txt
@@ -61,6 +61,17 @@ set(content_folder
   "${CMAKE_CURRENT_SOURCE_DIR}/Content"
 )
 
+# Find all header files recursively
+file(GLOB_RECURSE HEADER_FILES "${CMAKE_CURRENT_SOURCE_DIR}/Source/*.h")
+
+# Extract unique directories from those headers
+set(GAME_INC_DIRS "")
+foreach(header ${HEADER_FILES})
+  get_filename_component(dir "${header}" DIRECTORY)
+  list(APPEND GAME_INC_DIRS "${dir}")
+endforeach()
+list(REMOVE_DUPLICATES GAME_INC_DIRS)
+
 if(APPLE)
   ax_mark_multi_resources(common_content_files RES_TO "Resources" FOLDERS ${content_folder})
 elseif(WINDOWS)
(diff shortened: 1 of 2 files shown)


rh101:
Please do not merge this change in. The changes in this PR will cause problems, such as filename clashes, if a file of the same name exists in multiple folders.

The only folder that should be in the include list is the "Source" folder, and not the sub-folders.  If a developer wants to have a specific folder with all header files in them, then that developer can easily add a folder, such as "Includes", and then add an explicit reference to that folder.

For the example above, if there are header files that should be referenced, then they should be relative to the "Source" folder, such as:

```
#include "Common/file1.h"
#include "Components/file2.h"
#include "Game/file3.h"
#include "Game/Level/file4.h"
```

halx99:
> Please do not merge this change in. The changes in this PR will cause problems, such as filename clashes, if a file of the same name exists in multiple folders.
> 
> The only folder that should be in the include list is the "Source" folder, and not the sub-folders. If a developer wants to have a specific folder with all header files in them, then that developer can easily add a folder, such as "Includes", and then add an explicit reference to that folder.
> 
> For the example above, if there are header files that should be referenced, then they should be relative to the "Source" folder, such as:
> 
> ```
> #include "Common/file1.h"
> #include "Components/file2.h"
> #include "Game/file3.h"
> #include "Game/Level/file4.h"
> ```

I agree

halx99 (commented):
Sorry, I can't merge this PR as @rh101 said.

aismann:
> Sorry, I can't merge this PR as @rh101 said.

I understand. No problem.

aismann:
@rh101 @halx99 
I think you missinterprete this PR.
Its add only the missed folder structure on the Visual Studio  "Additional Include directories".
See screenshot on first entry.

How often are you think will the same file on different folders?
- Renaming the files or
- Removing the folder on the "Additional Include directories" is often easier for the e.g. 2 files which be on different folders as add "the correct folder structure"  to each #include  (on each *.h and *.cpp file).


What are you mean?

Here a horrible example which Im try to port to Axmol (Thats the reason for this PR of course):
Hint: this folder structure comes with the current CMakeList.txt
image

aismann:
@rh101 @halx99 
What are you think if it will be optional?

halx99:
I think it's not good to add so much include dirs if I'm not misunderstand your logic is list all dirs of source files recursive.

aismann:
> I think it's not good to add so much include dirs if I not misunderstand your logic is list all dirs of source files recursive.

Ok I close it. I use it if I need the snippet for my private stuft (mostly for ports to axmol)

rh101:
> @rh101 @halx99 I think you missinterprete this PR. Its add only the missed folder structure on the Visual Studio "Additional Include directories". See screenshot on first entry.
> 

I did not misunderstand the implementation.  Adding folders to the include list, as clearly shown in the original post, "Additional Include Directories", is not correct, at least not for the issue you're describing.

> How often are you think will the same file on different folders?
> 

The example I gave of having files that are the same name was just one reason, and adding something like this should not be done in the template.

> Here a horrible example which Im try to port to Axmol (Thats the reason for this PR of course): Hint: this folder structure comes with the current CMakeList.txt 

You will need to handle this in a different way.  For example, try adding a separate `CMakeLists.txt` to each sub-folder, and explicitly listing the files inside it.  Reference those `CMakeLists.txt` files from the parent `CMakeLists.txt`, and do not use `GLOB_RECURSE`.  The files and folders should all show up correctly in the VS project, with the correct structure.

For example, in the root `CMakeLists.txt`:
```
set(GAME_HEADER)
set(GAME_SOURCE)

include(${CMAKE_CURRENT_LIST_DIR}/Source/CMakeLists.txt)
```

In `Source/CMakeLists.txt`:
```
include(${CMAKE_CURRENT_LIST_DIR}/TestFolder/CMakeLists.txt)

list(APPEND GAME_SOURCE
  ${CMAKE_CURRENT_LIST_DIR}/AppDelegate.cpp
  ${CMAKE_CURRENT_LIST_DIR}/MainScene.cpp
  )
	 
list(APPEND GAME_HEADER
  ${CMAKE_CURRENT_LIST_DIR}/AppDelegate.h
  ${CMAKE_CURRENT_LIST_DIR}/MainScene.h
  )
```

In `Source/TestFolder/CMakeLists.txt`:
```
list(APPEND GAME_SOURCE
  ${CMAKE_CURRENT_LIST_DIR}/SomeFile.cpp
  )
	 
list(APPEND GAME_HEADER
  ${CMAKE_CURRENT_LIST_DIR}/SomeFile.h
  )
```

When you generate the project, this is what you will see:
image

PR #2740, opened 2025-09-14
Add support for playing audio at any position in 3D space

rh101:
## Describe your changes

This is initial support for playing a sound at a specific position in 3D space.  It will be relative to the listener.  The position of the listener can also be set, allowing the listener position to be modified, such as when a player object is moving around.

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/core/audio/AudioEngineImpl.cpp
+++ b/core/audio/AudioEngineImpl.cpp
@@ -585,7 +585,12 @@ AUDIO_ID AudioEngineImpl::play2d(std::string_view filePath, bool loop, float vol
     return _currentAudioID;
 }
 
-int AudioEngineImpl::play3d(std::string_view filePath, const Vec3& position, bool loop, float volume, float time)
+int AudioEngineImpl::play3d(std::string_view filePath,
+                            const Vec3& position,
+                            float distanceScale,
+                            bool loop,
+                            float volume,
+                            float time)
 {
     if (s_ALDevice == nullptr)
     {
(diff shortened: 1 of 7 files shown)


halx99:
What about let `play2d` invoke `play3d` with positon: `Vec3::ZERO` ?

rh101:
> What about let `play2d` invoke `play3d` with positon: `Vec3::ZERO` ?

I did do that originally, but then realized that there would be other things to add to the 3d audio, such as the velocity, Doppler effects etc., so it would need its own method to handle all those things, since they cannot be applied to the 2D audio.  

It may still be possible to reduce the code duplication significantly with refactoring, but for the time being, perhaps we just get an initial implementation working, and then improve it later?

halx99:
> perhaps we just get an initial implementation working, and then improve it later?

It's ok

rh101:
The cpp-tests Scene3D test seemed to be a good place to add a test for this, so if you move the character around, you can tell that the position of the audio changes correctly.

This implementation seems to be working well enough, and I'm sure in the future we can add more functionality to it.  If there is nothing else, then this PR should be ready.

core/audio/AudioEngineImpl.cpp
```diff
@@ -613,6 +672,34 @@ void AudioEngineImpl::_play2d(AudioCache* cache, AUDIO_ID audioID)
...
+                if (AudioEngine::_audioIDInfoMap.find(audioID) != AudioEngine::_audioIDInfoMap.end())
+                {
+                    AudioEngine::_audioIDInfoMap[audioID].state = AudioEngine::AudioState::PLAYING;
+                }
+            });
+        }
+    }
+    else
+    {
+        AXLOGD("AudioEngineImpl::_play2d, cache was destroyed or not ready!");
```
halx99:
```suggestion
        AXLOGD("AudioEngineImpl::_play3d, cache was destroyed or not ready!");
```

halx99 (dismissed):

halx99 (approved):

Issue #2726, opened 2025-09-12
Audio panning and reverb


rudiHammad:
**Describe the solution you'd like**
Panning: allowing smooth left-right stereo positioning (currently in progress here by @rh101 https://github.com/rh101/axmol/tree/audio-enhancements. Note: there is a limitation in OpenAL that prevents any of the channels from being fully muted.”)
Reverb effects: Useful for simulating environmental acoustics, such as being inside a cave or a car entering a tunnel, to create more immersive audio experiences.

rh101:
Panning is done, but reverb and other effects will need more work. They're supported by OpenAL, but it would be good to have a straightforward API in Axmol to be able to add any effects and filters, rather than just enable reverb and nothing else.

This is the architecture of the effects and filters in OpenAL (from page 16 of the [OpenAL Effects Extensions Guide](https://openal-soft.org/misc-downloads/Effects%20Extension%20Guide.pdf)):

<img width="651" height="619" alt="Image" src="https://github.com/user-attachments/assets/68b72b68-9548-4912-a5c6-aac433c9508d" />

So, it would be good to have all that flexibility in the API so Axmol devs can use these features.

rh101:
Panning and reverb support are merged in, so closing this issue.

PR #2772, opened 2025-09-18
Initial implementation adding support for audio effects and filters using OpenAL

rh101:
## Describe your changes

Reverb effect is implemented, with preset reverb settings available that developers can use (in `AudioEffects.h`)

A simple test that loops through a few of the reverb effects is in the audio test section of cpp-tests.

@halx99 I wasn't sure what to name the files, so if you need anything changed, then let me know and I'll sort it out.

#2726 

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 


rh101:
@halx99 For Apple and Wasm, does the `alext.h` not exist on those platforms?  I can see this in the `alconfig.h`:

```
#if !defined(__APPLE__) && !defined(__EMSCRIPTEN__)
#    if !defined(AX_USE_ALSOFT)
#        define AX_USE_ALSOFT 1
#    endif
#endif

#if defined(__EMSCRIPTEN__)
    #import <AL/al.h>
    #import <AL/alc.h>
    #define MAX_AUDIOINSTANCES 128
#else
    #if !AX_USE_ALSOFT
    #    import <OpenAL/al.h>
    #    import <OpenAL/alc.h>
    #    define MAX_AUDIOINSTANCES 128
    #else
    #    define AL_ALEXT_PROTOTYPES 1
    #    include "AL/al.h"
    #    include "AL/alc.h"
    #    include "AL/alext.h"
    #    define MAX_AUDIOINSTANCES 128
    #endif
#endif
```

rh101:
Is `ALC_EXT_EFX` is not defined, and not equal to 1, then the effects code will no longer be compiled, and won't be supported for that build.  Other than that, I'm not sure how else to deal with this at the moment.

halx99:
I think, the class  `AudioEffectsExtensionOpenAL` should be rename to `AudioEffectsExtension` due to the audio engine only based on openal

halx99:
> ALC_EXT_EFX

wasm alext.h exist, but no `ALC_EXT_EFX` in it: https://github.com/emscripten-core/emscripten/blob/main/system/include/AL/alext.h

rh101:
> wasm alext.h exist, but no `ALC_EXT_EFX` in it: https://github.com/emscripten-core/emscripten/blob/main/system/include/AL/alext.h

That is unfortunate, so it means audio effects will not be supported on both WASM and iOS/MacOS.  All other target platforms seem to work fine.

halx99:
Yes, you can just check whether `AX_USE_ALSOFT` defined in cmake and c++ files.

rh101:
> Yes, you can just check whether `AX_USE_ALSOFT` defined in cmake and c++ files.

Will do.

core/audio/AudioEffects.h
halx99, line 1: can we include 
https://github.com/axmolengine/axmol/tree/dev/3rdparty/openal/include/AL/efx-presets.h directly since the this PR feature always depends on openal-soft ? refer: https://github.com/axmolengine/axmol/blob/dev/axmol/audio/oal_port.h

halx99:
> That is unfortunate, so it means audio effects will not be supported on both WASM and iOS/MacOS. All other target platforms seem to work fine.

On dev(v3), axmol can forcing `AX_USE_ALSOFT=ON` for all platforms include wasm (PR #2775) , macos, ios. in the future maybe v3, we can make openal-soft as default OpenAL vender for all platforms which is supported by axmol.

core/audio/AudioEffects.h
rh101: If OpenAL is the only audio backend being supported, then sure, we can remove the presets defined in this file and just use the ones defined in efx-presets.h.

It may be better to keep `AudioEffects.h` and just include `efx-presets.h` in it, since we can always add more custom presets ourselves too for different effects into this file. What do you think?

It would just be like this:
```
#pragma once

#include "alconfig.h"

#if AX_USE_ALSOFT
#    include "3rdparty/openal/include/AL/efx-presets.h"

namespace ax
{
typedef EFXEAXREVERBPROPERTIES ReverbProperties;
}  // namespace ax

#endif
```

rh101: Actually, we may have an issue with this.  Since `fx-presets.h` is not available on all platforms, the preset defines will not exist for certain target builds, so this will cause issues with game code that is referencing those defines.  It would mean that the developers would have to add conditional code in their apps to check for `AX_USE_ALSOFT` before attempting to use the defines etc..

Is this acceptable, or should the engine code always have the preset defines, so that in the app there is no need to add that conditional code?

rh101:
> On dev(v3), axmol can forcing `AX_USE_ALSOFT=ON` for all platforms include wasm (PR #2775) , macos, ios. in the future maybe v3, we can make openal-soft as default OpenAL vender for all platforms which is supported by axmol.

That would be good, keeping a consistent interface and feature set for developers on all platforms.

core/audio/AudioEffects.h
halx99: > It may be better to keep `AudioEffects.h` and just include `efx-presets.h` in it, since we can always add more custom presets ourselves too for different effects into this file. What do you think?

I agree totally.

halx99 (dismissed):

core/audio/alconstants.h
halx99, line 1: Does these consts from https://github.com/axmolengine/axmol/blob/dev/3rdparty/openal/include/AL/efx.h ?

rh101: The constants are from `alext.h` and `efx.h`.  Since certain target platforms may or may not have these defined, we would need to define them ourselves. This seems to be the way that other frameworks integrating OpenAL deal with this issue.  

An example of this problem is:
https://github.com/axmolengine/axmol/blob/54fb1f8b4807e18c22a843210d30b3d2d531ec16/core/audio/AudioEngineImpl.cpp#L1023

At the time `0x1030` had to be used instead of the `AL_STEREO_ANGLES`, since `AL_STEREO_ANGLES` may not exist for that build target.

Do you want this to be handled differently?  We may be able to just use the existing OpenAL defines, and add checks for `AX_USE_ALSOFT` in all locations where the constants are used, but it would just mean scattering more preprocessor checks in the code.

core/audio/AudioEffects.h
halx99: > Actually, we may have an issue with this. Since `fx-presets.h` is not available on all platforms, the preset defines will not exist for certain target builds, so this will cause issues with game code that is referencing those defines. It would mean that the developers would have to add conditional code in their apps to check for `AX_USE_ALSOFT` before attempting to use the defines etc..
> 
> Is this acceptable, or should the engine code always have the preset defines, so that in the app there is no need to add that conditional code?

suggestion:

```cpp
// axmol/audio/oal_port.h , original name: axmol/audio/alconfig.h
#pragma once

#if !defined(__APPLE__) && !defined(__EMSCRIPTEN__)
#    if !defined(AX_USE_ALSOFT)
#        define AX_USE_ALSOFT 1
#    endif
#else
#    if !defined(AX_USE_ALSOFT)
#        define AX_USE_ALSOFT 0
#    endif
#endif

// Use the OpenAL Soft headers bundled with Axmol instead of system-installed AL/xxx.h.
// This ensures consistent behavior across platforms and prevents Emscripten from
// accidentally including system headers.
// Define AL_STOP_SOURCES_ON_DISCONNECT_SOFT here to avoid pulling in the internal
// header "alc/inprogext.h".
#if AX_USE_ALSOFT
#    define AL_ALEXT_PROTOTYPES 1
#    include "3rdparty/openal/include/AL/al.h"
#    include "3rdparty/openal/include/AL/alc.h"
#    include "3rdparty/openal/include/AL/alext.h"
#    include "3rdparty/openal/include/AL/efx.h"
#    include "3rdparty/openal/include/AL/efx-presets.h"

#    ifndef AL_STOP_SOURCES_ON_DISCONNECT_SOFT
#        define AL_STOP_SOURCES_ON_DISCONNECT_SOFT 0x19AB
#    endif
#else
#    if defined(__APPLE__)
#        import <OpenAL/al.h>
#        import <OpenAL/alc.h>
#    elif defined(__EMSCRIPTEN__)
#        include "AL/al.h"
#        include "AL/alc.h"
#        include "AL/alext.h"
#    else
#        error "OpenAL Soft is required on non-(Apple/Emscripten) platforms"
#    endif

#endif

#ifndef MAX_AUDIOINSTANCES
#    define MAX_AUDIOINSTANCES 128
#endif

#if !AX_USE_ALSOFT

// define dummy efx macros and consts used in axmol audio effect implementation
// the macros or consts 's value is dummy, could be any, for example, for non-enum check
// in axmol audio engine, can be 0xFFFFFFFF for placeholder, if macros  or consts used by switch, can be 0xFFFFFFFF, 0xFFFFFFFE, 0xFFFFFFFD, 0xFFFFFFFC ...
// ...

#endif

```

rh101: Just to clarify, do you mean adding the `oal_port.h` file to the `release/2.x` branch, or just copying the contents of `oal_port.h` to use it in the existing `axmol/audio/alconfig.h`, and add the dummy defines in there like this:

```
#ifndef AL_STEREO_ANGLES
#define AL_STEREO_ANGLES
#endif
```

halx99: yes, just rename alconfig.h to oal_port.h in release/2.x, I will handle merge conflicits when merge release/2.x to `dev`

rh101: All good, I'll sort that out.

halx99 (approved):

halx99:
lgtm, will merge later

halx99:
I got error when merge to dev(v3), not sure does release/2.x works, will test later

```txt
[ALSOFT] (WW) Error generated on context 0x2a2bfc32120, code 0xa003, "EAX Reverb hfreference out of range"
Exception thrown at 0x00007FFFF27A804A in cpp-tests.exe: Microsoft C++ exception: al::base_exception at memory location 0x000000FC68DFE990.
E/[2025-09-20 15:53:47.975][PID:74d4][TID:ba88]OpenAL error 0xA003 in D:\dev\simdsoft\axmol\axmol\audio\AudioEffectsExtension.cpp ax::AudioEffectsExtension::setEffectParamFloat 90

[ALSOFT] (WW) Error generated on context 0x2a2bfc32120, code 0xa003, "EAX Reverb lfreference out of range"
Exception thrown at 0x00007FFFF27A804A in cpp-tests.exe: Microsoft C++ exception: al::base_exception at memory location 0x000000FC68DFE990.
E/[2025-09-20 15:53:47.977][PID:74d4][TID:ba88]OpenAL error 0xA003 in D:\dev\simdsoft\axmol\axmol\audio\AudioEffectsExtension.cpp ax::AudioEffectsExtension::setEffectParamFloat 90

[ALSOFT] (WW) Error generated on context 0x2a2bfc32120, code 0xa003, "EAX Reverb hfreference out of range"
Exception thrown at 0x00007FFFF27A804A in cpp-tests.exe: Microsoft C++ exception: al::base_exception at memory location 0x000000FC68DFE990.
E/[2025-09-20 15:53:50.974][PID:74d4][TID:ba88]OpenAL error 0xA003 in D:\dev\simdsoft\axmol\axmol\audio\AudioEffectsExtension.cpp ax::AudioEffectsExtension::setEffectParamFloat 90

[ALSOFT] (WW) Error generated on context 0x2a2bfc32120, code 0xa003, "EAX Reverb lfreference out of range"
Exception thrown at 0x00007FFFF27A804A in cpp-tests.exe: Microsoft C++ exception: al::base_exception at memory location 0x000000FC68DFE990.
E/[2025-09-20 15:53:50.976][PID:74d4][TID:ba88]OpenAL error 0xA003 in D:\dev\simdsoft\axmol\axmol\audio\AudioEffectsExtension.cpp ax::AudioEffectsExtension::setEffectParamFloat 90

```

halx99:
> I got error when merge to dev(v3), not sure does release/2.x works, will test later
> 
> ```
> [ALSOFT] (WW) Error generated on context 0x2a2bfc32120, code 0xa003, "EAX Reverb hfreference out of range"
> Exception thrown at 0x00007FFFF27A804A in cpp-tests.exe: Microsoft C++ exception: al::base_exception at memory location 0x000000FC68DFE990.
> E/[2025-09-20 15:53:47.975][PID:74d4][TID:ba88]OpenAL error 0xA003 in D:\dev\simdsoft\axmol\axmol\audio\AudioEffectsExtension.cpp ax::AudioEffectsExtension::setEffectParamFloat 90
> 
> [ALSOFT] (WW) Error generated on context 0x2a2bfc32120, code 0xa003, "EAX Reverb lfreference out of range"
> Exception thrown at 0x00007FFFF27A804A in cpp-tests.exe: Microsoft C++ exception: al::base_exception at memory location 0x000000FC68DFE990.
> E/[2025-09-20 15:53:47.977][PID:74d4][TID:ba88]OpenAL error 0xA003 in D:\dev\simdsoft\axmol\axmol\audio\AudioEffectsExtension.cpp ax::AudioEffectsExtension::setEffectParamFloat 90
> 
> [ALSOFT] (WW) Error generated on context 0x2a2bfc32120, code 0xa003, "EAX Reverb hfreference out of range"
> Exception thrown at 0x00007FFFF27A804A in cpp-tests.exe: Microsoft C++ exception: al::base_exception at memory location 0x000000FC68DFE990.
> E/[2025-09-20 15:53:50.974][PID:74d4][TID:ba88]OpenAL error 0xA003 in D:\dev\simdsoft\axmol\axmol\audio\AudioEffectsExtension.cpp ax::AudioEffectsExtension::setEffectParamFloat 90
> 
> [ALSOFT] (WW) Error generated on context 0x2a2bfc32120, code 0xa003, "EAX Reverb lfreference out of range"
> Exception thrown at 0x00007FFFF27A804A in cpp-tests.exe: Microsoft C++ exception: al::base_exception at memory location 0x000000FC68DFE990.
> E/[2025-09-20 15:53:50.976][PID:74d4][TID:ba88]OpenAL error 0xA003 in D:\dev\simdsoft\axmol\axmol\audio\AudioEffectsExtension.cpp ax::AudioEffectsExtension::setEffectParamFloat 90
> ```

`release/2.x` same error

rh101:
I'll take a look

PR #2774, opened 2025-09-18
New method: Color().random()

aismann:
## Describe your changes

new method: Color::random(alpha)

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/axmol/math/Color.h
+++ b/axmol/math/Color.h
@@ -177,6 +177,18 @@ struct AX_DLL Color : public Vec4Adapter<Color>
         else
             return Color{r, g, b, static_cast<float>(alpha) / 255.0f};
     }
+	
+    template <typename T>
+    Color random(T alpha)
+    {
+        static_assert(std::is_floating_point_v<T> || std::is_unsigned_v<T>,
+                      "withAlpha: alpha must be float (0~1) or unsigned integer (0~255)");
+
+        if constexpr (std::is_floating_point_v<T>)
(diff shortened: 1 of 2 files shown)


axmol/math/Color.h
```diff
@@ -178,6 +178,18 @@ struct AX_DLL Color : public Vec4Adapter<Color>
             return Color{r, g, b, static_cast<float>(alpha) / 255.0f};
     }
 
+    template <typename T>
```
halx99: shoud be static method like:
 ```cpp
    static Color random()
    {
            return Color(AXRANDOM_0_1(), AXRANDOM_0_1(), AXRANDOM_0_1(), AXRANDOM_0_1());
    }
```

then usage:
```cpp
auto color1 = Color::random();        // full random color
auto color2 = color1.withAlpha(0.5f);  // only rgb is random
```

halx99 (changes requested):

axmol/math/Color.h
aismann: Is it usefull to have the `alpha` also random?
See also the tester, alpha has to be set.
Alpha should be under control of the developer, for random alpha use: `Color::random(AXRANDOM_0_1())`

There is a  `Color withAlpha(T alpha) const` returning rgb and defined alpha

halx99:
`Color::random().withAlpha(1.0f)` also good, it's semantic more clearly

aismann:
ready for merge

halx99 (approved):

PR #2779, opened 2025-09-20
Fix reverb properties values

rh101:
### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/core/audio/AudioEffects.h
+++ b/core/audio/AudioEffects.h
@@ -25,13 +25,43 @@
 
 #include "audio/oal_port.h"
 
-#if AX_USE_ALSOFT
-
 namespace ax
 {
 
+#if AX_USE_ALSOFT
+
 typedef EFXEAXREVERBPROPERTIES ReverbProperties;
+
+#else
+
+struct ReverbProperties
+{
+    float flDensity;
+    float flDiffusion;
+    float flGain;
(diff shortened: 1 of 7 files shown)


halx99:
does possible remove preprocessor check:

```cpp
#if AX_USE_ALSOFT
    /**
     * Sets and enables reverb for an audio track.
     *
     * @param audioId        An audioID returned by the play2d function.
     * @param reverbProperties The pointer to reverb effect settings. If this is nullptr, then reverb effect will be
     * disabled.
     */
    static void setReverbProperties(AUDIO_ID audioId, const ReverbProperties* reverbProperties);
#endif
```

or hidden for luabindings ?

halx99:
image

lua build will raise compile error when build for wasm without openal-soft

rh101:
There is an alternative solution. I'll push the changes soon.

rh101:
To avoid issues, we can always have `setReverbProperties` declared and defined, and just create a dummy `ReverbProperties` for  when openal-soft is not used.

halx99:
> To avoid issues, we can always have `setReverbProperties` declared and defined, and just create a dummy `ReverbProperties` for when openal-soft is not used.

ok

halx99 (approved):

PR #2799, opened 2025-09-24
Make Director::getRunningScene() return the expected result for any type of scene switch when called from Node::onEnter()

### LINKED ISSUE #2798: Discrepancy between instance scene switches and transitions when calling Director::getRunningScene()
- axmol version:
- devices test on:
- developing environments
   - NDK version: r23c
   - Xcode version: 14.2+
   - Visual Studio: 
      - VS version: 2022 (17.9+)
      - MSVC version: 19.39+
   - cmake version: 
Steps to Reproduce:

This issue was raised in the discord server by @AlexandreK38 

With normal scene switches, using `pushScene` or `replaceScene`, this is the sequence of calls and value of `getRunningScene()`:

```
1. oldScene->onExit(), getRunningScene() = oldScene
2. newScene->onEnter(), getRunningScene() = newScene
```

If using transitions, this is what happens:
```
1. oldScene->onExit(), getRunningScene() = oldScene
2. newScene->onEnter(), getRunningScene() = oldScene
```

So, there is a difference in the value returned by `getRunningScene()` when using transitions. It should be the new scene, since `onEnter()` may have code that impacts the new scene (such as adding child nodes etc.).

The suggestion is to change the implementation of `Director::setNextScene()`, which will ensure that calling `Director::getRunnningScene()` in `onEnter()` will always return the new scene, regardless of how the scene switch occurs (push/replace or transition).

rh101:
## Describe your changes

This PR addresses the issue described in issue #2798, where the return value of `Director::getRunningScene()` is not consistent between instant scene switches and transitions if called in `onEnter()` of the new (incoming) scene.

The value of `getRunningScene()` when called from `onEnter()` will now always be the new scene. 

The value of `getRunningScene()` when called from `onExit()` will now also be the new scene.

#2798 

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/core/base/Director.cpp
+++ b/core/base/Director.cpp
@@ -1165,41 +1165,42 @@ void Director::setNextScene()
 {
     _eventDispatcher->dispatchEvent(_beforeSetNextScene);
 
-    bool runningIsTransition = dynamic_cast<TransitionScene*>(_runningScene) != nullptr;
-    bool newIsTransition     = dynamic_cast<TransitionScene*>(_nextScene) != nullptr;
+    auto outgoingScene = _runningScene;
+    _runningScene      = _nextScene;
+    _nextScene         = nullptr;
 
-    // If it is not a transition, call onExit/cleanup
-    if (!newIsTransition)
+    bool outgoingSceneIsTransition = dynamic_cast<TransitionScene*>(outgoingScene) != nullptr;
+
+    if (outgoingScene)
     {
-        if (_runningScene)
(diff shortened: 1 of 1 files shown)


halx99:
Wrong target branch I guess, so I switch to `release/2.x`

rh101:
> Wrong target branch I guess, so I switch to `release/2.x`

Sorry, yes, it should have been `release/2.x`. Thank you for fixing it

halx99 (approved):

PR #2800, opened 2025-09-24
Fix for crash when director is reset during scene transition

rh101:
## Describe your changes

If the application is closed during certain scene transitions, such as `TransitionFade`, the application will crash in `Node::~Node()` since the incoming scene is still active (`Node::_running == true`). This is due to the `Node::onExit()` not being called on the incoming scene in the transition.

This PR address that issue, by making it the responsibility of the transition to ensure that the incoming scene is correctly deactivated (`Node::_running == false`) if it detects that the `Director` has been reset, meaning that the app is shutting down.

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/core/2d/Transition.cpp
+++ b/core/2d/Transition.cpp
@@ -211,6 +211,14 @@ void TransitionScene::onExit()
     // only the onEnterTransitionDidFinish
     _inScene->onEnterTransitionDidFinish();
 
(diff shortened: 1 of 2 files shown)


rh101:
Incorrect base branch. Will create another PR

PR #2831, opened 2025-10-22
fix: [ui::Text] Overflow::SHRINK does not restore the original font size when text is shortened

### LINKED ISSUE #2830: [ui::Text] Overflow::SHRINK does not restore the original font size when text is shortened
### Detailed Description

A ui::Text using Overflow::SHRINK fails to scale its font size back to the originally specified value when the text string is changed from a long one (which triggered scaling down) to a short one (which should fit at the original size).

### Code Sample to Reproduce

```

    auto size = Director::getInstance()->getWinSize();

    {
        // Create a Label with SHRINK overflow and fixed dimensions
        TTFConfig ttfConfig("fonts/arial.ttf", 30);
        auto label = Label::createWithTTF(ttfConfig, "");
        label->setOverflow(Label::Overflow::SHRINK);
        label->setPosition(Vec2(size.width * 0.5f, size.height * 0.6f));
        label->setDimensions(100, 50);

        // Step 1: Set a long string (triggers correct scaling down)
        label->setString(
            "This is an extremely long string that will definitely exceed the width of the label container.");

        label->runAction(Sequence::createWithTwoActions(DelayTime::create(1.f), CallFunc::create([=] {
            // Step 2: Set a short string (triggers correct scaling down)
            label->setString("Short");  // The font size returning to 30.
        })));

        addChild(label);
    }

    {
        auto text = ui::Text::create();
        text->setFontSize(30);
        text->setFontName("fonts/arial.ttf");
        text->setTextAreaSize(Size(100, 50));
        text->setPosition(Vec2(size.width * 0.5f, size.height * 0.4f));
        dynamic_cast<Label*>(text->getVirtualRenderer())->setOverflow(Label::Overflow::SHRINK);

        // Step 1: Set a long string (triggers correct scaling down)
        text->setString(
            "This is an extremely long string that will definitely exceed the width of the label container.");

        text->runAction(Sequence::createWithTwoActions(DelayTime::create(1.f), CallFunc::create([=] {
            // Step 2: Set a short string (BUG: font size does not scale back up)
            text->setString("Short");  // The font size remains small instead of returning to 30.
        })));

        addChild(text);
    }


```

tkzcfc:
fix: https://github.com/axmolengine/axmol/issues/2830

### DIFF
--- a/core/2d/Label.cpp
+++ b/core/2d/Label.cpp
@@ -2574,6 +2574,9 @@ const Vec2& Label::getContentSize() const
 {
     if (_systemFontDirty || _contentDirty)
     {
+        if (_overflow == Overflow::SHRINK && this->getRenderingFontSize() < _originalFontSize)
+            const_cast<Label*>(this)->rescaleWithOriginalFontSize();
+
         const_cast<Label*>(this)->updateContent();
     }
     return _contentSize;


aismann:
@tkzcfc 
Is this not better: replace `<` with `!=`?
`getRenderingFontSize() != _originalFontSize)`

tkzcfc:
I referenced the existing code logic from our codebase ([line 2111 in Label.cpp](https://github.com/axmolengine/axmol/blob/dev/axmol/2d/Label.cpp#L2111)), which specifically checks for the shrink overflow scenario when the current rendering font size is smaller than the original.

The condition this->getRenderingFontSize() < _originalFontSize is intentional because:

- If the rendering size is already equal to the original size, no rescaling is needed

- If the rendering size is larger than the original, we also don't want to trigger rescaling in this shrink overflow scenario

- We only want to rescale when the current size has become smaller than the original

halx99 (approved):

PR #2848, opened 2025-11-03
fix install plainlua dll

sidyhe:
fix install error
when axmol as sub-project and ``set(CMAKE_DEBUG_POSTFIX _d)`` in topest CMakeLists.txt


paulocoutinhox:
I think that there is something wrong in your PR, because you fixed it to debug configurations in CMake files.

sidyhe:
when CMAKE_DEBUG_POSTFIX was set, the dll name is ``plainlua_d.dll`` not ``plainlua.dll``
if build type not debug, the expr is empty

PR #2868, opened 2025-11-13
Switch default audio backend on iOS/macOS from OpenAL.framework to openal-soft

### LINKED ISSUE #2866: Sounds crackle and intermittently restart playing
Developing Environment:
Xcode 26.0.1
axmol 2.8.1

Test Device:
iPhone SE
iOS 18.6.2

Hello Axmol community. Lately I've been working on my app's sound effects (I always do sound last). I generally work on my Mac but after upgrading Xcode and iOS, I tested the app on my iPhone and found that many of the sounds begin with a harsh crackle. Also, intermittently, a sound will restart playing. None of this occurs on the Mac, only on the phone. 

The crackle occurs at the beginning of the sound. Some sounds are worse than others. If the sound is set to repeat, the crackle only occurs at the first iteration. I emailed one of the crackly sounds to myself, downloaded it to my phone, and played it from the Files app. It played without crackling so this looks like an Axmol issue. After various tests, I came to the conclusion that our sound engine starts playing a sound at about the 25 millisecond mark. If a sound's waveform is near zero at that point, then the sound plays somewhat normally but if the waveform is near a crest at the 25 ms mark, then the speaker membrane has to go from 0 to 60 mph in 0 seconds and smashes against the casing, causing a crackle. As a workaround, I've added 25 ms of silence to the beginning of my sound files after which they play perfectly.

The intermittent restarting of sound effects, I can't explain.

Steps to Reproduce:

1. Create a new Axmol app.
2. Add three sound effects to the resources, say "Bing", "Bong", and "Boing" (I use .mp3).
3. Have the app play the sounds in repeating sequence, one sound every 0.75 seconds. Ex: "Bing... Bong... Boing... Bing... Bong... Boing... etc."
4. Let the app run in the background as you do something productive. Before too long, you will hear a sound restart. Sometimes it restarts early. Ex: "B-Bing". Sometimes it restarts later. Ex: "Boin-Boing".

My theory is that the latest iOS breaks something in Axmol's sound engine. I hope someone can confirm and fix these issues. Thanks.

### LINKED ISSUE #2867: Remove cmake AX_USE_ALSOFT optional for apple platforms

iOS play sounds have issue #2866 with `OpenAL.framework` which is marked deprecated by Apple few years.

**Finally**
Since Axmol 2.9.2, on iOS and macOS platforms, the default audio backend has been switched from OpenAL.framework (deprecated by Apple years ago) to openal-soft. The previous CMake option -DAX_USE_ALSOFT has been removed, as it is no longer necessary.

**Commit message**
Switch default backend on iOS/macOS from OpenAL.framework to openal-soft

- OpenAL.framework has long been deprecated by Apple
- openal-soft is now the enforced default backend for native platforms
- Removed obsolete CMake option `-DAX_USE_ALSOFT`

**Release Notes**

- On iOS and macOS, the default audio backend has been changed from `OpenAL.framework` (deprecated by Apple) to `openal-soft`.
- The CMake option `-DAX_USE_ALSOFT` has been removed, as it is no longer required.

halx99:
- OpenAL.framework has long been deprecated by Apple
- openal-soft is now the enforced default backend for native platforms
- Removed obsolete CMake option `-DAX_USE_ALSOFT`

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -253,7 +253,7 @@ endif()
 
 if(AX_ENABLE_AUDIO)
   # The openal-soft(LGPL 2.1)
-  if(AX_USE_ALSOFT AND NOT EMSCRIPTEN)
+  if(NOT EMSCRIPTEN)
     set(alsoft_opts "ALSOFT_DLOPEN OFF" "ALSOFT_UTILS OFF" "ALSOFT_EXAMPLES OFF" "ALSOFT_INSTALL OFF")
 
     if(AX_USE_ALSOFT_STATIC)
@@ -289,7 +289,6 @@ if(AX_ENABLE_AUDIO)
     ax_add_3rd(openal EXCLUDE_FROM_ALL TARGETS alcommon;OpenAL OPTIONS ${alsoft_opts})
 
     target_include_directories(3rdparty INTERFACE openal)
-    target_compile_definitions(3rdparty INTERFACE AX_USE_ALSOFT=1)
     set_target_properties(OpenAL alcommon PROPERTIES CXX_STANDARD ${_AX_CXX_STD})
 
     if(AX_USE_ALSOFT_STATIC)
--- a/CMakeOptions.md
+++ b/CMakeOptions.md
@@ -16,9 +16,6 @@
   - AX_ENABLE_CONSOLE: whether to enable debug tool console support, default: `TRUE`
   - AX_ENABLE_OPUS: whether to enable audio engine play .opus files support, default: `TRUE`
 - AX_USE_XXX:
-  - AX_USE_ALSOFT: whether use openal-soft for all platforms
-    - Apple platform: Use openal-soft instead system deprecated: `OpenAL.framework`
-    - Other platforms: Always use openal-soft even this option not enabled
   - AX_USE_LUAJIT: whether use luajit, default: `FALSE`, use plainlua
 - AX_ENABLE_EXT_XXX for extensions
   - AX_ENABLE_EXT_GUI: the traditional GUI extension, default: `TRUE`
--- a/cmake/Modules/AXBuildHelpers.cmake
+++ b/cmake/Modules/AXBuildHelpers.cmake
@@ -519,7 +519,7 @@ function(ax_setup_app_config app_name)
 
   target_link_libraries(${app_name} ${_AX_EXTENSION_LIBS})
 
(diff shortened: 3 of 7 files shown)


g-salvie:
After abandoning the standard OpenAL framework, an issue appeared on iOS in this and all subsequent versions of Axmol when uploading the app to the App Store. Xcode reports that OpenAL is not code-signed, and if you choose not to embed it, the app builds successfully but crashes at runtime.

rh101:
> After abandoning the standard OpenAL framework, an issue appeared on iOS in this and all subsequent versions of Axmol when uploading the app to the App Store. Xcode reports that OpenAL is not code-signed, and if you choose not to embed it, the app builds successfully but crashes at runtime.

We have an iOS app that uses OpenAL-soft, built with Axmol 2.10, and it was approved with no issues.  The OpenAL-soft library must be embedded into the output, otherwise, as you noted, it will crash at runtime if it is missing that library.  

Please show a screenshot of the Xcode message regarding the code signing of the OpenAL-soft library.

g-salvie:
> > After abandoning the standard OpenAL framework, an issue appeared on iOS in this and all subsequent versions of Axmol when uploading the app to the App Store. Xcode reports that OpenAL is not code-signed, and if you choose not to embed it, the app builds successfully but crashes at runtime.
> 
> We have an iOS app that uses OpenAL-soft, built with Axmol 2.10, and it was approved with no issues. The OpenAL-soft library must be embedded into the output, otherwise, as you noted, it will crash at runtime if it is missing that library.
> 
> Please show a screenshot of the Xcode message regarding the code signing of the OpenAL-soft library.
Screenshot 2025-12-24 at 12 40 46

rh101:
If you click on the OpenAL entry in the Xcode project, then select "Info", can you copy the contents of that screen? This is what I see in the project that has no issues:
image

As you can see there, Bundle Version exists, so that error you are seeing is really strange.

Also, go to Window->Organiser, then select the archive created in the Archive section, right click it, then click "Show in Finder".  Right click the xcarchive, and select "Show Package Contents", then select Products -> Applications->{your app name}.app.  Right click the app and select "Show Package Contents", and browse to "Frameworks->soft_oal.framework", where you will see the `Info.plist` file.  Open that file, and you should see something like this (which includes the BundleVersion):
image

Does your app show the same as the screenshots above?

PR #2873, opened 2025-11-15
Fix invalid AL_STOPPED transition caused by background route changes on iOS

lucycycy:
In some iOS environments, when the app is in the background, inserting or removing headphones can trigger AVAudioSessionRouteChangeNotification. After returning to the foreground, AudioEngine::resumeAll restores the source state from AL_PAUSED back to AL_PLAYING. However, at the moment resumeAudioDevice is invoked during UIApplicationDidBecomeActiveNotification, the source state may unexpectedly transition to AL_STOPPED. As a result, rotateBufferThread attempts to restart playback and triggers "Error restarting playback!".

This fix makes rotateBufferThread skip all OpenAL buffer operations while the current context is NULL, avoiding invalid context access during this period and preventing incorrect state transitions that occur when returning from the background.

### DIFF
--- a/core/audio/AudioPlayer.cpp
+++ b/core/audio/AudioPlayer.cpp
@@ -457,6 +457,16 @@ void AudioPlayer::rotateBufferThread(int offsetFrame)
 
         while (!_isDestroyed)
         {
+
+#if AX_TARGET_PLATFORM == AX_PLATFORM_IOS && !AX_USE_ALSOFT
+            if (alcGetCurrentContext() == nullptr)
+            {
+                std::unique_lock<std::mutex> lk(_sleepMutex);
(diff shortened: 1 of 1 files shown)


core/audio/AudioPlayer.cpp
```diff
@@ -457,6 +457,16 @@ void AudioPlayer::rotateBufferThread(int offsetFrame)
 
         while (!_isDestroyed)
         {
+
+#if AX_TARGET_PLATFORM == AX_PLATFORM_IOS && !AX_USE_ALSOFT
```
halx99, line 461: Note, since axmol-2.10.0 (not relased), the AX_USE_ALOSFT always 1 on native platforms

PR #2883, opened 2025-11-21
Fix for xcarchive generation for iOS and MacOS when using OpenAL-Soft

### LINKED ISSUE #2876: [iOS] Xcode archive and distribution option broken in latest releases/2.x branch
- axmol version: releases/2.x branch, https://github.com/axmolengine/axmol/tree/b29f765003e70fba56917cc5fd8035cc680e6272
- devices test on: 
- developing environments
   - NDK version: r23c
   - Xcode version: 16.2
   - Visual Studio: 
      - VS version: 2022 (17.9+)
      - MSVC version: 19.39+
   - cmake version: 4.2
Steps to Reproduce:

1. Build app for distribution in Xcode using the Archive command
2. The build is successful, but when the organizer opens, it is not listed under the correct app target name, but instead shows "{Target} (other item)", and the entry is missing the version code. For example:

Image

Xcode prevents the build from being distributed to the App Store because it is not valid.

Building with Axmol v2.9.1 fixes the issue.

Could this possibly have something to do with the changes to cmake for the use of OpenAL-soft on iOS?

rh101:
## Describe your changes

This addresses the issue described in #2876

An extra change was required in `3rdparty/CMakeLists.txt` to avoid a strange linker error when this attribute is set for the OpenAL-soft:

`XCODE_ATTRIBUTE_CONFIGURATION_BUILD_DIR "$(inherited)"`

It doesn't happen on iOS, but it shows a linker error on MacOS.  The only way to fix it was to not force the setting of the output directories for shared libraries in `3rdparty/CMakeLists.txt`.  The linker error disappears and everything builds correctly with this change.

#2876 

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 


halx99 (dismissed):

halx99:
@rh101 prebuilt workflow check fail: `LINK : fatal error LNK1104: cannot open file 'OpenAL32.lib' [D:\a\axmol\axmol\HelloCpp\build\HelloCpp.vcxproj]`, could you take a look?

rh101:
> @rh101 prebuilt workflow check fail: `LINK : fatal error LNK1104: cannot open file 'OpenAL32.lib' [D:\a\axmol\axmol\HelloCpp\build\HelloCpp.vcxproj]`, could you take a look?

I think it's fixed now, just took a while testing it out to make sure the MacOS build still worked without generating the linker error.

PR #2907, opened 2025-11-29
Add Node::getHashOfName().

j-jorge:
`ax::Node` maintains a hash of the node's name internally for faster lookup of the children. Exposing the hash allows to benefit from faster look-up from outside Axmol too.

### DIFF
--- a/core/2d/Node.cpp
+++ b/core/2d/Node.cpp
@@ -54,13 +54,13 @@ THE SOFTWARE.
 #    define RENDER_IN_SUBPIXEL(__ARGS__) (ceil(__ARGS__))
 #endif
 
-/*
- * 4.5x faster than std::hash in release mode
- */
-#define AX_HASH_NODE_NAME(name) (!name.empty() ? XXH3_64bits(name.data(), name.length()) : 0)
-
 namespace ax
 {
+AX_DLL uint64_t hashNodeName(std::string_view name)
+{
+  // 4.5x faster than std::hash in release mode.
+  return !name.empty() ? XXH3_64bits(name.data(), name.length()) : 0;
+}
 
 // FIXME:: Yes, nodes might have a sort problem once every 30 days if the game runs at 60 FPS and each frame sprites are
 // reordered.
@@ -717,6 +717,11 @@ void Node::setName(std::string_view name)
     _name = name;
 }
 
+uint64_t Node::getHashOfName() const
+{
+  return _hashOfName;
+}
+
 void Node::updateParentChildrenIndexer(int tag)
 {
     auto parentChildrenIndexer = getParentChildrenIndexer();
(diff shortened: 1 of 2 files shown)


core/2d/Node.cpp
```diff
@@ -54,13 +54,13 @@ THE SOFTWARE.
...
 #endif
 
-/*
- * 4.5x faster than std::hash in release mode
- */
-#define AX_HASH_NODE_NAME(name) (!name.empty() ? XXH3_64bits(name.data(), name.length()) : 0)
-
 namespace ax
 {
+uint64_t hash_node_name(std::string_view name)
```
halx99: maybe shoud static function?

j-jorge: I can't, because I need the function to be accessible from other compilation units. For context, [here is my client code](https://github.com/j-jorge/bim/blob/365730a11ec013454999eeb0a02e9859134e306f/modules/axmol/core/main/src/bim/axmol/find_child_by_path.cpp#L17)

```c++
 const std::uint64_t hash = ax::hash_node_name(name);

  for (ax::Node* n : node.getChildren())
    if ((n->getHashOfName() == hash) && (n->getName() == name))
      // …
```

This is a function to get a child node via some kind of path, like `/child/grandchild/something`. I did not use `Node::getChildByName()` because it does not handle the case where multiple children have the same name. I've seen `Node::enumerateChildren()` too but seeing its use of recursive functions, regexes, and `std::string`, I think it is a bit too expensive for what I need.

Hope this help :)

j-jorge:
I've renamed the function `hash_node_name` -> `hashNodeName` to match the conventions in place.

core/2d/Node.cpp
```diff
@@ -54,13 +54,13 @@ THE SOFTWARE.
...
 #endif
 
-/*
- * 4.5x faster than std::hash in release mode
- */
-#define AX_HASH_NODE_NAME(name) (!name.empty() ? XXH3_64bits(name.data(), name.length()) : 0)
-
 namespace ax
 {
+uint64_t hashNodeName(std::string_view name)
```
halx99:
```suggestion
AX_DLL uint64_t hashNodeName(std::string_view name)
```

core/2d/Node.h
```diff
@@ -86,6 +86,8 @@ class EventListener;
 
 typedef std::map<uint64_t, Node*> NodeIndexerMap_t;
 
+uint64_t hashNodeName(std::string_view);
```
halx99:
```suggestion
AX_DLL uint64_t hashNodeName(std::string_view);
```

halx99 (changes requested):

core/2d/Node.cpp
j-jorge: I believe the export on the declaration, in the header, should be enough. Isn't it?

halx99: I checked Utils.cpp and lua source code, both header and c/cpp has export declaration

j-jorge: Ok, I've added it in the cpp file too.

halx99 (approved):

halx99 (approved):

PR #2910, opened 2025-11-30
AX_CORE_PROFILE fix usage of "ifndef AX_CORE_PROFILE"

crazyhappygame:
## Describe your changes
compilatin fails after enablaing ifndef AX_CORE_PROFILE

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -111,7 +111,7 @@ cmake_dependent_option(AX_ENABLE_NAVMESH "Build NavMesh support" ON "AX_ENABLE_3
 
 option(AX_UPDATE_BUILD_VERSION "Update build version" ON)
 option(AX_DISABLE_GLES2 "Whether disable GLES2 support" OFF)
-option(AX_CORE_PROFILE "Whether strip deprecated features" ON)
+option(AX_CORE_PROFILE "Whether strip deprecated features" OFF)
 
 # default value for axmol extensions modules to Build
 # total supported extensions count: 13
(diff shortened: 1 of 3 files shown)


crazyhappygame:
it looks like flakyh unit test. https://github.com/axmolengine/axmol/actions/runs/19804252495/job/56735998776?pr=2910
closing and reopen to trigger actions

crazyhappygame:
https://github.com/axmolengine/axmol/actions/runs/19804252495/job/56735998776?pr=2910

```
/Users/runner/work/axmol/axmol/tests/unit-tests/Source/core/base/MapTests.cpp:194: ERROR: CHECK( different > mapForRandom.size() * 3 / 4 ) is NOT correct!
Can't convert string to UTF-32: ��
  values: CHECK( 15 >  15 )
```
@halx99 any idea why it fails?

halx99:
#2909 duplicated PR ?

halx99:
I don't think we should set `AX_CORE_PROFILE` to `ON`, due to it's only use when user want prepare upgrading to next major release of axmol

crazyhappygame:
> #2909 duplicated PR ?

different branch

crazyhappygame:
> I don't think we should set `AX_CORE_PROFILE` to `ON`, due to it's only use when user want prepare upgrading to next major release of axmol

the issie is that it does not work now ...

halx99 (approved):

PR #2917, opened 2025-12-03
Minor perf-related changes in FontFreeType

j-jorge:
The blending loops were processing the bitmap column-major, but the bitmap is actually stored row-major. I've switched the loops to have a more cache-friendly iteration.

### DIFF
--- a/core/2d/FontFreeType.cpp
+++ b/core/2d/FontFreeType.cpp
@@ -545,9 +545,9 @@ unsigned char* FontFreeType::getGlyphBitmapByIndex(unsigned int glyphIndex,
 
                 auto px = outlineMinX - blendImageMinX;
                 auto py = blendImageMaxY - outlineMaxY;
-                for (int x = 0; x < outlineWidth; ++x)
+                for (int y = 0; y < outlineHeight; ++y)
                 {
-                    for (int y = 0; y < outlineHeight; ++y)
+                    for (int x = 0; x < outlineWidth; ++x)
                     {
                         index                 = px + x + ((py + y) * blendWidth);
                         index2                = x + (y * outlineWidth);
@@ -557,9 +557,9 @@ unsigned char* FontFreeType::getGlyphBitmapByIndex(unsigned int glyphIndex,
 
                 px = glyphMinX - blendImageMinX;
                 py = blendImageMaxY - glyphMaxY;
-                for (int x = 0; x < outWidth; ++x)
+                for (int y = 0; y < outHeight; ++y)
                 {
-                    for (int y = 0; y < outHeight; ++y)
+                    for (int x = 0; x < outWidth; ++x)
                     {
                         index                     = px + x + ((y + py) * blendWidth);
                         index2                    = x + (y * outWidth);


halx99 (changes requested):
Please don't reformat full-file, refer: https://github.com/axmolengine/axmol?tab=readme-ov-file#-branches

j-jorge:
> Please don't reformat full-file, refer: https://github.com/axmolengine/axmol?tab=readme-ov-file#-branches

Done, with my apologies.

rh101:
Multi-line blocks should be surrounded by curly brackets, such as this code in the current PR:

    for (BufferPool& p : _availableBuffers)
        if (p.capacity == it->second)
        {
            p.buffers.emplace_back(buffer);
            break;
        }

should be:

    for (BufferPool& p : _availableBuffers)
    {
        if (p.capacity == it->second)
        {
            p.buffers.emplace_back(buffer);
            break;
        }
    }

Same goes for this, where the comment block is below the conditional statement:

    if ((new_pool_index == n) || (_availableBuffers[new_pool_index].capacity != size))
        // Either all pools have capacities smaller than the requested size, or else
        // We have pools with capacity larger than the requested size but none had
        // an available buffer. We create a new pool for the requested size so
        // we'll be ready to receive the buffer in releaseBuffer.
        _availableBuffers.emplace(_availableBuffers.begin() + new_pool_index)->capacity = size;

where it should be:

    if ((new_pool_index == n) || (_availableBuffers[new_pool_index].capacity != size))
    {
        // Either all pools have capacities smaller than the requested size, or else
        // We have pools with capacity larger than the requested size but none had
        // an available buffer. We create a new pool for the requested size so
        // we'll be ready to receive the buffer in releaseBuffer.
        _availableBuffers.emplace(_availableBuffers.begin() + new_pool_index)->capacity = size;
    }

halx99:
Thanks for the contribution! Before we consider merging, could you please share some benchmarks or profiling results that demonstrate the actual performance improvement of this change? It would help us evaluate whether the added complexity is justified.

j-jorge:
@halx99 Here are some measurements on two of my devices. I measure the time spent in `FontFreeType::getGlyphBitmap`, as reported by `std::chrono::steady_clock`. There are 507 calls on the Pixel and 443 on the Galaxy Tab. The comparison is done on the min/max/median of 10 runs.

|  | Pixel 3a | | | Galaxy Tab A | | |
| -- | -- | -- | -- | -- | -- | -- |
|  | Base | Loops | Loops+alloc | Base | Loops | Loops+alloc
| Min | 100,00 % | -1,26 % | -0,52 % | 100,00 % | -0,37 % | -0,11 %
| Max | 100,00 % | -1,14 % | 0,07 % | 100,00 % | -19,29 % | -19,44 %
| Median | 100,00 % | -0,66 % | 0,47 % | 100,00 % | -0,20 % | -0,07 %

The second commit, with the allocations, is pure loss (I should have done separate measurements from the start). I suggest to remove it and keep the first commit, which changes the loops.

[axmol-benchmark-font-free-type.ods](https://github.com/user-attachments/files/24019650/axmol-benchmark-font-free-type.ods)

halx99:
I have test by visual studio CPU usage tool, and the malloc/free not the hot path of FontAtlas::prepareLetterDefinitions:
image

j-jorge:
@halx99 I've removed the commit changing the allocations, keeping only the part about the blending loops.

halx99 (approved):

PR #2978, opened 2025-12-26
Fix sprite sheet creation logic

suhailmalik07:
## Describe your changes
Instead of creating new SpriteSheet on loading, Now it'll check if that SpriteSheet already exists, if it does use and fill it otherwise create new.

Let me know if needs more details or other issues in the PR.

Currently if SpriteSheet is partially deleted by android/ios partial memory cleanup. Reloading it again loads only missing sprites to a new SpriteSheet and marks it as `full` but it's not full. So If the cleanup happens again some sprites will be missed an not loaded again even after multiple tries.

```
Example.
Assume `p.plist` contains {`a.png`, `b.png`}

on first fresh load.
SpriteSheet{P}
  full: true
  _spriteFrames: {a.png, b.png}


first cleanup, let's assume a.png is removed
SpriteSheet{P}
  full: false
  _spriteFrames: {b.png}

If we load it again, current SpriteSheet will gets replaced, and it only loads missing png and marks it full.
SpriteSheet{P}
  full: true
  _spriteFrames: {a.png}


`b.png` will still be usable as SpriteFrameCache stores name to sprite.
but after second cleanup, let's assume `b.png` gets cleaned up. 
`p.plist` will not get loaded again as it's already full and `b.png` will not be found as as `SpriteSheet{P}` and cache doesn't contains it anymore.
```

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/axmol/2d/PlistSpriteSheetLoader.cpp
+++ b/axmol/2d/PlistSpriteSheetLoader.cpp
@@ -190,9 +190,16 @@ void PlistSpriteSheetLoader::addSpriteFramesWithDictionary(ValueMap& dictionary,
     if (dictionary["frames"].getType() != ax::Value::Type::MAP)
         return;
 
-    auto spriteSheet    = std::make_shared<SpriteSheet>();
-    spriteSheet->format = getFormat();
-    spriteSheet->path   = plist;
+    auto spriteSheet    = cache.getSpriteSheet(plist);
+
+    if (spriteSheet == nullptr)
+    {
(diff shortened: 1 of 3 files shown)


rh101:
Should this be targeting the `release/2.x` branch instead?

halx99 (approved):

halx99:
I merge this PR patch to `release/2.x` manually, Thanks for your PR @suhailmalik07

PR #2983, opened 2025-12-28
Check 'CMakeCache.txt' file exists...

aismann:
## Describe your changes
image

Not only a beginner issue ... 

Output after `axmol build -c`

```
CMake Error at CMakeLists.txt:38 (message):
  Please remove 'D:/_git/axmol/build' directory or at least
  'D:/_git/axmol/build/CMakeCache.txt' before you build.
```

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,13 @@ cmake_minimum_required(VERSION 3.22...4.2)
 
 project(axmol)
 
+# Check 'CMakeCache.txt' exists:
+# Deleting the cache file 'CMakeCache.txt often fixes weird CMake issues.
+set(CACHE_FILE "${CMAKE_BINARY_DIR}/CMakeCache.txt")
+if(EXISTS ${CACHE_FILE})
+ message(SEND_ERROR  "Please remove '${CMAKE_BINARY_DIR}' directory or at least '${CACHE_FILE}' before you build.")
+endif()
+
 if(XCODE)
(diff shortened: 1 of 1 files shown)


rh101:
What do you mean by a "new cmake"? I don't quite understand the purpose of this change, besides constantly displaying an error message when cmake is run more than once.

Does the `-f` flag not work correctly?

halx99:
This PR will disable cmake cache feature, that means.

rh101:
> This PR will disable cmake cache feature, that means.

I understand.  I just realized that it applies to the engine root `CMakeLists.txt`, and not the templates and normal projects, which seems fine.  A change like this would not be good in the templates etc..

halx99:
I prefer add FAQ instead this patch which is not cmake best practice.

rh101:
> I prefer add FAQ instead this patch which is not cmake best practice.

I agree with you; documenting it in the FAQ is better than forcing an error in cmake.

aismann:
> > I prefer add FAQ instead this patch which is not cmake best practice.
> 
> I agree with you; documenting it in the FAQ is better than forcing an error in cmake.

I agree also. Its more my own build process which I am write in this PR.
Close it. Thanks for your feedback. Its make clearer for me too

grif-on:
In the initial discord discussion we talked about adding a note about CMakeCache.txt into the FAQ or/and the extensions wiki page .

So yeah , having caching explicitly turned off completely is an overkill .
Especially with such error that unconditionally require manual actions to be taken .

PR #2984, opened 2025-12-29
Improve APIs which contains std::vector<XX>

halx99:
### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 


AlexandreK38:
I'm curious, why is the advantage to use std::span instead of a std::vector?

halx99:
> I'm curious, why is the advantage to use std::span instead of a std::vector?

For example:

image

By using `std::span`, you can avoid constructing a temporary vector, which may cause unnecessary heap allocations.

Issue #3040, opened 2026-01-19
[WASM] Wasm Downloader storage issue


AlexandreK38:
- axmol version: 2.7.0
- devices test on: Webgl2 Chrome MacOS
Steps to Reproduce:

Currently, the Wasm-Downloader.cpp is using `emscripten_fetch` with attributes 
`attr.attributes = EMSCRIPTEN_FETCH_LOAD_TO_MEMORY | EMSCRIPTEN_FETCH_PERSIST_FILE;`

By using `EMSCRIPTEN_FETCH_PERSIST_FILE`, emscripten will save the downloaded file in indexedDB database `emscripten_filesystem`, and also have the bytes in memory thanks for `EMSCRIPTEN_FETCH_LOAD_TO_MEMORY`.
The issue is that the file is saved twice, once at the location we gave (in axmol writable path) and the other because of the `EMSCRIPTEN_FETCH_PERSIST_FILE`  emscripten "cache".
I don't think we could move the file from the cache to the location we want so to me the best fix would just to be to remove the `EMSCRIPTEN_FETCH_PERSIST_FILE` attribute.

PR #3049, opened 2026-01-23
[WASM] Fix Downloader storage issue, new options for wasm

AlexandreK38:
## Describe your changes

Fix the wasm storage issue described in #3040 by removing `EMSCRIPTEN_FETCH_PERSIST_FILE` and only use memory.

Add 2 new options for WASM:
`AX_WASM_ASSETS_PRELOAD_FILE`: Whether enable to bundle the assets to a .data file and preload them into an IndexedDB cache at first launch (`--preload-file`), otherwise bundle them into the wasm file itself and don't use the IndexedDB at all (`--embed-file`)", default `TRUE`
`AX_WASM_GENERATE_SYMBOL_FILE`: Generate a symbol file for the built wasm file, default `FALSE`

### DIFF
--- a/cmake/Modules/AXBuildHelpers.cmake
+++ b/cmake/Modules/AXBuildHelpers.cmake
@@ -591,9 +591,16 @@ function(ax_setup_app_config app_name)
       if(WINRT)
         ax_target_embed_compiled_shaders(${app_name} ${rt_output} FILES ${app_all_shaders})
       else()
-        # --preload-file
-        # refer to: https://emscripten.org/docs/porting/files/packaging_files.html
-        target_link_options(${app_name} PRIVATE "--preload-file" ${AXSLCC_OUT_DIR}@axslc/)
+
+        set(_WASM_ASSETS_LINKER_FILE "--embed-file")
+
+        if(AX_WASM_ASSETS_PRELOAD_FILE)
+          # --preload-file
+          # refer to: https://emscripten.org/docs/porting/files/packaging_files.html
+          set(_WASM_ASSETS_LINKER_FILE "--preload-file")
+        endif()
+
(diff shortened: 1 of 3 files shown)


cmake/Modules/AXBuildHelpers.cmake
```diff
@@ -591,9 +591,16 @@ function(ax_setup_app_config app_name)
...
-        # refer to: https://emscripten.org/docs/porting/files/packaging_files.html
-        target_link_options(${app_name} PRIVATE "--preload-file" ${AXSLCC_OUT_DIR}@axslc/)
+
+        set(_WASM_ASSETS_LINKER_FILE "--embed-file")
+
+        if(AX_WASM_ASSETS_PRELOAD_FILE)
+          # --preload-file
+          # refer to: https://emscripten.org/docs/porting/files/packaging_files.html
+          set(_WASM_ASSETS_LINKER_FILE "--preload-file")
+        endif()
```
halx99, line 601: Better logic:
```cmake
        if(AX_WASM_ASSETS_PRELOAD_FILE)
          # --preload-file
          # refer to: https://emscripten.org/docs/porting/files/packaging_files.html
          set(_WASM_ASSETS_LINKER_FILE "--preload-file")
        else()
          set(_WASM_ASSETS_LINKER_FILE "--embed-file")
        endif()
```

AlexandreK38: We could do this only once below the option(..)

halx99 (approved):

AlexandreK38:
I didn't make the changes yet 😅

halx99:
it works as excepted, so merge is ok

PR #3076, opened 2026-03-02
DrawNode local properties full supported now, removed all hardcoded clamps;  DrawNode cpp-tests switched to ImGUI ;  

### LINKED ISSUE #3070: DrawNode: thickness is not always correct drawing e.g drawLine(...,1.9f)
DrawNode: thickness is not always correct e.g drawLine(...1.9f) 

see 
Removed hardcoded thickness clamp in DrawNode  #3068

[x] remove all "hardcoded" clamps
[ ] improve/add tests on cpp-tests
[x] check/fix thickness=>1.0,<2.0 issue
[ ] thickness behavior should be the same on all drawing primitives

@w1257  as optional reviewer

aismann:
## Describe your changes
DrawNode::properties full_support + all hardcoded clamps removed
_drawTriangles removed (= part of _drawPoly or _drawPolygon)

=> Redesign: "flat Properties members"

- [x] removed all "_hardcoded clamps_"
- [x] add `const` to all properties getters 
- [x] add missing `DrawNode::properties.getScale()`
- [x] removed  `pragma push_macro("TRANSPARENT")` section
- [x] cpp-tests DrawNode: switch from `ax::ui` to `ax::extension::ImGUI`
- [x] class ax::Label adapt properties: underline/strikethrough
---
- [x]  check/adapt axmol\ui  (axmol\ui\UIMediaPlayer.cpp: 
- [x] check/adapt AX_LABEL_DEBUG_DRAW
- [x] check/adapt AX_SPRITE_DEBUG_DRAW
- [x] check/adapt AX_SPRITEBATCHNODE_DEBUG_DRAW (not implemented)
- [x] check/adapt AX_LABELBMFONT_DEBUG_DRAW 
- [x] check/adapt AX_LABELATLAS_DEBUG_DRAW (not implemented)
- [ ] ~~check/adapt FairyGUI (extensions\fairygui\src\fairygui\GGraph.cpp)~~
- [ ] ~~check/adapt Spine (extensions\spine\src\spine\SkeletonRenderer.cpp)~~
- [x] check/adapt templates\cpp\Source\MainScene.cpp
- [ ] ~~check/adapt tests\live2d-tests\Source\SampleScene.cpp~~
- [x] check/adapt cpp-tests:(ActionsTest/BugsTest/ClippingNodeTest/MultiTouchTest/PhysicsTest/RendererTest/SpritePolygonTest/TileMapTest/UIListViewTest)

see also diffs

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 


aismann:
@w1257 please give feedback too (= optional reviewer)

aismann:
Adapt Label.cpp underline/strikethrough thickness:
image

axmol/2d/DrawNode.cpp
w1257, line 527: Found a issue while testing on one of my projects. Tested on Windows 11.

Image

`linksV->drawSolidRect({7,7},{8,8},{0,0,0,1},1.0f,{1,0,0,1});` 
parent node scale: 32
node scale: 1
Properties scale: 1
Properties factor: 1

tests/cpp-tests/Source/DrawNodeTest/DrawNodeTest.cpp
w1257, line 1: At 21:DrawNode Properties Tests
- unexpected offset after visiting any draw tests in the following list due to most likely transform leftover in Properties.
  - draw(Colored)Triangle
  - drawSolidStar
  - drawStar
  - drawSolidTriangle
  - drawQuadBezier
  - drawCubicBezier
  - drawCardinalSpline
  - drawCatmullRom
- unable to test edge case with node scale 32 in tests (caps at 10). Inspector is not effective.

edge case (same PR review): https://github.com/axmolengine/axmol/pull/3076/changes#r2886060407

axmol/2d/DrawNode.cpp
w1257: Check `Director::getContentScaleFactor()` assumed formula.

w1257 (commented):
For the future PR(s), I suggest moving some of the sliders to the ImGUI inspector.

tests/cpp-tests/Source/DrawNodeTest/DrawNodeTest.cpp
aismann: These tests are not the best (some hard-coded elements, not all drawings are controlled with the sliders).
@w1257
Please add your planned test (your own test).

aismann:
> For the future PR(s), I suggest moving some of the sliders to the ImGUI inspector.

Good idea!
But switching between the inspector and the sliders on the scene is strange.
I think about it.


I like the idea, add an own Dialog:
image

axmol/2d/DrawNode.cpp
aismann: @w1257 
You think the formula is wrong? Its Cocos2dx stuff, old and used so much.

aismann: Workaround:
```
    drawNode->properties.setDrawOrder(true);
    drawNode->drawSolidRect({7, 7}, {8, 8}, {0, 0, 0, 1}, 1.0f, {1, 0, 0, 1});
```
image

I made a improvement for draw(Solid)Rect: 
image

aismann: The reason for this is that rounded corners with different angles work better on polygons.
image

The “special case”: rectangles that always have an angle of 90° have not yet been implemented.
Thanks for point it out. But this will not be part of this PR.

w1257: The previous code assumed that bigger factor = thinner, currently it assumes bigger factor = thicker. Director may still have old assumption.

w1257: I will definitely highlight the importance of DrawOrder when I write the documentation.

tests/cpp-tests/Source/DrawNodeTest/DrawNodeTest.cpp
w1257: I will add the soft test after updating the comment documentation of the DrawNode which I plan on doing next. I will add the hard test after adding the soft test.

axmol/2d/DrawNode.cpp
aismann: @w1257 
I think its fine. 
Have you any problems with this PR or works it as expected?

aismann:
@halx99 
can you have a look plz?
```
[ 40%] Built target plainlua
gmake[1]: *** [CMakeFiles/Makefile2:3637: tests/lua-tests/CMakeFiles/lua-tests.dir/rule] Error 2
gmake: *** [Makefile:879: lua-tests] Error 2
build.ps1: /home/runner/work/axmol/axmol/tools/cmdline/axmol.ps1:89
Line |
  89 |          . $build_script @sub_args
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~
     | Build fail, ret=2
Error: Process completed with exit code 1.
```

axmol/2d/DrawNode.h
```diff
@@ -715,6 +709,11 @@ class AX_DLL DrawNode : public Node, public BlendProtocol
         */
         void setScale(Vec2 s) { scale = s; };
 
+       /** Get the DrawNode scale for each drawing primitive after this.
+
+        */
+        ax::Vec2 getScale() { return scale; };
```
halx99:
```suggestion
        const ax::Vec2& getScale() const { return scale; };
```

aismann: @halx99 
Will add `const` to all properties getters.

rh101:
> > For the future PR(s), I suggest moving some of the sliders to the ImGUI inspector.
> 
> Good idea! But switching between the inspector and the sliders on the scene is strange. I think about it.
> 
ImGUI is not implemented for all platforms (such as iOS), so using ImGUI to control aspects of test cases would cause issues if running the cpp-tests project on such platforms.

aismann:
> > > For the future PR(s), I suggest moving some of the sliders to the ImGUI inspector.
> > 
> > 
> > Good idea! But switching between the inspector and the sliders on the scene is strange. I think about it.
> 
> ImGUI is not implemented for all platforms (such as iOS), so using ImGUI to control aspects of test cases would cause issues if running the cpp-tests project on such platforms.

```
#if AX_ENABLE_EXT_IMGUI && AX_TARGET_PLATFORM != AX_PLATFORM_ANDROID
        addTest("Box2D - TestBed", []() { return new Box2DTestBedTests(); });
#endif
```
I can add this also for DrawNode (for the moment): 
```
#if AX_ENABLE_EXT_IMGUI && AX_TARGET_PLATFORM != AX_PLATFORM_ANDROID
      addTest("DrawNode", []() { return new DrawNodeTests(); });
#endif
```

rh101:
> ```
> #if AX_ENABLE_EXT_IMGUI && AX_TARGET_PLATFORM != AX_PLATFORM_ANDROID
>         addTest("Box2D - TestBed", []() { return new Box2DTestBedTests(); });
> #endif
> ```
> 
> I can add this also for DrawNode (for the moment):
> 
> ```
> #if AX_ENABLE_EXT_IMGUI && AX_TARGET_PLATFORM != AX_PLATFORM_ANDROID
>       addTest("DrawNode", []() { return new DrawNodeTests(); });
> #endif
> ```

What would happen if someone needs to test an issue with `DrawNode` that only occurs on iOS? Not having the test enabled in cpp-tests would mean that is not possible.  It may be best to not use ImGUI for things like this until it works on all platforms.

aismann:
> What would happen if someone needs to test an issue with `DrawNode` that only occurs on iOS? Not having the test  enabled in cpp-tests would mean that is not possible. It may be best to not use ImGUI for things like this until it works on all platforms.

I hope on PR #3083

aismann:
>What would happen if someone needs to test an issue with DrawNode?

@rh101
As usual: Someone sends a code snippet or their own PR with their problem.
Cpp-Test DrawNode is not required if someone wants to test a DrawNode problem.

A small code snippet (for testing) can be tested anywhere on cpp-tests:
I mostly use Bug-350 for testing problems.

rh101:
> @rh101 As usual: Someone sends a code snippet or their own PR with their problem. Cpp-Test DrawNode is not required if someone wants to test a DrawNode problem.
> 

Generally, yes, but the other point of having cpp-tests is to check that functionality is consistent across platforms. Functionality like `DrawNode` is part of the core, and not platform specific, so the tests should cover as many platforms as possible. Hopefully PR #3083 does allow ImGUI to work without issues on the relevant platforms.

aismann:
>  Hopefully PR #3083 does allow ImGUI to work without issues on the relevant platforms.

I also ;)

aismann:
@halx99 
can I use this #if line now?
```
#if AX_ENABLE_EXT_IMGUI
      addTest("DrawNode", []() { return new DrawNodeTests(); });
#endif
```

aismann:
Ready for a further review (if you want)

@halx99 
can be merged.

Tested also on android_armv7 => works.

halx99:
> @halx99 can I use this #if line now?
> 
> ```
> #if AX_ENABLE_EXT_IMGUI
>       addTest("DrawNode", []() { return new DrawNodeTests(); });
> #endif
> ```

Yes, you can

aismann:
> > @halx99 can I use this #if line now?
> > ```
> > #if AX_ENABLE_EXT_IMGUI
> >       addTest("DrawNode", []() { return new DrawNodeTests(); });
> > #endif
> > ```
> 
> Yes, you can

done

axmol/2d/DrawNode.h
```diff
@@ -733,7 +733,7 @@ class AX_DLL DrawNode : public Node, public BlendProtocol
         /** Get the DrawNode center of rotation for each drawing primitive after this.
 
         */
-        Vec2 getCenter() { return center; };
+        const Vec2 getCenter() const { return center; };
```
halx99:
```suggestion
        const Vec2& getCenter() const { return center; };
```

axmol/2d/DrawNode.h
```diff
@@ -743,15 +743,22 @@ class AX_DLL DrawNode : public Node, public BlendProtocol
         /** Get the DrawNode position for drawing primitive.
 
         */
-        Vec2 getPosition() { return position; };
+        const Vec2 getPosition() const { return position; };
```
halx99:
```suggestion
        const Vec2& getPosition() const { return position; };
```

axmol/2d/DrawNode.h
```diff
@@ -743,15 +743,22 @@ class AX_DLL DrawNode : public Node, public BlendProtocol
...
 
-        /** Set all default DrawNode properties.
+        /** Get the DrawNode factor for drawing primitive.
+
+        */
+        const float getFactor() const { return factor; };
```
halx99: No const qualifier requried when return type is non-ref scalar types

axmol/2d/DrawNode.h
```diff
@@ -708,13 +702,19 @@ class AX_DLL DrawNode : public Node, public BlendProtocol
...
 
         /** Set the DrawNode scale for each drawing primitive after this.
 
         */
         void setScale(Vec2 s) { scale = s; };
 
+        /** Get the DrawNode scale for each drawing primitive after this.
+
+         */
+        // ax::Vec2 getScale() { return scale; };
```
halx99: Please cleanup comment code

aismann: done

axmol/2d/DrawNode.h
aismann: fix it for the other properties getters also (return type float, bool)

axmol/2d/DrawNode.h
aismann: done

axmol/2d/DrawNode.h
```diff
@@ -743,15 +742,22 @@ class AX_DLL DrawNode : public Node, public BlendProtocol
...
         */
-        Vec2 getPosition() { return position; };
+        const Vec2& getPosition() const { return position; };
 
-        /** Set all default DrawNode properties.
+        /** Get the DrawNode factor for drawing primitive.
 
         */
-        void setDefaultValues();
         float getFactor() { return factor; };
```
halx99:
```suggestion
        float getFactor() const { return factor; };
```

aismann:
@halx99 
is anything missed?
Review @rh101 ?

rh101:
> Review @rh101 ?

Sorry, I won't be able to due to time constraints, and my entire setup is still based on v2, so not running anything with v3 at the moment.

aismann:
> > Review @rh101 ?
> 
> Tut mir leid, ich kann das wegen Zeitmangels nicht, und mein gesamtes Setup basiert immer noch auf v2, daher lasse ich momentan nichts mit v3 laufen.
@rh101 
That's fine. I think we can merge even without your valued comments.
Thanks as always.
@halx99  can it merged, or any open points?

halx99:
About properties' bool member naming style, I suggest:

- `properties.transform` -> `properties.transformEnabled`
- `properties.drawOrder` -> `properties.preserveDrawOrder`
- `properties.getTransform` -> `properties.isTrasnformEnabled`
- `properties.getDrawOrder` -> `properties.isPreserveDrawOrder` 


And `DrawNode::properties -> `DrawNode::_properties`

aismann:
> About properties' bool member naming style, I suggest:
> 
> * `properties.transform` -> `properties.transformEnabled`
> * `properties.drawOrder` -> `properties.preserveDrawOrder`
> * `properties.getTransform` -> `properties.isTrasnformEnabled`
> * `properties.getDrawOrder` -> `properties.isPreserveDrawOrder`
> 
> And `DrawNode::properties -> `DrawNode::_properties`



> About properties' bool member naming style, I suggest:
> 
> * `properties.transform` -> `properties.transformEnabled`
> * `properties.drawOrder` -> `properties.preserveDrawOrder`
> * `properties.getTransform` -> `properties.isTrasnformEnabled`
> * `properties.getDrawOrder` -> `properties.isPreserveDrawOrder`
> 
> And `DrawNode::properties -> `DrawNode::_properties`

@halx99, @rh101 
Long commands. Is that a good idea?
=> Maybe it would be better if we renamed all the property names?

As for `DrawNode::_properties`, isn't the underscore “_” supposed to be used for private members?

rh101:
> Long commands. Is that a good idea? => Maybe it would be better if we renamed all the property names?
> 
> As for `DrawNode::_properties`, isn't the underscore “_” supposed to be used for private members?

It is an internal field, and to keep the interface consistent with the rest of Axmol, then it would have been better to use methods to access a private `_properties` member.  Something like:

`Properties& getProperties() const { return _properties; }`

It would have made sense to also provide something like `setProperties(const Properties& properties)` for cases where you may want to copy pre-configured settings to a `DrawNode`, without having to set individual members.

The "long commands" are not really relevant, since it's not the conventional style, and would make more sense like this:
```
auto& propertie = drawNode->getProperties();
properties.transformEnabled = true;
```
That does not seem excessive in any way, and would follow the convention used in the other parts of Axmol.

Note, this is just my personal opinion on the matter, nothing more.

halx99:
since all members of class `DrawNode::Properties` is public, all access methods of `DrawNode::Properties` could be removed:

```cpp
   struct Properties
    {
        float factor{1.0f};  /// thickness scale factor

        // transforming stuff
        Vec2 scale;
        Vec2 center;
        float rotation{0};
        Vec2 position;

        // Drawing flags
        bool transform = false;
        bool drawOrder = false;
      }
```

aismann:
Add ax::Label underline/strikethrough test to DrawNode
image

aismann:
@halx99 
From my point of view, ready for review again.

axmol/2d/DrawNode.cpp
```diff
@@ -1592,19 +1408,18 @@ void DrawNode::applyTransform(const Vec2* from, Vec2* to, unsigned int count)
     }
 }
 
-void DrawNode::Properties::setDefaultValues()
-{
-    auto fac = Director::getInstance()->getContentScaleFactor();
-    factor   = fac;
+//void DrawNode::Properties::setDefaultValues()
```
halx99: comment chould be removed

axmol/2d/DrawNode.h
```diff
@@ -665,94 +659,94 @@ class AX_DLL DrawNode : public Node, public BlendProtocol
...
-        /** Get the DrawNode transform
-         *
-         */
-        bool getTransform(void) { return transform; };
 
-        /** Set the DrawNode scale for each drawing primitive after this.
 
-        */
-        void setScale(Vec2 s) { scale = s; };
+    // Flat Properties members section
```
halx99: this comment should be removed

axmol/2d/DrawNode.h
```diff
@@ -665,94 +659,94 @@ class AX_DLL DrawNode : public Node, public BlendProtocol
...
+    Vec2 getLocalScale() const { return _localScale; }
 
-        */
-        void setCenter(Vec2 c) { center = c; };
+    void setLocalCenter(const Vec2& c)
+    {
+        _localCenter    = c;
+        _trianglesDirty = _linesDirty = true;
+    }
+    Vec2 getLocalCenter() const { return _localCenter; }
```
halx99:
```suggestion
    const Vec2& getLocalCenter() const { return _localCenter; }
```

axmol/2d/DrawNode.h
```diff
@@ -665,94 +659,94 @@ class AX_DLL DrawNode : public Node, public BlendProtocol
...
+        _linesDirty = _trianglesDirty = true;
+    }
 
-        /** Set the DrawNode center of rotation for each drawing primitive after this.
+    void setLocalScale(const Vec2& s)
+    {
+        _localScale     = s;
+        _trianglesDirty = _linesDirty = true;
+    }
+    Vec2 getLocalScale() const { return _localScale; }
```
halx99:
```suggestion
    const Vec2& getLocalScale() const { return _localScale; }
```

axmol/2d/DrawNode.h
```diff
@@ -665,94 +659,94 @@ class AX_DLL DrawNode : public Node, public BlendProtocol
...
+        _localRotation         = 0.0f;
+        _localRotationRad      = 0.0f;
+        _localPosition         = Vec2::ZERO;
+        _preserveDrawOrder     = false;
+        _localTransformEnabled = false;
+        _trianglesDirty = _linesDirty = _pointsDirty = true;
+    }
 
-        /** Get the DrawNode position for drawing primitive.
+private:
```
halx99:
```suggestion
protect:
```

halx99 (changes requested):

aismann (commented):
Approved. Thanks for helping hands

halx99:
I also fix the duplicated condition check `_preserveDrawOrder` in drawPoint, please review

image

aismann:
> I also fix the duplicated condition check `_preserveDrawOrder` in drawPoint, please review
> 
It looks good on my computer. As always, thank you very much.
I will improve some of the DrawNode tests in a new PR in a few days.

halx99 (approved):

PR #3200, opened 2026-06-17
Refactor math constants to lowercase naming

halx99:
- Convert all static math constants (Vec2, Vec3, Vec4, Mat4, Color32, Color, Rect, Quaternion, AffineTransform) to lowercase naming convention
- Keep uppercase aliases as [[deprecated]] for backward compatibility
- Replace all references to old uppercase constants with new lowercase names across engine core, extensions, and tests
- Make constructors constexpr for literal type support (Mat4, Vec4, Color, Color32)
- Replace static factory methods with constexpr data members (Quaternion identity)
- Use inline constexpr out-of-class definitions for C++17 compatibility
- Special case: Color::TRANSPARENT -> Color::clear (not transparent)

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 


halx99:
@gemini review

PR #3277, opened 2026-08-10
[v2] Provide access to get current editing state of EditBox

### LINKED ISSUE #3276: Editing state of EditBox is not consistent across platforms and not exposed via EditBox API
- axmol version: 2.11.4
- devices test on:
- developing environments
   - NDK version: r23c
   - Xcode version: 14.2+
   - Visual Studio: 
      - VS version: 2022 (17.9+)
      - MSVC version: 19.39+
   - cmake version: 
Steps to Reproduce:

The API `EditBoxImpl`  has a method named `bool isEditing()`, but the implementation is not consistent, and does not return the real state of the edit box:

Win32/Android/Linux/WASM:
`return false;`

iOS/MaxOS:
`return [_systemControl isEditState] ? true : false;`

WinRT/UWP:
`return _isEditing;`

The `EditBoxImplCommon` has a member named `_editingMode`, which seems to be set to the correct state via these methods:

```
void EditBoxImplCommon::openKeyboard()
{
    _editBox->setBrightStyle(Widget::BrightStyle::HIGHLIGHT);
    _label->setVisible(false);
    _labelPlaceHolder->setVisible(false);
    _editingMode = true;
    this->setNativeVisible(true);
    this->nativeOpenKeyboard();
}

void EditBoxImplCommon::closeKeyboard()
{
    this->nativeCloseKeyboard();
    _editingMode = false;
}

void EditBoxImplCommon::onEndEditing(std::string_view /*text*/)
{
    _editBox->setBrightStyle(Widget::BrightStyle::NORMAL);
    _editingMode = false;
    this->setNativeVisible(false);
    refreshInactiveText();
}
```

The WinRT implementation has its own internal `_isEditing` flag, but the `EditBoxImplCommon::_editingMode ` is synchronized to the same value each time it changes.

The main issue is that some platforms just return a hard-coded `false`, which is not correct, and in addition to that, there is no way to read this state from the `EditBox` API.

The `EditBox` does have a `ax::ui::EditBoxDelegate` that can be assigned to the `EditBox` when it is instantiated, but this only for external observers to the state.  The two main methods that handle this callback, along with changing the value of `_editingMode`, are not virtual, so cannot be hooked into when sub-classing `EditBox`:

```    
void EditBox::editBoxEditingDidBegin();
void EditBox::editBoxEditingDidEnd(std::string_view text,
                              EditBoxDelegate::EditBoxEndAction action = EditBoxDelegate::EditBoxEndAction::UNKNOWN);
```

So, is there any particular reason that `EditBoxImpl::isEditing()` does not return the correct value across all platforms? If there is no known reason, then for all platforms that currently return `false`, can we change this so that they return `EditBoxImplCommon::_editingMode`, such as:

```
bool EditBoxImplWin::isEditing()
{
    return _editingMode;
}
```

Also, in `EditBox`, can we add a method to give access to the editing state:
```
bool EditBox::isEditing()
{
    return _editBoxImpl->isEditing();
}
```

This change should not impact any existing usage of `EditBox` on any platform.

@halx99 Please let me know if you agree with the above, and if you want me to create PR with the changes.

rh101:
## Describe your changes

Remove hard-coded editing state for platforms that had it set, and instead return current editing state that was already available via the `_editingMode` flag.

Add `EditBox::isEditing()` method to expose the current editing state of the EditBox.

Tested Wasm, Win32, and Android, and they're working correctly.  The only one I can't test at the moment is Linux, but it shouldn't be any different, since it is returning the existing `_editingMode` value as well.  The other platforms were not changed, since they already returned the correct value.

#3276 

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/core/ui/UIEditBox/UIEditBox.cpp
+++ b/core/ui/UIEditBox/UIEditBox.cpp
@@ -881,12 +881,7 @@ void EditBox::setGlobalZOrder(float globalZOrder)
 
 bool EditBox::isEditing() const
 {
-    if (_editBoxImpl)
-    {
-        return _editBoxImpl->isEditing();
-    }
-
-    return false;
+    return _editBoxImpl && _editBoxImpl->isEditing();
 }
 
 #if AX_ENABLE_SCRIPT_BINDING
--- a/core/ui/UIEditBox/UIEditBox.cpp
+++ b/core/ui/UIEditBox/UIEditBox.cpp
@@ -879,6 +879,16 @@ void EditBox::setGlobalZOrder(float globalZOrder)
     }
 }
 
+bool EditBox::isEditing() const
+{
+    if (_editBoxImpl)
+    {
+        return _editBoxImpl->isEditing();
+    }
+
+    return false;
+}
+
 #if AX_ENABLE_SCRIPT_BINDING
 void EditBox::registerScriptEditBoxHandler(int handler)
 {
--- a/core/ui/UIEditBox/UIEditBox.h
+++ b/core/ui/UIEditBox/UIEditBox.h
@@ -639,6 +639,12 @@ public:
     void openKeyboard() const;
     void closeKeyboard() const;
 
+    /**
+     * Get the editing state of the EditBox
+     * @return true if editing
+     */
+    bool isEditing() const;
+
 protected:
     virtual void releaseUpEvent() override;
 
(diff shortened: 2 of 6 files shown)


core/ui/UIEditBox/UIEditBox.cpp
```diff
@@ -879,6 +879,16 @@ void EditBox::setGlobalZOrder(float globalZOrder)
     }
 }
 
+bool EditBox::isEditing() const
+{
+    if (_editBoxImpl)
```
halx99: more simple code: `return _editBoxImpl && _editBoxImpl->isEditing();`

halx99 (changes requested):

halx99 (approved):

PR #3284, opened 2026-08-17
[v2] Update to use UIScene for iOS13+

### LINKED ISSUE #3282: Warning `UIScene lifecycle will soon be required. Failure to adopt will result in an assert in the future.`
- axmol version: 2.11.4
- devices test on: iPhone 13 Pro
- developing environments
   - Xcode version: 26.6
  
Steps to Reproduce:

1. When running any Axmol project on Xcode, the first log message that will appear is `UIScene lifecycle will soon be required. Failure to adopt will result in an assert in the future.`

Following the [Apple documentation](https://developer.apple.com/documentation/UIKit/transitioning-to-the-uikit-scene-based-life-cycle), from iOS 27 apps that fail to migrate to the new UIScene lifecycle won't launch. Since iOS 27 is expected to come out in September, I'm marking this as urgent.

In the same page, there are instructions on [how to migrate](https://developer.apple.com/documentation/UIKit/transitioning-to-the-uikit-scene-based-life-cycle#Adopt-the-scene-based-life-cycle): adding Application Scene Manifest (`UIApplicationSceneManifest`) to `info.plist`, [configure a window scene](https://developer.apple.com/documentation/uikit/responding-to-the-launch-of-your-app) (`SceneDelegate`),  and [migrate the methods from "application" to "scene"](https://developer.apple.com/documentation/UIKit/transitioning-to-the-uikit-scene-based-life-cycle#Migrate-app-life-cycle-logic).

For supporting iOS 12 and earlier, something like this might be needed:

```
    if (@available(iOS 13.0, *)) {
        // iOS 13+ uses SceneDelegate, so do nothing here.
    } else {
        // iOS 12 and earlier: manually create the window.
        ...
    }
```
Probably there are more things to take in account for Axmol case, but I hope this helps.

rh101:
## Describe your changes

Changes made to support UIScene for iOS13.0, with the relevant checks added for building with iOS versions below 13.0.

This requires a change in the game-side templates as well:
1. Required fields added to `Info.plist` (refer to [this](https://developer.apple.com/documentation/uikit/transitioning-to-the-uikit-scene-based-life-cycle?language=objc#Adopt-the-scene-based-life-cycle))
3. New `GameSceneDelegate` source files added for the developer to override the default functionality.

These changes are backwards compatible as long as the game project does not override certain functionality in `GameAppController`, such as:
```
applicationDidBecomeActive
applicationWillResignActive
applicationDidEnterBackground
applicationWillEnterForeground
```

If you have a need to override those methods, then you will need the new `GameSceneDelegate.h/.mm` files and the updated section in `Info.plist`.  These are the equivalent methods that you may override in `GameSceneDelegate`:
```
sceneDidBecomeActive
sceneWillResignActive
sceneDidEnterBackground
sceneWillEnterForeground
```

Also, if a project uses custom code in `GameViewController`, then you will need to ensure that `GameSceneDelegate` is being used, since it will handle the creation of the `GameViewController`.

If someone has a better solution, then please put it forward.

#3282 

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 

### DIFF
--- a/core/platform/ios/AxmolAppController.mm
+++ b/core/platform/ios/AxmolAppController.mm
@@ -94,7 +94,7 @@ configurationForConnectingSceneSession:(UISceneSession *)connectingSceneSession
     return YES;
 }
 
-#if __IPHONE_OS_VERSION_MAX_ALLOWED < 130000
+#if defined(__IPHONE_OS_VERSION_MIN_REQUIRED) && __IPHONE_OS_VERSION_MIN_REQUIRED < 130000
 - (void)applicationWillResignActive:(UIApplication*)application
 {
     /*
@@ -124,7 +124,14 @@ configurationForConnectingSceneSession:(UISceneSession *)connectingSceneSession
      information to restore your application to its current state in case it is terminated later. If your application
      supports background execution, called instead of applicationWillTerminate: when the user quits.
      */
-    ax::Application::getInstance()->applicationDidEnterBackground();
+    if (@available(iOS 13.0, *))
+    {
+        // do nothing here for iOS13+
+    }
+    else
+    {
+        ax::Application::getInstance()->applicationDidEnterBackground();
+    }
 }
 
 - (void)applicationWillEnterForeground:(UIApplication*)application
@@ -133,7 +140,14 @@ configurationForConnectingSceneSession:(UISceneSession *)connectingSceneSession
      Called as part of  transition from the background to the inactive state: here you can undo many of the changes made
      on entering the background.
      */
-    ax::Application::getInstance()->applicationWillEnterForeground();
+    if (@available(iOS 13.0, *))
+    {
+        // do nothing here for iOS13+
+    }
+    else
+    {
+        ax::Application::getInstance()->applicationWillEnterForeground();
+    }
 }
 #endif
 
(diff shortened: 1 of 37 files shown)


rh101:
@halx99 This should be complete. Please let me know if you want a PR created for v3 as well.

halx99:
> @halx99 This should be complete. Please let me know if you want a PR created for v3 as well.

sure.

rh101:
Found an issue when working on the v3 version, so the extra commits are to address that.  Also, I'm not sure that `createRootViewController` in `GameAppController.mm` was ever working, since the return type did not match the base class method in `AxmolAppController.mm`.  Anyhow, it's all sorted now.  The v3 PR will be up shortly.

core/platform/ios/AxmolAppController.mm
```diff
@@ -33,40 +34,67 @@ of this software and associated documentation files (the "Software"), to deal
...
+        auto renderView = ax::RenderViewImpl::createWithFullScreen("axmol2");
+        _viewController = [self createRootViewController];
+        
+        renderView->showWindow(_viewController);
+        
+        // IMPORTANT: Setting the RenderView should be done after creating the RootViewController
+        Director::getInstance()->setRenderView(renderView);
+        
+        // run the axmol game scene
+        axmolApp->run();
```
halx99: Could we factor the common RenderView setup and axmolApp->run() logic into a shared helper, since we now have two launch paths?

rh101: Yes, I can do that. I'll sort it out tomorrow.

halx99 (approved):

PR #3288, opened 2026-08-19
Add extension: svg to support create Sprite from SVG files

### LINKED ISSUE #3243: Axmol 3:  New Feature: SpriteFromSVG
Disclaimer:
For all SVGs copyrights belong to vecteezy and his authors!
https://de.vecteezy.com

Here the SVG (Vertices: ~5167) 
Image

And here the corresponding sprite (Vertices: ~72)
Image

getContentSize:
Image

https://github.com/user-attachments/assets/91728e70-ffea-481a-9018-1b4c00dfcfd5

aismann:
## Describe your changes
New Feature: Sprite from SVG files

History:
-	Started with an own SVG parser/renderer
=> Aborted: To much effort and maintenance in the future of Axmol lifecircle  
-	Tested nanosvg ([memononen/nanosvg: Simple stupid SVG parser](https://github.com/memononen/nanosvg))
Good rendering but not all was as expected.
=> Aborted: This project is not actively maintained.
-	plutosvg/plutovg: 
=> Seems to be a good 3rdparty lib for feature “Sprite from SVG files”
      but we let it still experimental!

 #3243

### For each PR
       
   Optional:
          
### For core/new feature PR

### For each 3.x PR 


aismann:
@halx99 
Please check/edit the 3rdparty "cmake stuff" of plutosvg/plutovg (Im using the "orginal" but its to much there).
Thanks.

aismann:
Add some SVG files to the axmol-sample-assets:
https://github.com/axmolengine/axmol-sample-assets/pull/5
@halx99 
Please merge it. Thanks.

rh101:
~~Can this feature be optional, enabled by a cmake option?~~ I just noticed the `option(AX_WITH_SVG "Build with internal SVG support" ON)`

I personally feel that it should be off by default, and if it becomes widely used, then that would be a reason to set it to on by default later.

Also, in `Image.h/cpp`, there should be a check if it is enabled in the cmake options.  At the moment there doesn't seem to be any checks at all, so it will result in a build error if SVG is disabled.  Also, it doesn't seem like a good idea to have to change both a cmake option and `config.h`; there should be only one place to enable and disable it.

halx99:
> ~Can this feature be optional, enabled by a cmake option?~ I just noticed the `option(AX_WITH_SVG "Build with internal SVG support" ON)`
> 
> I personally feel that it should be off by default, and if it becomes widely used, then that would be a reason to set it to on by default later.

I agreed.

halx99:
And I think SVG support should becomes an optional Axmol extension and is disabled by default.

rh101:
> And I think SVG support should becomes an optional Axmol extension and is disabled by default.

Having it as an extension would make more sense.

aismann:
> > And I think SVG support should becomes an optional Axmol extension and is disabled by default.
> 
> Having it as an extension would make more sense.

I agree too!

aismann:
Planned changes:
- plutsvg/plutovg be still part of the 3rdpary folder
=> But will be clipped on/off via settings (e.g. cmake or config.h)

- Axmol svg stuff will be add as extension  
=> need some more redesign

**Another, easier way! is to switch on/off via #defines like its used for platforms or other extensions:**
e.g:
```
#if AX_ENABLE_EXT_IMGUI
...
#endif
```
=> No big changes needed (only hide SVG source snippet on class sprite/images/...) 
```
#if AX_WITH_SVG ...  (or any other name like: AX_ENABLE_EXT_SVG)
 ...
#endif
```

@rh101 , @halx99 
What are you think? I prefer for  **switch on/off via #defines**

rh101:
> What are you think? I prefer for **switch on/off via #defines**

That won't stop the 3rd party SVG library from being built.  The whole point is that we have a way to enable or disable the building of the library, otherwise it will take more time to build and wasted space in the final binary.

The cmake option to enable/disable the 3rd party SVG library should control everything.  Enabling this should automatically add a define to enable the Axmol-specific SVG support in `Image.h/cpp`, cpp-tests etc. etc., which means a separate pre-processor definition isn't required in `config.h`.  It is unlikely that a developer would enable the SVG library compilation and linking if they are not planning to use the SVG support in Axmol.

For example, in `/3rdparty/CMakeLists.txt`, something like:
```
if(AX_WITH_SVG)
  ax_add_3rd(plutovg)
  ax_add_3rd(plutosvg)
  target_compile_definitions(${_AX_CORE_LIB} PUBLIC AX_USE_SVG=1)
endif()
```

Does that make sense?

aismann:
> The cmake option to enable/disable the 3rd party SVG library should control everything. Enabling this should automatically add a define to enable the Axmol-specific SVG support in `Image.h/cpp`, cpp-tests etc. etc., which means a separate pre-processor definition isn't required in `config.h`. It is unlikely that a developer would enable the SVG library compilation and linking if they are not planning to use the SVG support in Axmol.
> 
> Does that make sense?

@rh101 
Correct. Using one place only makes sense. 
`config.h` changes will be removed.

halx99:
AX_WITH_XXX options are internal and undocumented, and should not be used as user-facing feature switches. Third-party dependencies should instead be enabled automatically based on Axmol feature options such as AX_ENABLE_EXT_XXX.
Actually, I also plan to gradually remove the existing AX_WITH_XXX options over time.

rh101:
@aismann I received a request for review notification, but I'm not sure why, since nothing has changed since the earlier conversation.  Was the review request accidental?

aismann:
> @aismann I received a request for review notification, but I'm not sure why, since nothing has changed since the earlier conversation. Was the review request accidental?
@rh101 
I only add you to the reviewer group. Thats often done before the review is starting.

aismann:
This PR is still in "draft". A prev-review makes often sense. 
@rh101 @halx99 please commit this changes:
I add a new `AX_ENABLE_3RD_SVG` which control 3rdparty libs, its like `AX_ENABLE_EXT_xxx`  (which control  extensions)

axmol/platform/Image.cpp
```diff
@@ -34,6 +34,9 @@ THE SOFTWARE.
 
 #include "axmol/tlx/utility.hpp"
 #include "axmol/base/Config.h"  // AX_USE_JPEG, AX_USE_WEBP
+#include <plutosvg.h>
```
rh101: Missing `#if AX_ENABLE_3RD_SVG` around the include statements for the external library headers. Also, naming it `AX_ENABLE_SVG` is perhaps better.

3rdparty/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ option(AX_WITH_CLIPPER2 "Build with internal Clipper2 support" ON)
 option(AX_WITH_POLY2TRI "Build with internal poly2tri support" ON)
 option(AX_WITH_FASTLZ "Build with internal fastlz support" ON)
 option(AX_WITH_TRACY "Build with internal Tracy support" ON)
+option(AX_WITH_SVG "Build with internal SVG support" ON)
```
rh101: These should be default `OFF`.

Also, as halx99 previous mentioned, avoid using the `AX_WITH` prefix for features like this, and instead use `AX_ENABLE_SVG` or `AX_ENABLE_EXT_SVG`.

tests/cpp-tests/CMakeLists.txt
```diff
@@ -350,6 +350,11 @@ list(APPEND GAME_SOURCE
   Source/ZipTest/ZipTests.cpp
 )
 
+if(AX_ENABLE_3RD_SVG)
```
rh101: The cpp-tests should be enabling this feature, and not checking if it is enabled.  The whole point is to include this feature in the tests.

3rdparty/CMakeLists.txt
aismann: > These should be default `OFF`

Correct. Its temporary 'ON' for testing the builds of all platforms with SVG support.
Will be 'OFF' before it goes from 'draft' to 'ready for review'.

rh101: The cpp-tests project is what gets built for testing, so the SVG support should be enabled in the `cpp-tests/ CMakeLists.txt`.  It does not need to be temporarily enabled in he `3rdparty/CMakeLists.txt`.

This may work:
`set(AX_WITH_SVG ON CACHE BOOL "Build with internal SVG support" FORCE)`

tests/cpp-tests/CMakeLists.txt
aismann: @rh101, good point. Thanks.

halx99:
About cmake public options, follow up: https://github.com/axmolengine/axmol/blob/dev/CMakeOptions.md

halx99:
so if svg is core optional feature: the option should be: AX_ENABLE_SVG

3rdparty/CMakeLists.txt
aismann: should be work now

aismann:
PR is ready for review now.
@rh101 @halx99 please review again. Thanks.

tests/cpp-tests/CMakeLists.txt
aismann: should be work now

rh101:
> PR is ready for review now. @rh101 @halx99 please review again. Thanks.

It seems that the cpp-tests still isn't quite right.  All core features should be enabled when building cpp-tests, since it's a pre-made project, which both tests the build and also showcases features. 

That means this:
```
if(AX_ENABLE_SVG)
  list(APPEND GAME_HEADER Source/SVGTest/SVGTest.h)
  list(APPEND GAME_SOURCE Source/SVGTest/SVGTest.cpp)
endif()
```

Should not be surrounded by a conditional.

It should be enabling it, and not checking if SVG is enabled.  

For example, add this to the `cpp-tests/CMakeLists.txt`:
`set(AX_ENABLE_SVG ON CACHE BOOL "Build with internal SVG support" FORCE)`

tests/cpp-tests/CMakeLists.txt
```diff
@@ -350,6 +350,11 @@ list(APPEND GAME_SOURCE
   Source/ZipTest/ZipTests.cpp
 )
 
+if(AX_ENABLE_SVG)
```
rh101: This should not have a conditional.  It should be always built with SVG support.

Turn SVG support on at the top of the `cpp-tests/CMakeLists.txt`:
set(AX_ENABLE_SVG ON CACHE BOOL "Build with internal SVG support" FORCE)

aismann:
@rh101 
Seems its not enough or to late:
Still disabled e.g on Image.h:
image

halx99:
> @rh101 Seems its not enough or to late: Still disabled e.g on Image.h: image

Check other preprocessor check: https://github.com/axmolengine/axmol/blob/dev/axmol/physics/3d/Collider3D.h#L35

3rdparty/CMakeLists.txt
```diff
@@ -425,6 +425,13 @@ if(AX_WITH_TRACY)
   ax_add_3rd(${tracy_SOURCE_DIR} TARGETS TracyClient)
 endif()
 
+# SVG
+if(AX_ENABLE_SVG)
+  ax_add_3rd(plutovg)
+  ax_add_3rd(plutosvg)
+  target_compile_definitions(plutosvg PUBLIC AX_ENABLE_SVG=1)
```
rh101: This is applying the pre-processor definition to the wrong target.

rh101:
> target_compile_definitions(plutosvg PUBLIC AX_ENABLE_SVG=1)

That is not correct. As I mentioned in a post [above](https://github.com/axmolengine/axmol/pull/3288#issuecomment-5342856018), it should be this:

`target_compile_definitions(${_AX_CORE_LIB} PUBLIC AX_USE_SVG=1)`

Even better, follow what @halx99 mentioned above, which is by using the `ax_config_pred` macro:

```
ax_config_pred(${_AX_CORE_LIB} AX_USE_SVG)
```

aismann:
@halx99 
All core features should be enabled when building cpp-tests, 
SVG support is not ON on cpp-tests:
image

halx99:
I prefer implement svg support as extension feature currently.

halx99 (approved):

halx99:
This PR still needs to be redesigned to avoid unnecessary intrusion into and complication of the core architecture.

aismann:
> This PR still needs to be redesigned to avoid unnecessary intrusion into and complication of the core architecture.

The goal of this PR is/was to show an effektive  way for the feature:  create  sprites from SVGs.
Axmol v3  has so much redesigns, thats only one more.
@halx99 
Make the needed changes plz.
New features have often a redesign!
About the redesign:
Please think about gif (or any other movie frame) support too.
GIFs (or movie frames) can handle  eg. like ax::SpriteFrame.

halx99:
I kept this PR focused on its original goal: providing a practical way to create sprites from SVG files.

The latest commits eeps the SVG implementation isolated in the ax::ext namespace. SVG can be loaded explicitly with createSVGSprite() or loadSVGTexture(), while the Core remains unaware of SVG-specific details.

This keeps the feature focused and provides a clean foundation for SVG support in Axmol v3.

halx99:
I’ve finished my work.

aismann:
My point of view:
1. Its works
2. I miss the nice feature: `ax::Sprite::create("aNiceSVGfile")`
    => auto selection of the right encoder for 'svg' files like for 'bmp', 'png', '...' which use also only:  `ax::Sprite::create("...");`
    
    => PR can be merged ;)
    
Another issue:
 I switch also the 'AX_ENABLE_EXT_EFFEKSEER = ON'
 and get this warning during build:
 
image

halx99:
@aismann Thansk for you testing, @rh101  do you have any other suggestions?

rh101:
> @aismann Thansk for you testing, @rh101 do you have any other suggestions?

My original concern was the source code would be compiled in without a way to disable it, since it may not be a feature that is widely used.  Having it as an extension makes sense, and doesn't seem to limit its usage in any way.

